PR #47, opened 2025-10-16
Ryanontheinside/feat/prompt blending

ryanontheinside:
# Prompt Blending

Adds support for blending multiple prompts with configurable weights and interpolation methods.

## Key Features
- **Multi-prompt input**: Up to 4 prompts with individual weight sliders
- **Interpolation methods**: Linear (2-4 prompts) or Slerp (2 prompts)
- **Smart caching**: LRU cache for encoded embeddings
- **Auto-normalize**: Weights automatically normalized to 100%

## Implementation
- New `PromptBlender` class with linear and slerp interpolation
- Frontend UI with expandable textarea and weight controls
- Integrated with LongLive and StreamDiffusionV2 pipelines

## Next Steps
- SeedBlender

### DIFF
--- a/frontend/src/components/PromptInput.tsx
+++ b/frontend/src/components/PromptInput.tsx
@@ -91,14 +91,18 @@ export function PromptInput({
 
   // Calculate normalized weights for display
   const totalWeight = prompts.reduce((sum, p) => sum + p.weight, 0);
-  const normalizedWeights = prompts.map(
-    p => (totalWeight > 0 ? (p.weight / totalWeight) * 100 : 0)
+  const normalizedWeights = prompts.map(p =>
+    totalWeight > 0 ? (p.weight / totalWeight) * 100 : 0
   );
 
   const isSinglePrompt = prompts.length === 1;
 
   // Render a single prompt field with expandable textarea
-  const renderPromptField = (index: number, placeholder: string, showRemove: boolean) => {
+  const renderPromptField = (
+    index: number,
+    placeholder: string,
+    showRemove: boolean
+  ) => {
     const isFocused = focusedIndex === index;
     const prompt = prompts[index];
 
@@ -130,9 +134,7 @@ export function PromptInput({
         <Button
           onClick={handleSubmit}
           disabled={
-            disabled ||
-            !prompts.some(p => p.text.trim()) ||
-            isProcessing
+            disabled || !prompts.some(p => p.text.trim()) || isProcessing
           }
           size="sm"
           className="rounded-full w-8 h-8 p-0 bg-black hover:bg-gray-800 text-white disabled:opacity-50 disabled:cursor-not-allowed"
--- a/frontend/src/components/ui/textarea.tsx
+++ b/frontend/src/components/ui/textarea.tsx
@@ -1,6 +1,6 @@
-import * as React from "react"
+import * as React from "react";
 
-import { cn } from "@/lib/utils"
+import { cn } from "@/lib/utils";
 
 const Textarea = React.forwardRef<
   HTMLTextAreaElement,
@@ -15,8 +15,8 @@ const Textarea = React.forwardRef<
       ref={ref}
       {...props}
     />
-  )
-})
-Textarea.displayName = "Textarea"
+  );
+});
+Textarea.displayName = "Textarea";
 
-export { Textarea }
+export { Textarea };


pipelines/interface.py
```diff
@@ -5,6 +5,15 @@
 import torch
 from pydantic import BaseModel
 
+# Parameters that are only used in prepare() and should not be passed to __call__()
+PREPARE_ONLY_PARAMS = frozenset(
```
leszko, line 9: Should it be in the `interface.py`? Maybe we could move it to `frame_processor.py`?

pipelines/streamdiffusionv2/pipeline.py
```diff
@@ -67,23 +68,29 @@ def __init__(
...
         self.current_end = self.stream.frame_seq_length * 2
 
     def prepare(self, should_prepare: bool = False, **kwargs) -> Requirements:
         if should_prepare:
-            logger.info("Initiating pipeline prepare for request")
+            logger.info("prepare: Initiating pipeline prepare for request")
 
         manage_cache = kwargs.get("manage_cache", None)
         prompts = kwargs.get("prompts", None)
+        prompt_interpolation_method = kwargs.get(
```
leszko, line 84: If all these pipeline parameters are moved here, I think we may want to remove them from the `__call__()`. They are not used in `__call()__` anymore.

pipelines/longlive/pipeline.py
```diff
@@ -112,9 +118,29 @@ def __call__(
         denoising_step_list: list[int] = None,
         manage_cache: bool = True,
     ):
-        self.prepare(
-            prompts=prompts,
-            denoising_step_list=denoising_step_list,
-            manage_cache=manage_cache,
-        )
+        # Note: prepare() was already called by frame_processor before __call__
```
yondonfu, line 121: Noting a bug here that might be related to changes around this area.

https://github.com/user-attachments/assets/4ee84996-864c-4771-b7e5-5d6d85b8d155

In this video, I am using longlive and I:

1. Start the stream.
2. Disable Manage Cache
3. Click the Reset Cache button

And the video starts looping.

frontend/src/components/InputAndControlsPanel.tsx
```diff
@@ -203,6 +215,18 @@ export function InputAndControlsPanel({
             </Button>
           </div>
         </div>
+
```
yondonfu, line 218: Screenshot 2025-10-17 at 11 17 00 AM

As we add more prompts, we can end up with vertical overflow. Can we update the InputAndControlsPanel so that it supports scrolling if the content will not fit vertically and show the scrollbar in that scenario? Can see how it is done in SettingsPanel for reference.

frontend/src/components/PromptInput.tsx
```diff
@@ -1,67 +1,242 @@
...
-        className="rounded-full w-8 h-8 p-0 bg-black hover:bg-gray-800 text-white disabled:opacity-50 disabled:cursor-not-allowed"
+  // Calculate normalized weights for display
+  const totalWeight = prompts.reduce((sum, p) => sum + p.weight, 0);
+  const normalizedWeights = prompts.map(
+    p => (totalWeight > 0 ? (p.weight / totalWeight) * 100 : 0)
+  );
+
+  const isSinglePrompt = prompts.length === 1;
+
+  // Render a single prompt field with expandable textarea
```
yondonfu, line 100: I see that by default we use Input and then switch to Textarea on focus.

IMO the visual jump from Input -> Textarea is a bit jarring. See below:

https://github.com/user-attachments/assets/08d0ecb4-6910-4369-8629-2181a028b9c5

Could we instead always use Textarea but show a smaller version with preview text and then on focus expand it to show the full text?

Also refer to screenshot of wireframes I have in mind in other comment.

lib/frame_processor.py
```diff
@@ -232,7 +234,7 @@ def process_chunk(self):
         pipeline = self.pipeline_manager.get_pipeline()
 
         # prepare() will handle any required preparation based on parameters internally
-        reset_cache = self.parameters.pop("reset_cache", None)
+        reset_cache = self.parameters.get("reset_cache", None)
```
yondonfu, line 234: I think this is causing the bug that I mentioned elsewhere. This is meant to be a one-time action right now which is why it used pop() to remove from self.parameters. This change would mean that when the frontend sends reset_cache = True then every time we run this function reset_cache = True meaning that the pipeline resets the cache every time via the prepare() call.

pipelines/streamdiffusionv2/pipeline.py
```diff
@@ -206,13 +209,11 @@ def __call__(
...
             raise ValueError("Input cannot be None for StreamDiffusionV2Pipeline")
 
-        exp_chunk_size = self.prepare(
-            prompts=prompts,
-            denoising_step_list=denoising_step_list,
-            noise_scale=noise_scale,
-            noise_controller=noise_controller,
-            manage_cache=manage_cache,
-        ).input_size
+        # Note: prepare() was already called by frame_processor before __call__
```
yondonfu, line 212: Noting that this assumption works here, but we should also be designing without the assumption of what the FrameProcessor specifically does. So, a more general assumption here would be something like "the caller must call prepare() before __call__()" and as long as that assumption holds then the next line works.

PR #55, opened 2025-10-23
Add krea-realtime-video

yondonfu:
Initial benchmarks:

**H100**

3.6 fps flash attn 2, 480x832
3.8 fps flash attn 3, 480x832
3.8 fps fp8, flash attn 2, 480x832
4 fps fp8, flash attn 3, 480x832
4.3 fps torch.compile, flash attn 2, 480x832
8.6 fps flash attn 2, 320x576

**5090 with fp8**

OOM.

**6000 Ada**

OOM without fp8, flash attn 2, 480x832
1.9fps with fp8, flash attn 2, 480x832
5.3fps with fp8, flash attn 2, 320x576

There seems to a bug that breaks things when trying to set `kv_cache_num_frames` to a value smaller than the default of 3. The reason to mess with this value is to reduce the length of the KV cache.

But, it seems that when using fp8 for the diffusion model with `kv_cache_num_frames = 1` after KV cache init VRAM consumption peaks at ~33 GB and with `kv_cache_num_frames = 3` VRAM consumption peaks at just above 40 GB (hence why we OOM on the 5090 with fp8 but do not OOM on the 6000 Ada).

---

To test the pipeline script:

```
uv run download_models.py --pipeline krea-realtime-video
uv run -m pipelines.krea_realtime_video.test
```

### DIFF
--- a/app.py
+++ b/app.py
@@ -24,6 +24,7 @@ from download_models import download_models
 from lib.models_config import get_models_dir, models_are_downloaded
 from lib.pipeline_manager import PipelineManager
 from lib.schema import (
+    HardwareInfoResponse,
     HealthResponse,
     PipelineLoadRequest,
     PipelineStatusResponse,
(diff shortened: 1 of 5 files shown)


yondonfu:
TODO: Move download for Krea into UI based workflow

PR #71, opened 2025-10-29
Ryanontheinside/feat/temporal interpolation

ryanontheinside:
Adds temporal interpolation for prompts. Based on technique from Krea realtime, generalized to all pipelines (StreamDiffusionv2, Longlive, Krea). 

- configurable transition lengths and interpolation methods
- per pipeline defaults

Next steps: integrate with timeline system.

PR #76, opened 2025-10-30
Ryanontheinside/feat/temporal blend timeline support

ryanontheinside:
Adds timeline support for temporal interpolation introduced in #71.

- Export and import timelines with temporal blend
- Support for timeline playback, with temporal information displayed
- Support for editing a prompt's temporal blending settings when a prompt block is selected in the timeline


yondonfu (changes requested):
Left a comment about observed bug with the current impl.

There are also some conflicts due to the refactor that happened in https://github.com/daydreamlive/scope/pull/87 so let's rebase as well.

yondonfu (approved):
LGTM

PR #83, opened 2025-10-31
feat: Add triton dependency for platform-specific support

BuffMcBigHuge:
Add platform-specific dependencies for triton.

Edit: I will resubmit this PR with proper requirements.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,8 @@ dependencies = [
     "torchcodec>=0.7.0",
     "torchao==0.13.0",
     "kernels>=0.10.4",
+    "triton; sys_platform != 'win32'",
+    "triton-windows; sys_platform == 'win32'",
 ]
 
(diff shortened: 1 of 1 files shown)


yondonfu:
FWIW triton is already installed specifically for Linux. Looks like torch lists it as a dep [here](https://github.com/daydreamlive/scope/blob/31cec90b3f7d06a6337be7968c158253aa7e15f0/uv.lock#L2510) and the package is listed [here](https://github.com/daydreamlive/scope/blob/31cec90b3f7d06a6337be7968c158253aa7e15f0/uv.lock#L2642) so it is included but not listed explicitly in pyproject.toml. But, since that package doesn't seem to be Windows compatible, triton-windows is still missing!

BuffMcBigHuge:
Moved to https://github.com/daydreamlive/scope/pull/98

PR #98, opened 2025-11-03
feat: Add triton dependency for platform-specific support

BuffMcBigHuge:
Triton installation in Windows. Still testing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ dependencies = [
     "torchcodec>=0.7.0",
     "torchao==0.13.0",
     "kernels>=0.10.4",
-    "triton; sys_platform != 'win32'",
+    "triton==3.4.0; sys_platform != 'win32'",
     "triton-windows==3.4.0.post21; sys_platform == 'win32'",
 ]
 


pyproject.toml
```diff
@@ -50,6 +50,8 @@ dependencies = [
     "torchcodec>=0.7.0",
     "torchao==0.13.0",
     "kernels>=0.10.4",
+    "triton; sys_platform != 'win32'",
```
yondonfu:
```suggestion
    "triton==3.4.0; sys_platform != 'win32'",
```

Technically, uv.lock already pins [triton 3.4.0](https://github.com/daydreamlive/scope/blob/29a6ff086de20711a3eed56d8c5780299e630eb0/uv.lock#L2714), but since we need to specify triton-windows explicitly here anyways also specifying triton explicitly here makes sense to me for clarity. I just suggest this change to make the version requirement clear when looking at this file - triton 3.5 only works for torch >= 2.9 and we pin torch 2.8 right now so we need to use triton 3.4 (similar to how we need to use triton-windows 3.4.0.post21).

yondonfu (changes requested):

pyproject.toml
BuffMcBigHuge: LGTM. I had set triton 3.4.0 for that very reason. I wasn't sure if the uv.lock handled this case or not.

PR #113, opened 2025-11-07
KV Attention Bias -> Cache Bias in UI

yondonfu:
- Renamed KV Attention Bias -> Cache Bias to more closely align with the naming of the other cache related params (the user of the term "cache")
- Fixed overflow issue for the pill value for Cache Bias - I could not see the value when it has a decimal places eg 0.1 on a small screen because of overflow
- Updated the tooltip description of Cache Bias
- Moved Cache Bias in the Settings panel so it is grouped with the other cache related params


ryanontheinside (approved):
lgtm

PR #117, opened 2025-11-08
fix: prevent zero-sized tensor in flex_attention padding removal

ryanontheinside:
When padded_length is 0, slicing with [:, :, :-0] creates a zero-sized tensor which causes 'amax(): Expected reduction dim 1 to have non-zero size' errors in subsequent operations.

### DIFF
--- a/pipelines/krea_realtime_video/vendor/wan2_1/modules/causal_model.py
+++ b/pipelines/krea_realtime_video/vendor/wan2_1/modules/causal_model.py
@@ -316,12 +316,15 @@ class CausalWanSelfAttention(nn.Module):
                     dim=1
                 )
 
-                x = flex_attention(
+                attn_out = flex_attention(
                     query=padded_roped_query.transpose(2, 1),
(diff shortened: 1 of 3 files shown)


pipelines/krea_realtime_video/vendor/wan2_1/modules/causal_model.py
```diff
@@ -357,17 +356,19 @@ def qkv_fn(x):
...
                     key=padded_roped_key.transpose(2, 1).contiguous(),
                     value=padded_v.transpose(2, 1).contiguous(),
                     block_mask=block_mask,
                     kernel_options={
                         "BLOCKS_ARE_CONTIGUOUS": True,
                     }
-
-                )[:, :, :-padded_length].transpose(2, 1)
+                )
+                if padded_length > 0:
```
yondonfu, line 371:
```suggestion
                if padded_length > 0:
                    x = x[:, :, :-padded_length]
                x = x.transpose(2, 1)
```

And then no need for the extra else.

Minor pref for readability wdyt?

PR #118, opened 2025-11-10
Modularize pipelines

leszko:
Modularize Krea Pipeline using Modular Diffusers.

**Changes:**
- Split `InferencePipeline` into 4 modules:
  - vae (base)
  - text_encoder (base)
  - scheduler (`KreaSchedulerWrapper`, we may try to generalize it later)
  - generator (`KreaGeneratorWrapper`, we may try to either split it further or generalize later)
- Created blocks for krea pipeline (check `modular_blocks.py` for more details)
- Moved wan2_1 general code into `wan2_1/components`

**Next Steps** (this PR or a follow-up PR(s))
1. Move logic from the `pipeline.py` into the related blocks
    - Ideally `pipeline.py` should only instantiate the related components and not do any additional logic
    - Ideally the `pipeline.py` should not even have the `prepare()` function
    - In particular, `set_timesteps.py` block does not do anything right now, but it should take the new denoising steps and apply them further
2. Untangle [Generator<>Scheduler](https://github.com/daydreamlive/scope/pull/118/files#diff-9e2f5abc387dbd1f065de29a2249fd457269a23014e531688a4fa2ecab0a5715R258), so that they are truly independent components
3. Modularize LongLive and reuse components and blocks
4. Modularize StreamDiffusion and reuse components and blocks 

**Worth mentioning**
- I added the Krea-specific logic by wrapping scheduler and generator as `KreaSchedulerWrapper` and `KreaGeneartorWrapper`, but some other options are to include this logic in diffusers blocks
- There is data passed between blocks and it's partially stored in components fields and partly as the input/output of blocks. I think it's ok, but just mentioning that we may want to revisit it at some point.

### DIFF
--- a/pipelines/blending.py
+++ b/pipelines/blending.py
@@ -328,3 +328,9 @@ class EmbeddingBlender:
         if self._state == BlenderState.TRANSITIONING:
             self._transition_queue.clear()
             self._state = BlenderState.IDLE
+
+    def reset(self) -> None:
+        """Fully reset temporal state for fresh sessions (e.g., after cache reset)."""
+        self._transition_queue.clear()
+        self._current_blend_embedding = None
+        self._state = BlenderState.IDLE
--- a/pipelines/wan2_1/blocks/embedding_blending.py
+++ b/pipelines/wan2_1/blocks/embedding_blending.py
@@ -116,6 +116,12 @@ class EmbeddingBlendingBlock(ModularPipelineBlocks):
     def __call__(self, components, state: PipelineState) -> tuple[Any, PipelineState]:
         block_state = self.get_block_state(state)
 
+        if state.get("init_cache", False):
+            components.embedding_blender.reset()
+            block_state._previous_embeds_signature = None
+            block_state.conditioning_embeds = None
+            block_state.current_prompts = None
+
         # Get inputs from state
         embeds_list = block_state.embeds_list
         embeds_weights = block_state.embeds_weights
--- a/pipelines/wan2_1/blocks/text_conditioning.py
+++ b/pipelines/wan2_1/blocks/text_conditioning.py
@@ -118,6 +118,11 @@ class TextConditioningBlock(ModularPipelineBlocks):
         block_state = self.get_block_state(state)
         self.check_inputs(block_state)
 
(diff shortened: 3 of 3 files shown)


pipelines/wan2_1/blocks/prepare_video_latents.py
```diff
@@ -71,6 +71,11 @@ def intermediate_outputs(self) -> list[OutputParam]:
                 description="Noisy latents to denoise",
             ),
             OutputParam("generator", description="Random number generator"),
+            OutputParam(
```
yondonfu: What if the `video` param is only used for the input vid and we use an `output` param for the output vid? Then, you could just check if `video` exists in SetupCachesBlock.

pipelines/streamdiffusionv2/modular_blocks.py
```diff
@@ -21,9 +21,9 @@
 ALL_BLOCKS = InsertableDict(
     [
         ("text_conditioning", TextConditioningBlock),
-        ("set_timesteps", SetTimestepsBlock),
         ("preprocess_video", PreprocessVideoBlock),
         ("noise_scale_controller", NoiseScaleControllerBlock),
+        ("set_timesteps", SetTimestepsBlock),
```
yondonfu: Was the problem that SetTimestepsBlock uses the noise_scale param which needs to be set by NoiseScaleControllerBlock first? Or something else?

leszko: Yes, that was the problem

pipelines/wan2_1/blocks/prepare_video_latents.py
leszko: Yeah, I think it's a good idea.

1. Maybe then call it `input_video` to avoid confusion? So we would have `input_video` for input video and `output` for the output video. Wdyt?
2. It tried to quickly change, but have some weird error, so probably need to take a closer look. I think it's ok if you merge it as it is into `main` and I'll do this refactor later.

yondonfu: 1. Would just do `input_video` and `output_video` then.
2. 👍

PR #131, opened 2025-11-13
Ryanontheinside/feat/blending on modular krea

ryanontheinside:
This PR adds a generic block for blending embeddings, restoring prompt spatial and temporal blending in the new modular block architecture. 

Next steps: 
- refine and optimize this implementation
- revisit once additional conditioning inputs are supported by the system; possibly supplant with hyper-generic 'conditioning' blender.

### DIFF
--- a/pipelines/krea_realtime_video/pipeline.py
+++ b/pipelines/krea_realtime_video/pipeline.py
@@ -169,4 +169,4 @@ class KreaRealtimeVideoPipeline(Pipeline):
             self.state.set("denoising_step_list", DEFAULT_DENOISING_STEP_LIST)
 
         _, self.state = self.blocks(self.components, self.state)
-        return postprocess_chunk(self.state.values["video"])
+        return postprocess_chunk(self.state.values["output_video"])
--- a/pipelines/krea_realtime_video/blocks/prepare_context_frames.py
+++ b/pipelines/krea_realtime_video/blocks/prepare_context_frames.py
@@ -28,7 +28,7 @@ class PrepareContextFramesBlock(ModularPipelineBlocks):
                 description="Denoised latents",
             ),
             InputParam(
-                "video",
+                "output_video",
                 required=True,
                 type_hint=torch.Tensor,
                 description="Decoded video frames",
@@ -109,7 +109,7 @@ class PrepareContextFramesBlock(ModularPipelineBlocks):
             block_state.decoded_frame_buffer = torch.cat(
                 [
                     block_state.decoded_frame_buffer,
-                    block_state.video.to(
+                    block_state.output_video.to(
                         block_state.decoded_frame_buffer.device,
                         block_state.decoded_frame_buffer.dtype,
                     ),
--- a/pipelines/longlive/pipeline.py
+++ b/pipelines/longlive/pipeline.py
@@ -137,4 +137,4 @@ class LongLivePipeline(Pipeline):
             self.state.set("denoising_step_list", DEFAULT_DENOISING_STEP_LIST)
 
         _, self.state = self.blocks(self.components, self.state)
-        return postprocess_chunk(self.state.values["video"])
+        return postprocess_chunk(self.state.values["output_video"])
--- a/pipelines/streamdiffusionv2/pipeline.py
+++ b/pipelines/streamdiffusionv2/pipeline.py
@@ -136,4 +136,4 @@ class StreamDiffusionV2Pipeline(Pipeline):
             self.state.set("denoising_step_list", DEFAULT_DENOISING_STEP_LIST)
 
         _, self.state = self.blocks(self.components, self.state)
-        return postprocess_chunk(self.state.values["video"])
+        return postprocess_chunk(self.state.values["output_video"])
--- a/pipelines/wan2_1/blocks/decode.py
+++ b/pipelines/wan2_1/blocks/decode.py
@@ -38,7 +38,7 @@ class DecodeBlock(ModularPipelineBlocks):
     def intermediate_outputs(self) -> list[OutputParam]:
         return [
             OutputParam(
-                "video",
+                "output_video",
                 type_hint=torch.Tensor,
                 description="Decoded video frames",
             ),
@@ -51,7 +51,7 @@ class DecodeBlock(ModularPipelineBlocks):
         # Decode to pixel space
         video = components.vae.decode_to_pixel(block_state.latents, use_cache=True)
 
-        block_state.video = video
+        block_state.output_video = video
 
         self.set_block_state(state, block_state)
         return components, state
--- a/pipelines/wan2_1/blocks/prepare_video_latents.py
+++ b/pipelines/wan2_1/blocks/prepare_video_latents.py
@@ -71,11 +71,6 @@ class PrepareVideoLatentsBlock(ModularPipelineBlocks):
                 description="Noisy latents to denoise",
             ),
             OutputParam("generator", description="Random number generator"),
-            OutputParam(
-                "input_video",
-                type_hint=bool,
-                description="Whether input video is present",
-            ),
         ]
 
     @torch.no_grad()
@@ -135,7 +130,6 @@ class PrepareVideoLatentsBlock(ModularPipelineBlocks):
             1 - block_state.noise_scale
         )
         block_state.generator = rng
-        block_state.input_video = True
 
         self.set_block_state(state, block_state)
         return components, state
--- a/pipelines/wan2_1/blocks/setup_caches.py
+++ b/pipelines/wan2_1/blocks/setup_caches.py
@@ -131,11 +131,10 @@ class SetupCachesBlock(ModularPipelineBlocks):
             init_cache = True
 
         # Clear KV cache when conditioning changes, if manage_cache is enabled and video input is present
-        input_video = getattr(block_state, "input_video", False)
         if (
             block_state.conditioning_embeds_updated
             and block_state.manage_cache
-            and input_video
+            and block_state.video is not None
         ):
             init_cache = True
 


lib/frame_processor.py
```diff
@@ -228,6 +228,16 @@ def process_chunk(self):
...
+        # For pipelines that support transitions, treat transition target prompts
+        # as the next prompts. This lets embedding blending handle transitions
+        # based on old vs new conditioning without needing a separate target
+        # prompts path.
+        transition = self.parameters.get("transition")
+        if transition is not None:
+            target_prompts = transition.get("target_prompts") or None
+            if target_prompts:
+                self.parameters["prompts"] = target_prompts
+
```
ryanontheinside: im not overly thrilled with this

lib/frame_processor.py
```diff
@@ -245,9 +255,7 @@ def process_chunk(self):
             requirements = pipeline.prepare(
                 **self.parameters,
             )
-
-        # Transition is consumed by prepare()
-        self.parameters.pop("transition", None)
+        self.is_prepared = True
```
yondonfu: Is this necessary? We set this to True on L341.

pipelines/wan2_1/blocks/text_conditioning.py
```diff
@@ -20,83 +23,129 @@ def expected_components(self) -> list[ComponentSpec]:
...
-                description="New prompts to condition denoising",
+                type_hint=list[dict],
+                description="Prompts to condition denoising as list of dicts with 'text' and 'weight' keys",
             ),
             InputParam(
                 "prompt_embeds",
                 type_hint=torch.Tensor,
                 description="Text embeddings to condition denoising",
             ),
+            InputParam(
```
yondonfu, line 49: If this is not used as an input in this block I think it can be removed.

pipelines/wan2_1/blocks/text_conditioning.py
```diff
@@ -20,83 +23,129 @@ def expected_components(self) -> list[ComponentSpec]:
...
-                type_hint=str | list[str] | None,
+                type_hint=list[dict] | None,
                 description="Current prompts conditioning denoising",
             ),
             InputParam(
                 "prompts",
                 required=True,
-                type_hint=str | list[str],
-                description="New prompts to condition denoising",
+                type_hint=list[dict],
```
yondonfu: It was nice being able to just pass in a string from a usability POV (as opposed to needing to pass in a dict with weight set to 100% just to use a single prompt when not doing any blending). Could we also support a string input?

pipelines/wan2_1/blocks/text_conditioning.py
```diff
@@ -20,83 +23,129 @@ def expected_components(self) -> list[ComponentSpec]:
...
+            ) -> tuple[list[torch.Tensor], list[float]]:
+                """Encode a list of prompt dicts into embeddings and weights."""
+                embeddings: list[torch.Tensor] = []
+                weights: list[float] = []
+
+                for prompt_item in prompt_items:
+                    text = prompt_item.get("text", "")
+                    weight = prompt_item.get("weight", 1.0)
+
+                    conditional_dict = components.text_encoder(text_prompts=[text])
```
yondonfu: Does the text encoder support batch generation of embeddings eg. pass in a list of prompts and get back a list of embeddings?

Something like:

```
texts, weights = zip(*((item["text"], item["weight"]) for item in prompt_items))
embeds = components.text_encoder(text_prompts=texts)
```

Might be more efficient if the text encoder can do batch generation and returns what we need.

pipelines/wan2_1/blocks/text_conditioning.py
```diff
@@ -20,83 +23,129 @@ def expected_components(self) -> list[ComponentSpec]:
...
+        block_state.embedding_weights = None
 
-        # Only run text_encoder if prompt changed
-        if (
+        # Check if prompts changed
+        prompts_changed = (
             block_state.current_prompts is None
             or block_state.current_prompts != block_state.prompts
+        )
+        block_state.conditioning_changed = prompts_changed
```
yondonfu: Why use this vs just signal with `prompt_embeds_updated`?

pipelines/wan2_1/blocks/text_conditioning.py
```diff
@@ -20,83 +23,129 @@ def expected_components(self) -> list[ComponentSpec]:
...
-                "prompt_embeds",
-                type_hint=torch.Tensor,
-                description="Text embeddings to condition denoising",
+                "embeds_list",
+                type_hint=list[torch.Tensor] | None,
+                description="List of individual embeddings for blending (when prompts is list[dict])",
             ),
             OutputParam(
-                "prompt_embeds_updated",
+                "embedding_weights",
```
yondonfu:
```suggestion
                "embeds_weights",
```

lib/frame_processor.py
leszko: Cannot we handle this in a block? Either text_conditioning or create a new one for the transition?

lib/frame_processor.py
```diff
@@ -268,18 +276,17 @@ def process_chunk(self):
...
             # Pass video input to pipeline
             if video_input is not None:
                 call_params["video"] = video_input
 
             output = pipeline(**call_params)
 
+            # Clear transition after processing - it's consumed by modular pipelines
+            # Note: Transitions only work with modular pipelines (StreamDiffusionV2, LongLive, Krea)
+            # Non-modular pipelines (Passthrough, VOD) ignore this parameter
+            self.parameters.pop("transition", None)
```
leszko: The same question as above, cannot we handle this in one of the blocks?

pipelines/krea_realtime_video/pipeline.py
```diff
@@ -19,7 +20,7 @@
 DEFAULT_DENOISING_STEP_LIST = [1000, 750, 500, 250]
 
 WARMUP_RUNS = 3
-WARMUP_PROMPT = "a majestic sunset"
+WARMUP_PROMPT = [{"text": "a majestic sunset", "weight": 1.0}]
```
leszko, line 23: We may want to consider accepting both formats:
- pure text
- JSON

I think we should use JSON everywhere in the app, but it would nice if the pipeline API user would be able to just pass a prompt as a text it would just work.

pipelines/wan2_1/blocks/embedding_blending.py
```diff
@@ -0,0 +1,198 @@
...
+    InputParam,
+    OutputParam,
+)
+
+from ...blending import EmbeddingBlender, parse_transition_config
+
+logger = logging.getLogger(__name__)
+
+
+class EmbeddingBlendingBlock(ModularPipelineBlocks):
```
leszko, line 17: This is great to see the blending stuff as a block 🚀

pipelines/wan2_1/blocks/embedding_blending.py
```diff
@@ -0,0 +1,198 @@
...
+                type_hint=dict | None,
+                description="Optional transition config for temporal blending",
+            ),
+            InputParam(
+                "prompt_embeds",
+                type_hint=torch.Tensor,
+                description="Existing prompt embeddings (optional)",
+            ),
+            InputParam(
+                "conditioning_changed",
```
leszko: This looks similar to `prompt_embeds_updated`, something to consider is if cannot reuse it or make one state variable to decide if anything about the prompts have changed.

pipelines/wan2_1/blocks/text_conditioning.py
leszko: +1

pipelines/wan2_1/blocks/text_conditioning.py
ryanontheinside: implemented

pipelines/wan2_1/blocks/text_conditioning.py
ryanontheinside: since y'all insist haha

lib/frame_processor.py
ryanontheinside: see 14b990d1cb66d270f3b744fc6591fe0c883c2297

it is a compromise due to arch constraints that we can potentially address later

yondonfu:
A few notes from testing:

- I confirmed that setting transition steps to > 0 results in the prompt change taking longer to have a visual effect with longlive + krea
- I confirmed that setting transition steps to 0 results in the visual effect of a prompt change happening quicker with longlive + krea
- I confirmed that spatial blending of prompts with different weights has the expected visual effect with longlive + streamdiffusionv2

I also noticed while testing streamdiffusionv2 that when using the timeline JSON file for [this example](https://github.com/daydreamlive/scope/blob/main/pipelines/streamdiffusionv2/docs/usage.md#examples) I get different visual effects now. It seems that the default behavior is no longer resetting cache on prompt change? I think @leszko had pushed something to keep that behavior in https://github.com/daydreamlive/scope/pull/118/commits/c65f9b0af80f2c8a0b8871b437a421155d13fa2d but it doesn't look like it is working.

This does bring up the question of what the expected behavior should be when using temporal transitions because if the KV cache is reset then the temporal transition stuff doesn't do anything because the KV cache is reset?

Perhaps:
- When transition steps = 0 a prompt change leads to auto cache reset which matches `main` default behavior
- When transition steps > 0 a prompt change does not lead to auto cache reset

WDYT?

An aside, I created https://github.com/daydreamlive/scope/issues/133 to track some UX improvements that came to mind while testing.

pipelines/streamdiffusionv2/modular_blocks.py
```diff
@@ -21,9 +22,10 @@
 ALL_BLOCKS = InsertableDict(
     [
         ("text_conditioning", TextConditioningBlock),
+        ("embedding_blending", EmbeddingBlendingBlock),
+        ("set_timesteps", SetTimestepsBlock),
```
yondonfu, line 26: I think @leszko had said that SetTimesteps needs to be after NoiseScaleController in https://github.com/daydreamlive/scope/pull/118/commits/d791f422e03d52f360c7c6c73cbc28ec45aa3c83. Maybe related to streamdiffusionv2 behavior I mentioned in other comment.

pipelines/wan2_1/blocks/text_conditioning.py
```diff
@@ -48,55 +60,120 @@ def intermediate_outputs(self) -> list[OutputParam]:
...
+            # Encode regular prompts if they changed
+            if prompts_changed:
+                (
+                    block_state.embeds_list,
+                    block_state.embeds_weights,
+                ) = encode_prompt_items(normalized_prompts)
+
+                # Store prompts_to_encode (which may be target_prompts) for comparison
+                # This ensures we track what was actually encoded
+                block_state.current_prompts = prompts_to_encode
```
yondonfu: I feel like the following might be more intuitive given that `current_prompts` is supposed to represent the current prompts in use while `prompts` is supposed to represent the new/next prompts to use.

1. TextConditionBlock sets `prompts` to the prompts to encode (which may be `transition.target_prompts`.
2. EmbeddingBlendingBlock executes the transition (if provided) each time it is called and then when the transition is complete (or if there is no transition) then it would set `current_prompts` to `prompts`.

There would be something like this in EmbeddingBlending block:

```
if not is_transitioning:
  block_state.current_prompts = block_state.prompts
```

And this might allow us to avoid the logic in FrameProcessor for setting prompts to target_prompts?

lib/frame_processor.py
```diff
@@ -210,6 +210,14 @@ def process_chunk(self):
             # Check if there are new parameters
             new_parameters = self.parameters_queue.get_nowait()
             if new_parameters != self.parameters:
+                # Clear stale transition when new prompts arrive without transition
```
yondonfu, line 213: I suspect we can simplify things here and wrote up some thoughts in https://github.com/daydreamlive/scope/issues/134.

It represents a bigger change though so rather than make smaller changes here to try to have cleaner logic my thinking is we can do a follow up for that issue outside of this PR.

ryanontheinside: I bet we are on the same page. I alluded to this in 14b990d1cb66d270f3b744fc6591fe0c883c2297

yondonfu: Yeah just wanted to make a note and since it requires some architectural changes figured can track in that issue and follow up separatley.

pipelines/streamdiffusionv2/modular_blocks.py
ryanontheinside: d791f422e03d52f360c7c6c73cbc28ec45aa3c83 actually seems to introduce unwanted behavior - the ordering in my branch behaves more correctly. Reversing them causes cache reset on every frame.

yondonfu: Fixed in https://github.com/daydreamlive/scope/pull/131/commits/88405f9a9321da650d7cf3336aa0898e1df26a9c and https://github.com/daydreamlive/scope/pull/131/commits/de63150ddd5e2bc36d6fb818ae6d239dab85e9ce

PR #137, opened 2025-11-15
Fixes to denoising step list cache init and conditioning change cache init

yondonfu:
The first problem that we had was that the NoiseScaleControllerBlock would potentially change the noise_scale param ever call and that would trigger a mutation to denoising_step_list in SetTimestepsBlock every call. As a result, if manage_cache = True, init_cache would be set to True every call resulting in a cache reset in SetupCachesBlock every call. 

The change in https://github.com/daydreamlive/scope/pull/118/commits/d791f422e03d52f360c7c6c73cbc28ec45aa3c83 might've addressed some other strange behavior, but the block ordering change actually causes this specific problem because the NoiseScaleControllerBlock running first means that it can change the noise_scale value read in SetTimestepsBlock every call.

I generally dislike how easy it is to shoot yourself in the foot by updating/reading from "global state" eg the PipelineState in these modular blocks and I think this property makes it more likely for these types of bugs to occur. If you read a param from state and then mutate it in a block it is very easy for that mutation to inadvertently affect other blocks. In this case, the inadvertent mutation happened with denoising_step_list since SetTimestepsBlocks was using noise_scale to directly update the step at the first index. 

The current fix in https://github.com/daydreamlive/scope/pull/137/commits/e53d24373a0a2ce5bf8df213b776a87ab24c09db involves the following:

- Do not mutate denoising_step_list and treat it as user input that should not be updated and instead only update current_denoising_step_list. When we set current_denoising_step_list we make sure to set it to a *clone* of denoising_step_list.
- Move the logic for applying noise_scale to current_denoising_step_list to the DenoiseBlock and instead of updating current_denoising_step_list we clone current_denoising_step_list and update the first index of the cloned tensor and use that in the denoising loop. This makes sure that we aren't constantly updating current_denoising_step_list every call and that it is always referencing a value that is copied from user input.

After this fix, the relative ordering of NoiseScaleControllerBlock + SetTimestepsBlock does not matter anymore because the noise_scale value set by NoiseScaleControllerBlock is used in DenoiseBlock.

As a part of this fix, I also added some comments denoting parameters that can be treated as "intermediate inputs". The modular diffusers [docs](https://huggingface.co/docs/diffusers/v0.35.1/modular_diffusers/pipeline_block) mention this, but it doesn't seem like they actually support denoting them explicitly in the v0.35.2 version (in a quick test I was unable to read intermediate inputs from block state). The basic idea is that intermediate inputs denote inputs that are typically passed by other blocks and they are expected to be modified while regular inputs are not expected to be modified. I don't think there is an easy way to actually enforce these rules at a code level right now (maybe we can add a thin layer) and since the intermediate inputs property doesn't actually work right now (it seems) I just added some comments highlighting which params can be thought of as intermediate inputs.

The second problem is that there was a bug in https://github.com/daydreamlive/scope/pull/118/commits/c65f9b0af80f2c8a0b8871b437a421155d13fa2d.

The fix in https://github.com/daydreamlive/scope/pull/137/commits/6a8ba38dc5b706020339b229b71948aa495df010 is to use `video` to represent input video that is not modified and use `output_video` to represent output video and then check for the presence of `video` in SetupCachesBlock.

### DIFF
--- a/pipelines/krea_realtime_video/blocks/prepare_context_frames.py
+++ b/pipelines/krea_realtime_video/blocks/prepare_context_frames.py
@@ -28,7 +28,7 @@ class PrepareContextFramesBlock(ModularPipelineBlocks):
                 description="Denoised latents",
             ),
             InputParam(
-                "video",
+                "output_video",
                 required=True,
                 type_hint=torch.Tensor,
                 description="Decoded video frames",
@@ -109,7 +109,7 @@ class PrepareContextFramesBlock(ModularPipelineBlocks):
             block_state.decoded_frame_buffer = torch.cat(
                 [
                     block_state.decoded_frame_buffer,
-                    block_state.video.to(
+                    block_state.output_video.to(
                         block_state.decoded_frame_buffer.device,
                         block_state.decoded_frame_buffer.dtype,
                     ),
--- a/pipelines/krea_realtime_video/pipeline.py
+++ b/pipelines/krea_realtime_video/pipeline.py
@@ -169,4 +169,4 @@ class KreaRealtimeVideoPipeline(Pipeline):
             self.state.set("denoising_step_list", DEFAULT_DENOISING_STEP_LIST)
 
         _, self.state = self.blocks(self.components, self.state)
-        return postprocess_chunk(self.state.values["video"])
+        return postprocess_chunk(self.state.values["output_video"])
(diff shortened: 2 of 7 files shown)


leszko:
I think that the worse thing is that the Modular Diffusers framework does not have any checks if you actually do not modify the state other than mentioned in the output params.

I fired a PR to check this and apply minor corrections: https://github.com/daydreamlive/scope/pull/146
I think it addresses Points 1-3 from @BuffMcBigHuge 
The other points, I think we'll need to keep in mind, but probably not apply in the https://github.com/daydreamlive/scope/pull/118 PR

PR #141, opened 2025-11-15
Ryanontheinside/feat/LoRA

ryanontheinside:
## Summary

Add end-to-end LoRA adapter support for WAN-based pipelines, including multi-lora support with independent scales, configurable merge strategies, runtime scale updates, timeline export support, and a UI for selecting and controlling LoRAs.

## Backend

- Introduce LoRA-aware load parameters via `LoRAEnabledLoadParams` and update `StreamDiffusionV2`, `LongLive`, and `KreaRealtimeVideo` load schemas to accept:
  - `loras: [{ path, scale }]`
  - `lora_merge_mode: "permanent_merge" | "runtime_peft"`
- Add `LoRAMergeMode`, `LoRAConfig`, and `LoRAScaleUpdate` to `schema.py` and extend `Parameters` with `lora_scales` for runtime updates.
- Implement `LoRAEnabledPipeline` mixin that:
  - Initializes LoRAs once per pipeline using a shared `LoRAManager`.
  - Handles per-frame `lora_scales` updates for strategies that support them.
  - Exposes `loaded_lora_adapters` for UI/runtime introspection.
- Implement `pipelines/wan2_1/lora` subsystem:
  - `LoRAManager` façade dispatching to strategy implementations.
  - `PermanentMergeLoRAStrategy`: one-time weight merging at load for max FPS, no runtime updates.
  - `PeftLoRAStrategy`: PEFT-based LoRA with instant runtime scale updates; optional `torch.compile` via `PEFT_LORA_TORCH_COMPILE` env flag [[this needs testing for efficacy]]
  - `ModuleTargetedLoRAStrategy`: replicates and supplants LongLive's original LoRA loading behavior, targeting specific module types (CausalWanAttentionBlock/WanAttentionBlock) for compatibility with existing LongLive LoRA files.
  - Shared utilities for parsing LoRA state dicts and mapping them onto model params (`utils.py`).
- Wire the mixin into WAN-based pipelines:
  - `StreamDiffusionV2Pipeline`, `LongLivePipeline`, `KreaRealtimeVideoPipeline` now all call `_init_loras` at construction and `_handle_lora_scale_updates` in `_generate`.
- Extend `PipelineManager`:
  - Add `_apply_load_params` helper to apply `height`/`width`/`seed`/`loras`/`lora_merge_mode`.
  - Include `loaded_lora_adapters` in `/api/v1/models/status` so the frontend can see which LoRAs are actually loaded.
- Add `/api/v1/lora/list` endpoint to enumerate LoRA files under `models/lora/**` (name, path, size, folder).

## Frontend

- Extend settings state and API types:
  - Add `LoRAConfig` and `LoraMergeStrategy` to `frontend/src/types`.
  - Add `loras` and `lora_merge_mode` to relevant load param types and pipeline status in `frontend/src/lib/api.ts`.
  - Persist `settings.loras` and `settings.loraMergeStrategy` in `useStreamState`.
- Add LoRA management UI:
  - New `LoRAManager` component for:
    - Adding/removing LoRA entries.
    - Picking LoRA files from a grouped tree of discovered adapters.
    - Editing scale locally with a slider + "Apply" semantics (with streaming-time behavior depending on merge strategy).
  - New reusable `FilePicker` component for folder-grouped file selection.
  - Expose LoRA controls in `SettingsPanel` for LoRA-capable pipelines:
    - `LoRAManager` section.
    - `LoRA Strategy` select with tooltips explaining `Permanent Merge` vs `Runtime PEFT`.
- Streaming/runtime behavior:
  - `StreamPage`:
    - Clears `settings.loras` on pipeline change.
    - Uses `buildLoRAParams` to send `loras` + `lora_merge_mode` when loading LoRA-enabled pipelines.
    - On `loras` change while streaming, calls `sendLoRAScaleUpdates`:
      - Filters to LoRAs that are actually loaded (based on `pipelineInfo.loaded_lora_adapters`).
      - Sends `lora_scales` over the data channel for runtime updates.
- Capability wiring:
  - `PIPELINES` metadata extended with `supportsLoRA` and helper `pipelineSupportsLoRA`.
  - LoRA UI is only shown when the active pipeline declares `supportsLoRA: true`.

## API / UX Notes

- `Permanent Merge` mode:
  - LoRAs are merged once at load, no runtime scale updates.
  - Maximum FPS, best suited for fixed configurations / production.
- `Runtime PEFT` mode:
  - LoRAs are attached via PEFT `LoraLayer`.
  - Scale updates are instant during streaming, at the cost of extra per-frame overhead.
- LoRA file discovery is driven by the backend `/api/v1/lora/list` endpoint, so the frontend has no hardcoded knowledge of which adapters exist; it simply lists whatever is present under `models/lora`.

## Next steps: 
[Experimental lora merge](https://github.com/daydreamlive/scope/blob/ryanontheinside/experiments/lora-experimental-runtime-strategies/pipelines/krea_realtime_video/test_lora_merge_modes.py) strategies (so-called GPU Reconstruct showing promise) were abandoned for expediency, and can optionally be pursued at a later date.

### DIFF
--- a/frontend/src/components/LoRAManager.tsx
+++ b/frontend/src/components/LoRAManager.tsx
@@ -1,7 +1,7 @@
 import { useState, useEffect } from "react";
 import { Button } from "./ui/button";
 import { SliderWithInput } from "./ui/slider-with-input";
-import { Plus, X, RefreshCw, Check } from "lucide-react";
+import { Plus, X, RefreshCw } from "lucide-react";
 import {
   Tooltip,
   TooltipContent,
@@ -56,14 +56,6 @@ export function LoRAManager({
     setLocalScales(newLocalScales);
   }, [loras]);
 
-  // Check if there are unsaved scale changes
-  const hasUnsavedScales = loras.some(lora => {
-    const localScale = localScales[lora.id];
-    return (
-      localScale !== undefined && Math.abs(localScale - lora.scale) > 0.001
-    );
-  });
-
   const handleAddLora = () => {
     const newLora: LoRAConfig = {
       id: crypto.randomUUID(),
@@ -87,13 +79,8 @@ export function LoRAManager({
     setLocalScales(prev => ({ ...prev, [id]: scale }));
   };
 
-  const handleApplyScales = () => {
-    onLorasChange(
-      loras.map(lora => ({
-        ...lora,
-        scale: localScales[lora.id] ?? lora.scale,
-      }))
-    );
+  const handleScaleCommit = (id: string, scale: number) => {
+    handleLoraChange(id, { scale });
   };
 
   return (
@@ -178,6 +165,9 @@ export function LoRAManager({
                         onValueChange={value => {
                           handleLocalScaleChange(lora.id, value);
                         }}
+                        onValueCommit={value => {
+                          handleScaleCommit(lora.id, value);
+                        }}
                         min={-10}
                         max={10}
                         step={0.1}
@@ -196,7 +186,7 @@ export function LoRAManager({
                     <p className="text-xs">
                       {isStreaming && loraMergeStrategy === "permanent_merge"
                         ? "Runtime adjustment is disabled with Permanent Merge strategy. LoRA scales are fixed at load time."
-                        : "Adjust LoRA strength. Click Apply to update during streaming. 0.0 = no effect, 1.0 = full strength"}
+                        : "Adjust LoRA strength. Updates automatically when you release the slider or use +/- buttons. 0.0 = no effect, 1.0 = full strength"}
                     </p>
                   </TooltipContent>
                 </Tooltip>
@@ -205,20 +195,6 @@ export function LoRAManager({
           </div>
         ))}
       </div>
-
-      {hasUnsavedScales && (
-        <Button
-          size="sm"
-          onClick={handleApplyScales}
-          disabled={
-            disabled || (isStreaming && loraMergeStrategy === "permanent_merge")
-          }
-          className="w-full"
-        >
-          <Check className="h-3 w-3 mr-1" />
-          Apply Scale Changes
-        </Button>
-      )}
     </div>
   );
 }


frontend/src/components/LoRAManager.tsx
```diff
@@ -0,0 +1,222 @@
...
+                    </p>
+                  </TooltipContent>
+                </Tooltip>
+              </TooltipProvider>
+            </div>
+          </div>
+        ))}
+      </div>
+
+      {hasUnsavedScales && (
```
yondonfu: CleanShot 2025-11-19 at 11 43 57

I found it a bit cumbersome to have to press another button after moving the slider. Can we just send the new scale value when the slider thumb stops like what we do for other sliders?

frontend/src/components/LoRAManager.tsx
```diff
@@ -0,0 +1,222 @@
...
+              </Button>
+            </div>
+
+            <div className="flex items-center gap-2">
+              <span className="text-xs text-muted-foreground w-12">Scale:</span>
+              <TooltipProvider>
+                <Tooltip>
+                  <TooltipTrigger asChild>
+                    <div className="flex-1">
+                      <SliderWithInput
```
yondonfu, line 163: CleanShot 2025-11-19 at 11 45 31

There is weird behavior here where after incrementing the slider with + a few times the value shown has many decimal places. I would expect pressing + a few times to just lead to 1.1, 1.2, etc. instead of showing all those decimal places which felt disorienting.

pipelines/wan2_1/lora/mixin.py
```diff
@@ -0,0 +1,139 @@
...
+
+        logger.info(
+            "_init_loras: Completed, loaded %d LoRA adapters",
+            len(self.loaded_lora_adapters),
+        )
+
+        # Return the original model (PEFT wrapping is handled internally by runtime_peft strategy)
+        return model
+
+    def _handle_lora_scale_updates(
```
yondonfu, line 117: WDYT about treating LoRA scale updates in the same way that we treat other parameter updates that affect visual quality (eg denoising_step_list, noise_scale) and trigger a cache reset on update?

frontend/src/components/SettingsPanel.tsx
```diff
@@ -596,6 +605,76 @@ export function SettingsPanel({
...
+                          <TooltipTrigger asChild>
+                            <SelectItem value="permanent_merge">
+                              Permanent Merge
+                            </SelectItem>
+                          </TooltipTrigger>
+                          <TooltipContent side="right" className="max-w-xs">
+                            <p className="text-xs">
+                              Maximum performance, no runtime updates. LoRA
+                              scales are permanently merged into model weights
+                              at load time. Ideal for production when scales are
```
yondonfu:
```suggestion
                              at load time. Ideal for when you already know what scale to use
```

yondonfu: ^ in addition to cutting the next line.

Rationale: The user won't necessarily know what production means in this context.

frontend/src/components/LoRAManager.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+      <div className="space-y-2">
+        {loras.map(lora => (
+          <div
+            key={lora.id}
+            className="rounded-lg border bg-card p-3 space-y-2"
+          >
+            <div className="flex items-center justify-between gap-2">
+              <div className="flex-1">
+                <FilePicker
```
yondonfu, line 132: CleanShot 2025-11-19 at 12 18 18

There is horizontal overflow on screens with less horizontal width.

pipelines/krea_realtime_video/model.yaml
```diff
@@ -7,3 +7,6 @@ local_attn_size: 6
 vae_spatial_downsample_factor: 8
 patch_embedding_spatial_downsample_factor: 2
 max_rope_freq_table_seq_len: 1024
+# LoRA merge mode: runtime_peft (instant updates, overhead), gpu_reconstruct (slow updates, no overhead),
+# permanent_merge (no updates, zero overhead), cuda_graph_recapture (CUDA graph + PEFT)
+lora_merge_mode: runtime_peft
```
yondonfu: Is this needed?

pipelines/wan2_1/lora/strategies/peft_lora.py
```diff
@@ -0,0 +1,609 @@
+"""
+PEFT-based LoRA manager for WAN models with real-time scale updates.
+
+This implementation uses PEFT's LoraLayer for runtime LoRA application
+without weight merging, enabling instant scale updates (<1s) suitable for
+real-time video generation with FP8 quantization.
```
yondonfu:
```suggestion
real-time video generation.
```

frontend/src/components/SettingsPanel.tsx
```diff
@@ -596,6 +605,76 @@ export function SettingsPanel({
             </div>
           </div>
         )}
+
+        {pipelineSupportsLoRA(pipelineId) && (
```
yondonfu: Can we move this to the higher in the Settings panel below the pipeline card and above the height/width parameters? While we're at it we can remove the Parameters heading too to save vertical space.

Rationale: To give more prominence to LoRAs.

CleanShot 2025-11-19 at 13 10 48

pipelines/krea_realtime_video/pipeline.py
```diff
@@ -158,6 +162,13 @@ def __call__(self, **kwargs) -> torch.Tensor:
         return self._generate(**kwargs)
 
     def _generate(self, **kwargs) -> torch.Tensor:
+        # Handle runtime LoRA scale updates before writing into state.
+        lora_scales = kwargs.get("lora_scales")
+        if lora_scales is not None:
+            self._handle_lora_scale_updates(
```
yondonfu, line 168: This mixin approach works now, but I'm interested in whether we can move this into a block later.

pipelines/wan2_1/lora/manager.py
```diff
@@ -0,0 +1,144 @@
...
+      + Maximum inference performance (zero overhead)
+      - No runtime scale updates
+
+    - runtime_peft: Uses PEFT LoraLayer for runtime application
+      + Instant scale updates (<1s)
+      - ~50% inference overhead per frame
+
+    - module_targeted: Targets specific module types (like LongLive)
+      + Compatible with existing module-driven LoRA files
+      - Uses PEFT wrapping with runtime scale updates
```
yondonfu: The strategy impl for this notes that runtime scale updates aren't supported though?

pipelines/wan2_1/lora/mixin.py
```diff
@@ -0,0 +1,139 @@
...
+
+def _get_config_value(config: Any, key: str, default: Any = None) -> Any:
+    """Get value from config whether it's dict-like or object-like."""
+    if hasattr(config, "get"):
+        return config.get(key, default)
+    return getattr(config, key, default)
+
+
+class LoRAEnabledPipeline:
+    """Shared LoRA integration for WAN-based pipelines.
```
yondonfu, line 32: Just to clarify (because I haven't gone through line by line in the impl) - the LoRA integration is specific to the Wan2.1 architecture?

ryanontheinside:
> A UX pain point that I have on a remote GPU (eg RunPod) setup:
> 
> The easiest way to download a LoRA from Civitai or HF is to download it to my computer. You could use wget/curl to download the LoRA from HF, but Civitai is gated behind auth now it seems so signing in in-browser and downloading to my computer is easier than trying to get an API key and then downloading via CLI.
> 
> The issue is that if the server is running on a remote machine it doesn't have access to my local filesystem. I have the LoRA on my computer, but now I need to get it to the remote machine.

This is definitely annoying - I had the same experience using Comfy on remote servers. 

Civitai downloads can be done with wget, but you must have an account to obtain an [auth token](https://education.civitai.com/civitais-guide-to-downloading-via-api/). 

Something to consider, is that ComfyUI recently implemented a feature that helps to download missing models found in a workflow. This would be a massive feature for Scope - both in complexity and impact. Could be worth revisiting when there is more granular model support ( many base models, vaes, loras, controlnets, etc).

frontend/src/components/LoRAManager.tsx
ryanontheinside: I chose to use a button for a couple of reasons:
1) it is easy to overshoot with the slider, especially when the interface is small. This can result in accidentally setting the lora higher than intended which I found to be annoying. The decision to use -10,10 as the range was driven by personal experience, and the fact that this is what ComfyUI uses. That being said, even if we reduced it to -5,5 the problem would still remain. 
2) i found it to be nice to experiment with updating multiple loras at the same time

I do not have a strong preference here - let me know, happy to change it to use the thumb

yondonfu:
> How long did you let it load? On my windows 5090, the permanent merge step of the lora alone takes 60 seconds.

I was unable to load at all on a 5090 with 62GB RAM on Runpod (machine becomes unresponsive). I was able to load in 60+ seconds (I didn't record the actual time, but it was 60+ seconds I think) on a 5090 with 141GB RAM. I added specific notes in [this checklist](https://github.com/daydreamlive/scope/pull/141#issuecomment-3554155327).

I suspect the higher RAM requirements are related to how we do the LoRA loading/fusing on CPU (eg [here](https://github.com/daydreamlive/scope/pull/141/files#diff-51d87731af889ada46a2bcc1125e63be1e912e76e2d2ff6e31e020f1f033f931)) which I think happens because we need to have the LoRAs fused before the quantization step which is what moves weights from RAM to VRAM. And we do quantization on CPU (and move each tensor to GPU as it is ready with [this](https://github.com/daydreamlive/scope/blob/85602f486677fb867b60cd66363236f4101b589d/pipelines/krea_realtime_video/pipeline.py#L77)) because we can't fit everything into GPU up front.

It seems the minimum RAM when using a LoRA with a 14B model is somewhere between 62 and 64GB right now.

I'm wondering if PEFT has any utilities for this scenario? For example, it looks like they have a [merge_and_unload()](https://huggingface.co/docs/peft/en/developer_guides/lora#merge-lora-weights-into-the-base-model), but I haven't looked into how it works under the hood. From the sounds of the name maybe it iteratively updates the layers of the base model and then tosses out the adapter weights so that you don't need to have merged weights and base weights in memory at the same time? That could help I think...If there is a simple solve for the increased RAM requirement via a PEFT utility then I think it is worth doing. If not then I think this can be addressed outside of this PR.

If PEFT doesn't help, I still think there is a more memory efficient approach possible, but it'd be more complex eg as each layer is LoRA fused run dynamic quantization on that that and move the layer to the GPU before repeating with the next layer. And that would avoid the overhead of having both the merged weights and the base weights in RAM at the same time.

pipelines/wan2_1/lora/mixin.py
ryanontheinside: I tried this and found the gradual effect to be more visually pleasing. That being said, we could make it configurable. 

option 1: new global lora cache flag - determines cache behavior for lora updates
option 2: roll it in with existing manage cache

I would be partial to option 1 here.

WDYT?

frontend/src/components/SettingsPanel.tsx
ryanontheinside: This also works logically, putting the lora and the pipeline (base model) near each other

frontend/src/components/LoRAManager.tsx
ryanontheinside: A third reason I just remembered: clicking the + and - buttons sends a bunch of updates

yondonfu: I have a personal pref to send the update when the thumb stops mainly because I felt like I wanted to see the effect of a value change ASAP and then adjust as is needed. And for 2 I feel like my personal process right now would be to change the scale of each LoRA individually and tweak them iteratively.

Could changing the increment/decrement value in the slider to something smaller eg 0.01 help with 1?

I think we'll know more as we use it more and maybe I'll change my mind (quite possible...) but if you don't have a strong pref I'm in favor of sending update on thumb.

pipelines/krea_realtime_video/pipeline.py
ryanontheinside: Since lora is a one time configuration step as a whole, I logically grouped it with model loading, quantization, etc..

Extracting scale update to a block makes a lot of sense though.

pipelines/wan2_1/lora/mixin.py
yondonfu: Fair enough re: finding gradual effect more visually pleasing. I liked it at times, but at other times especially with V2V using streamdiffusionv2 I preferred seeing a clean shift. Feels like a sign that both should be supported.

Perhaps there could be a manage cache flag per parameter that affects visual quality eg some sort of toggle like "Reset Cache on Update?". But, I feel like it would add a bunch of clutter to the UI right now so I'd be interested in figuring out a better way to present all of that information first.

Given that WDYT about rolling into existing Manage Cache for now and you can still disable that to get the gradual visual effect you mention and then separately figuring out the above?

frontend/src/components/LoRAManager.tsx
yondonfu: > A third reason I just remembered: clicking the + and - buttons sends a bunch of updates

IIUC that's how the Denoising Step List and Noise Scale updates work too. Do you see that behavior as a negative?

pipelines/wan2_1/lora/mixin.py
yondonfu: Answer from offline - yes Wan2.1 specific.

pipelines/wan2_1/lora/mixin.py
yondonfu: Decision from offline - rolling into existing Manage Cache for now and the default is that changes to LoRA scale will trigger cache reset unless user has disabled Manage Cache and reevaluate per parameter auto cache reset separately.

frontend/src/components/LoRAManager.tsx
yondonfu: Decision from offline - remove button, see how we/users feel and then reevaluate slider behavior generally.

yondonfu (approved):
LGTM

I can now load Krea with a LoRA on a 62GB RAM Runpod machine. I do run into sporadic timeouts due to the way the Runpod proxy works, but will look into that separately.

PR #150, opened 2025-11-17
Fix CleanKVCacheBlock

leszko:
Signed-off-by: Rafal Leszko <rafal@livepeer.org>

### DIFF
--- a/pipelines/wan2_1/blocks/clean_kv_cache.py
+++ b/pipelines/wan2_1/blocks/clean_kv_cache.py
@@ -91,6 +91,9 @@ class CleanKVCacheBlock(ModularPipelineBlocks):
 
         generator_param = next(components.generator.parameters())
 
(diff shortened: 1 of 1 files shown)


pipelines/wan2_1/blocks/clean_kv_cache.py
```diff
@@ -114,6 +114,15 @@ def __call__(self, components, state: PipelineState) -> tuple[Any, PipelineState
 
         # Run the generator with the clean latent at timestep = 0 to update the KV cache
         conditional_dict = {"prompt_embeds": block_state.conditioning_embeds}
+        context_timestep = (
```
yondonfu, line 103: L100-110 can be removed given this, but let's preserve the comments above those lines here.

yondonfu (changes requested):
Just one comment.

PR #152, opened 2025-11-17
Remove prepare() function

leszko:


yondonfu:
https://github.com/daydreamlive/scope/pull/205 was the first part of the solution and https://github.com/daydreamlive/scope/issues/218 will be the follow up so closing this PR as it has an outdated solution.

PR #156, opened 2025-11-18
Restructure packages

leszko:
### Directory structure
```
/
├── frontend/
│       └── ...
├── scope/
│   ├── core/
│   │   └── pipelines/
│   │       ├── krea_realtime_video/
│   │       ├── longlive/
│   │       ├── streamdiffusionv2/
│   │       └── ...
│   └── server/
├── tests/
├── docs/
├── Dockerfile
├── pyproject.toml
└── README.md
```

### Usage

Here's a sample project presenting how you can now use the `core` package: https://github.com/leszko/scope-example

In short:
```python
from scope.core.pipelines import LongLivePipeline
```

### Additional changes
- Remove `CLAUDE.md` => We don't maintain it, it does not make sense to keep it
- Add `build.sh` => Build Python whl package

### What does not work?
1. `uv run app.py` => You need to use one of the following ways:
    - `uv run python -m scope.server.app` => For dev
    - `uv run build && uv run daydream-scope` => For build
2. Using uvx and (Test)PyPI package dependency
    - I published a new package into TestPyPI: https://test.pypi.org/project/daydream-scope/
    - In theory both these two should work:
      - Running the whole app by just typing `uvx run daydream-scope`
      - Importing the core (or server or all) dependencies from (Test)PyPI
    - It does not work, because of the (flash/sage)attention dependencies which are custom links (I think that pip/uvx does not support it for the security reasons, so that you don't download random indirect dependencies)

### Other

Original doc: https://www.notion.so/livepeer/Scope-Package-Structure-2ab0a34856878097b465dfeb1d6eafd2

### DIFF
--- a/scope/core/pipelines/common.py
+++ b/scope/core/pipelines/common.py
@@ -1,9 +1,30 @@
 """Core common definitions for Scope pipelines."""
 
 from enum import Enum
+from pathlib import Path
+
+from omegaconf import OmegaConf
 
 
 class Quantization(str, Enum):
     """Quantization method enumeration."""
 
     FP8_E4M3FN = "fp8_e4m3fn"
+
+
+def load_model_config(config, pipeline_file_path: str | Path) -> OmegaConf:
+    """
+    Load model configuration from config or auto-load from model.yaml.
+
+    Args:
+        config: Configuration object that may contain a model_config attribute
+        pipeline_file_path: Path to the pipeline's __file__ (used to locate model.yaml)
+
+    Returns:
+        OmegaConf: The model configuration, either from config or loaded from model.yaml
+    """
+    model_config = getattr(config, "model_config", None)
+    if not model_config:
+        model_yaml_path = Path(pipeline_file_path).parent / "model.yaml"
+        model_config = OmegaConf.load(model_yaml_path)
+    return model_config
--- a/scope/core/pipelines/krea_realtime_video/pipeline.py
+++ b/scope/core/pipelines/krea_realtime_video/pipeline.py
@@ -4,7 +4,7 @@ import time
 import torch
 from diffusers.modular_pipelines import PipelineState
 
-from scope.core.pipelines.common import Quantization
+from scope.core.pipelines.common import Quantization, load_model_config
 
 from ..blending import EmbeddingBlender
 from ..components import ComponentsManager
@@ -38,7 +38,8 @@ class KreaRealtimeVideoPipeline(Pipeline):
         tokenizer_path = getattr(config, "tokenizer_path", None)
         vae_path = getattr(config, "vae_path", None)
 
-        model_config = getattr(config, "model_config", {})
+        # Auto-load model.yaml if model_config not provided
+        model_config = load_model_config(config, __file__)
         base_model_name = getattr(model_config, "base_model_name", "Wan2.1-T2V-14B")
         base_model_kwargs = getattr(model_config, "base_model_kwargs", {})
 
--- a/scope/core/pipelines/longlive/pipeline.py
+++ b/scope/core/pipelines/longlive/pipeline.py
@@ -5,6 +5,7 @@ import torch
 from diffusers.modular_pipelines import PipelineState
 
 from ..blending import EmbeddingBlender
+from ..common import load_model_config
 from ..components import ComponentsManager
 from ..interface import Pipeline
 from ..process import postprocess_chunk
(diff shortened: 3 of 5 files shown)


yondonfu (changes requested):
Can you separate out commits such that there is a single commit that moves files around the others can include additional changes? This would make it much easier to review since the reviewer can just skim the one commit moving files around and then focus more on changes in the other commits.

yondonfu:
The example project was a helpful reference! This can be addressed outside of this PR, but I don't like that the user has to specify a model.yaml file themselves (eg [here](https://github.com/leszko/scope-example/blob/936fd3e11927b90683cfef1b8019da6e53b17154/test_pipeline.py#L25)). I'm thinking that the default behavior of the pipeline can be to populate the `model_config` field of the `config` object on its own by default in the pipeline class `__init__()` function using the model.yaml file included in the pipeline directory eg `longlive`, `streamdiffusionv2`, etc. The user can still set their own `model_config` field to override the defaults, but they don't have to.

yondonfu:
> Remove CLAUDE.md => We don't maintain it, it does not make sense to keep it

I use this with Claude Code so I'd prefer to keep it. It does require an update, but that's just a matter of asking Claude Code to do so.

leszko:
> The example project was a helpful reference! This can be addressed outside of this PR, but I don't like that the user has to specify a model.yaml file themselves (eg [here](https://github.com/leszko/scope-example/blob/936fd3e11927b90683cfef1b8019da6e53b17154/test_pipeline.py#L25)). I'm thinking that the default behavior of the pipeline can be to populate the `model_config` field of the `config` object on its own by default in the pipeline class `__init__()` function using the model.yaml file included in the pipeline directory eg `longlive`, `streamdiffusionv2`, etc. The user can still set their own `model_config` field to override the defaults, but they don't have to.

Implemented in this PR here: https://github.com/daydreamlive/scope/pull/156/commits/2ff19db4d3698ec5b5e7e9c2a56cfeb07b207ee1

leszko:
> Can you separate out commits such that there is a single commit that moves files around the others can include additional changes? This would make it much easier to review since the reviewer can just skim the one commit moving files around and then focus more on changes in the other commits.

Sure, separated:
1. https://github.com/daydreamlive/scope/pull/156/commits/86928dab0ebe341e9478e296722fa7c722d523b2 => Moving files + updating imports
2. https://github.com/daydreamlive/scope/pull/156/commits/71fdeb353a7ca9443197f5ca513742be12c52d2a => Other updates
3. https://github.com/daydreamlive/scope/pull/156/commits/4e5e43a24656cad3fc7fbc504840a3eddbe7d3f4 => Add script to build dist package (We don't necessarily need it, the same as publish.sh, but I'd keep it, because we want to get this PyPI package sorted out)
4. https://github.com/daydreamlive/scope/pull/156/commits/2ff19db4d3698ec5b5e7e9c2a56cfeb07b207ee1 => New change, adding default model.yaml for pipelines

pyproject.toml
```diff
@@ -53,12 +49,19 @@ dependencies = [
     "kernels>=0.10.4",
     "triton==3.4.0; sys_platform != 'win32'",
     "triton-windows==3.4.0.post21; sys_platform == 'win32'",
+    "pydantic>=2.0.0",  # Required by scope.core.pipelines.interface
```
yondonfu, line 52: I see that uv supports [optional dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#optional-dependencies):

> For example, Pandas has an [excel extra](https://pandas.pydata.org/docs/getting_started/install.html#excel-files) and a [plot extra](https://pandas.pydata.org/docs/getting_started/install.html#visualization) to avoid installation of Excel parsers and matplotlib unless someone explicitly requires them. Extras are requested with the package[<extra>] syntax, e.g., pandas[plot, excel].

Is that applicable here? Or would making any of these dependencies explicitly optional make running the app more cumbersome eg can't just do `uv run daydream-scope` as-is?

leszko: I think we could make it optional, but let's maybe wait for https://github.com/daydreamlive/scope/pull/152. If we decide to remove the `prepare()` function, then we don't need this dependency at all.

yondonfu: FWIW what I meant was make the server dependencies optional. But, it looks like as-is that won't work if we want to make sure `uv run daydream-scope` works unless bigger changes are made.

From Claude:

```
Based on the code, no, uv run daydream-scope will NOT work if you make scope.server dependencies optional without installing them.

  Looking at the entrypoint in pyproject.toml:59:
  daydream-scope = "scope.server.app:main"

  The scope.server.app module has hard imports for server dependencies:
  - uvicorn (line 17)
  - fastapi and FastAPI components (lines 18-22)
  - Plus all the WebRTC and other server components

  If you make these dependencies optional (e.g., move them to an [project.optional-dependencies] or [dependency-groups] section), the command will fail with import errors unless you:

  Option 1: Install with the optional group
  uv sync --extra server  # or whatever you name the group
  uv run daydream-scope

  Option 2: Use inline extras
  uv run --extra server daydream-scope

  Option 3: Keep server deps required for the daydream-scope script
  If you want uv run daydream-scope to work without flags, the FastAPI/uvicorn/aiortc dependencies must remain in the main dependencies list.

  Why you might make them optional: If you're planning to split the package into client and server components, or allow users to install only the ML pipeline parts without the server.

  Would you like me to help you restructure the dependencies to support this pattern?
```

PR #158, opened 2025-11-19
Add Server API

leszko:


src/scope/server/build.py
```diff
@@ -35,6 +35,16 @@ def run_command(cmd: list[str], cwd: Path | None = None) -> None:
 
 def main() -> None:
     """Main build function."""
+    import argparse
+
+    parser = argparse.ArgumentParser(description="Build daydream-scope")
```
yondonfu: I don't think this is necessary. If we don't need to build the frontend, then the user should be able to just run:

```
uv run daydream-scope
```

docs/server.md
```diff
@@ -0,0 +1,347 @@
...
+
+## Prerequisites
+
+```bash
+uv sync
+uv run build --server-only
+uv run download_models --pipeline <PIPELINE_ID>
+```
+
+Pipeline IDs: `streamdiffusionv2` (video input), `longlive`, `krea-realtime-video` (no video input), `passthrough` (testing)
```
yondonfu:
```suggestion
Pipeline IDs: `streamdiffusionv2` (video input), `longlive` (no video input), `krea-realtime-video` (no video input), `passthrough` (testing)
```

docs/server.md
```diff
@@ -0,0 +1,347 @@
...
+```javascript
+async function getPipelineStatus() {
+  const response = await fetch("http://localhost:8000/api/v1/pipeline/status");
+  return await response.json();
+}
+```
+
+## Connecting to the Server
+
+**Video-input pipelines** (`streamdiffusionv2`): Send video to server (bidirectional)
```
yondonfu: The formatting needs to be fixed so these are on separate lines.

docs/server.md
```diff
@@ -0,0 +1,347 @@
...
+pc.onconnectionstatechange = onConnectionStateChange;
+pc.oniceconnectionstatechange = onIceConnectionStateChange;
+pc.onicecandidate = onIceCandidate;
+
+// 6. Create offer
+const offer = await pc.createOffer();
+await pc.setLocalDescription(offer);
+```
+
+**Order:** Create data channel → Add transceiver/track → Set event handlers → Create offer
```
yondonfu: Formatting issue needs to be on separate lines.

src/scope/server/app.py
```diff
@@ -260,12 +260,16 @@ async def health_check():
 
 @app.get("/")
 async def root():
-    """Serve the frontend at the root URL."""
+    """Serve the frontend at the root URL or API info in server-only mode."""
```
yondonfu: Do we need a server only mode? 

AFAICT what we want is:

- Easily discoverable docs
- Easy for a LLM to then use those docs to get all the necessary information to know how to use the API

Are we assuming that we're always giving `server.md` in this case? If so, curious what the purpose of these specific changes are?

src/scope/server/build.py
leszko: Well, if we remove it from the build script, then the script will always try to build frontend. So, we always require having `npm` even if user wants just the server. That sounds wrong. I think that you should be able to build the app in server api without requiring the frontend-related stuff.

I'd rather remove `--server-only` from `app.py`, but we can leave it as well to give the ability to run daydream in server api mode even if you have the frontend built.

src/scope/server/app.py
leszko: We don't necessarily need it, but then we always serve frontend as well. That may be ok. I just think it may create a confusion, because you're developing your own frontend, but the backend is serving frontend as well.

docs/server.md
leszko: Cursor was always sooo good with markdown formatting...what happened?!? 🙃

src/scope/server/build.py
yondonfu: Clarified offline - we don't need server-only in the build script because uv run daydream-scope is sufficient to run the server on its own.

src/scope/server/app.py
yondonfu: Clarified offline - if you don't want the frontend you just don't run `uv run build` and the server will have no frontend files to serve.

docs/server.md
```diff
@@ -0,0 +1,347 @@
+# Scope Server API
+
+## Prerequisites
+
+```bash
+uv sync
+uv run build --server-only
```
yondonfu: Clarified offline - we wouldn't need this when just using the server.

docs/server.md
```diff
@@ -0,0 +1,347 @@
...
+uv run download_models --pipeline <PIPELINE_ID>
+```
+
+Pipeline IDs: `streamdiffusionv2` (video input), `longlive` (no video input), `krea-realtime-video` (no video input), `passthrough` (testing)
+
+## Starting the Server
+
+```bash
+uv run daydream-scope --server-only
```
yondonfu: Clarified offline - don't think we need the --server-only flag.

docs/server.md
```diff
@@ -0,0 +1,347 @@
+# Scope Server API
+
+## Prerequisites
+
+```bash
+uv sync
```
yondonfu: Clarified offline - actually don't need this bc uv run daydream-scope will auto handle

PR #160, opened 2025-11-19
Ryanontheinside/experiments/inputs/universal input modes

ryanontheinside:
DRAFT

### DIFF
--- a/frontend/src/constants/modes.ts
+++ b/frontend/src/constants/modes.ts
@@ -1,0 +1,15 @@
+export const GENERATION_MODE = {
+  VIDEO: "video",
+  TEXT: "text",
+} as const;
+
+export type GenerationMode =
+  (typeof GENERATION_MODE)[keyof typeof GENERATION_MODE];
+
+export const VIDEO_SOURCE_MODE = {
+  VIDEO: "video",
+  CAMERA: "camera",
+} as const;
+
+export type VideoSourceMode =
+  (typeof VIDEO_SOURCE_MODE)[keyof typeof VIDEO_SOURCE_MODE];
--- a/frontend/src/pages/StreamPage.tsx
+++ b/frontend/src/pages/StreamPage.tsx
@@ -19,6 +19,8 @@ import {
   getModeDefaults,
 } from "../lib/utils";
 import { getPipelineModeCapabilities } from "../lib/pipelineModes";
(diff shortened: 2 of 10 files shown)


pipelines/wan2_1/blocks/prepare_video_latents.py
```diff
@@ -77,6 +77,22 @@ def intermediate_outputs(self) -> list[OutputParam]:
...
 
+        # Align video latents with the generator's dtype and device so that downstream
+        # convolutions (e.g. patch_embedding) always receive tensors compatible with
+        # model parameters. This mirrors the behaviour of PrepareLatentsBlock.
+        generator_param = next(components.generator.model.parameters())
+
+        # If no video is provided, skip video latent preparation.
+        # This allows pipelines to share the same block graph for both
+        # text-to-video and video-to-video workflows.
+        if not hasattr(block_state, "video") or block_state.video is None:
```
yondonfu, line 88: Can we use [AutoPipelineBlocks](https://huggingface.co/docs/diffusers/main/en/modular_diffusers/auto_pipeline_blocks)?

IIUC this can be used to conditionally use different blocks depending on what input is provided. So, if `video` is provided then use `PrepareVideoLatentsBlock` but if it is omitted `PrepareLatentBlocks` is used.

pipelines/streamdiffusionv2/modular_blocks.py
```diff
@@ -18,15 +19,20 @@
...
+# Video mode: consumes input video and uses motion-aware noise control.
+VIDEO_BLOCKS = InsertableDict(
     [
         ("text_conditioning", TextConditioningBlock),
         ("embedding_blending", EmbeddingBlendingBlock),
         ("set_timesteps", SetTimestepsBlock),
         ("preprocess_video", PreprocessVideoBlock),
         ("noise_scale_controller", NoiseScaleControllerBlock),
         ("setup_caches", SetupCachesBlock),
+        # PrepareLatentsBlock always runs to create base latents for text-to-video
```
yondonfu, line 32: See other comment about using AutoPipelineBlocks. Maybe that would result in less duplication since there would just be a conditional switch on presence of an input `video` to use PrepareLatentsBlock vs. PrepareVideoLatentsBlock.

Issue #163, opened 2025-11-20
Fix timeout issue with Runpod proxy when pipelines take a long time to load (eg when fusing LoRAs)


yondonfu:
The pipeline load time increases when using LoRAs (particularly when using the Permanent Merge strategy).

This can be upwards of 60+ seconds and even 100+ seconds on a RTX 5090 when using a LoRA with a 14B model like in krea-realtime-video. This introduces an issue when accessing Scope via the Runpod proxy because it uses Cloudflare which has a default timeout of 100s:

> Cloudflare has a default timeout of approximately 100 seconds if the connection is not kept alive through some external means. So if you send an API request that simply waits for a response with no further communication, if that request takes longer than 100 seconds to fulfill it will time out. This can especially be an issue with very large hosted LLMs working on large context bodies of work, or intensive video or image generation. If the request starts streaming within 100 seconds that is fine, but if it doesn't even start the initial stream by then, nothing will be received at all as the connection will be closed.

I ran into this a couple of times today. 

A mitigation to this issue is if LoRA merging/fusing is faster and there are ways to address that. But, I don't think we'll get to that immediately. And either way the UI/backend should be more resilient to issues like this.

I suspect the reason that we're hitting this timeout issue is because of a synchronous operation while loading a pipeline - there might be a blocking request somewhere that times out? A resolution of this issue would involve making sure that even if loading a LoRA with krea-realtime-video takes 100+ seconds eg on a RTX 5090 on Runpod that we don't hit this timeout issue.

Note: This doesn't affect non-Runpod users (eg fully local), but since Runpod is one of easiest ways to use Scope for those without the requisite GPUs and since the Runpod proxy makes usage of the UI convenient I think we should address this issue.

PR #164, opened 2025-11-20
WIP: Load pipeline async

leszko:
Fix https://github.com/daydreamlive/scope/issues/163


yondonfu (changes requested):
The issue with timing out with Cloudflare when using the Runpod proxy URL (not when you port forward to localhost) still persists so there is still a problem somewhere that is causing a blocking operation that triggers the timeout.

PR #166, opened 2025-11-20
Ryanontheinside/fix/perm merge

ryanontheinside:
Fix: Enable community LoRA support with multi-adapter compatibility
- Supports community LoRA formats (lora_up/down, lora_unet_* prefixes)
- Preserves pre-existing PEFT adapters during merge (e.g., LongLive performance LoRA)
- Auto-converts incompatible formats to PEFT-compatible structure
- Uses PEFT's merge_and_unload for all adapters simultaneously

TESTED: 
-Single + Multi lora, with and without alternate adapter formats, for longlive and streamdiffusion
-Single lora support for Krea (examples of alternate adapter format unknown, not explicitly tested)

### DIFF
--- a/pipelines/wan2_1/lora/strategies/permanent_merge_lora.py
+++ b/pipelines/wan2_1/lora/strategies/permanent_merge_lora.py
@@ -6,11 +6,8 @@ providing zero inference overhead. LoRA scales are fixed at load time and cannot
 updated at runtime. Ideal for production deployment where maximum FPS is critical.
 """
 
-import logging
 import os
 import tempfile
-import time
-from pathlib import Path
 from typing import Any
 
 import torch
@@ -18,8 +15,6 @@ from safetensors.torch import load_file, save_file
 
 from pipelines.wan2_1.lora.strategies.peft_lora import PeftLoRAStrategy
 from pipelines.wan2_1.lora.utils import find_lora_pair, normalize_lora_key
-
-logger = logging.getLogger(__name__)
 
 __all__ = ["PermanentMergeLoRAStrategy"]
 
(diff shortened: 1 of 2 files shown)


yondonfu:
Tested the following:

- [x] 1 LoRA perm merge with streamdiffusionv2
- [x] 1 LoRA perm merge with longlive
- [x] 2 LoRA perm merge with longlive
- [x] 2 LoRA runtime PEFT with diff scales with longlive
- [x] 1 LoRA perm merge with krea
- [x] 1 LoRA runtime PEFT with diff scales with krea

pipelines/wan2_1/lora/strategies/permanent_merge_lora.py
```diff
@@ -6,20 +6,124 @@
...
 
 from pipelines.wan2_1.lora.strategies.peft_lora import PeftLoRAStrategy
-
-logger = logging.getLogger(__name__)
+from pipelines.wan2_1.lora.utils import find_lora_pair, normalize_lora_key
 
 __all__ = ["PermanentMergeLoRAStrategy"]
 
 
+def convert_community_lora_to_peft_format(
```
yondonfu, line 22: Is there a specific name for this format? And are there many alternative formats in the community or just this one?

pipelines/wan2_1/lora/strategies/permanent_merge_lora.py
```diff
@@ -6,20 +6,124 @@
...
+        normalized_key = normalize_lora_key(base_key)
+
+        if not normalized_key.startswith("diffusion_model."):
+            normalized_key = f"diffusion_model.{normalized_key}"
+
+        # Store pre-scaled weights (no alpha tensor needed)
+        converted_state[f"{normalized_key}.lora_A.weight"] = lora_A
+        converted_state[f"{normalized_key}.lora_B.weight"] = lora_B_scaled
+
+    temp_fd, temp_path = tempfile.mkstemp(
```
yondonfu, line 111: Not a need to fix for this PR, but noting that it looks like we have some repeated overhead here - every time we load a non-PEFT compliant format LoRA we load the state dict, re-format keys, write the converted state dict to a temp file. In the future, maybe worth considering writing the converted state dict to disk.

yondonfu (approved):

pipelines/wan2_1/lora/strategies/permanent_merge_lora.py
ryanontheinside: Not that I know of or could find specifically. I considered naming the function after the keys, considering we go from `  lora_unet_blocks_0_cross_attn_k.alpha` -> `diffusion_model.blocks.0.cross_attn.k.lora_A.weight`, but I am not sure what other formats we may come across. A better name than 'community lora' may have been 'non peft'.

pipelines/wan2_1/lora/strategies/permanent_merge_lora.py
ryanontheinside: We can actually eliminate IO entirely using functionality I previously [implemented ](https://github.com/daydreamlive/scope/blob/c343342956c5a333e1cd7797e9f9248fb3730d15/src/scope/core/pipelines/wan2_1/lora/strategies/peft_lora.py#L58) but overlooked.

https://github.com/daydreamlive/scope/pull/171

WDYT?

PR #171, opened 2025-11-21
Improve LoRA Conversion + Fixes

ryanontheinside:
This PR does the following:
- renames conversion function for clarity
- fixes bug in Runtime PEFT conversion case
- removes dead calculate method
- Improves LoRA conversion process - this is more efficient and removes temp file workaround:

| Phase             | Temp file     | In Mem     | Improvement              |
|-------------------|---------|---------|---------------------------|
| Convert           | 0.141s  | 0.052s  | +89ms (-63%)           |
| Load              | 1.030s  | 0.906s  | +124ms (-12%)             |
| Merge             | 5.665s  | 5.517s  | +148ms (-3%)              |
| Cleanup           | 0.000s  | 0.000s  | Same                      |
| Temp File Cleanup | 0.007s  | 0.000s  | +7ms                      |

### DIFF
--- a/src/scope/core/pipelines/wan2_1/lora/strategies/peft_lora.py
+++ b/src/scope/core/pipelines/wan2_1/lora/strategies/peft_lora.py
@@ -396,7 +396,7 @@ class PeftLoRAStrategy:
         logger.debug(f"load_adapter: Sample LoRA keys from file: {sample_lora_keys}")
 
         # Standardize LoRA format if needed (handles lora_up/lora_down, missing diffusion_model prefix, etc.)
-        converted_state = standardize_lora_for_peft(lora_path=lora_path, model=model)
+        converted_state = standardize_lora_for_peft(lora_path=lora_path)
         if converted_state is not None:
             logger.debug(
                 f"load_adapter: Converted LoRA to PEFT format ({len(converted_state)} keys)"
--- a/src/scope/core/pipelines/wan2_1/lora/utils.py
+++ b/src/scope/core/pipelines/wan2_1/lora/utils.py
@@ -152,7 +152,7 @@ def find_lora_pair(
 
 
 def standardize_lora_for_peft(
-    lora_path: str, model: torch.nn.Module
+    lora_path: str,
 ) -> dict[str, torch.Tensor] | None:
     """
     Standardize LoRA formats to PEFT-compatible format.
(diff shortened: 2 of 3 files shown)


src/scope/core/pipelines/wan2_1/lora/strategies/peft_lora.py
```diff
@@ -394,6 +395,19 @@ def load_adapter(
...
 
+        # Standardize LoRA format if needed (handles lora_up/lora_down, missing diffusion_model prefix, etc.)
+        converted_state = standardize_lora_for_peft(lora_path=lora_path, model=model)
+        if converted_state is not None:
+            logger.debug(
+                f"load_adapter: Converted LoRA to PEFT format ({len(converted_state)} keys)"
+            )
+            lora_state = converted_state
+            # Sample converted keys for debugging
+            sample_converted_keys = list(lora_state.keys())[:5]
```
leszko, line 406: Stupid question: why do we take only first 5 elements `[:5]` from the converted state?

src/scope/core/pipelines/wan2_1/lora/utils.py
```diff
@@ -148,20 +151,96 @@ def find_lora_pair(
     return (base_key, alpha_key, lora_A, lora_B)
 
 
-def calculate_lora_scale(alpha: float | None, rank: int) -> float:
+def standardize_lora_for_peft(
+    lora_path: str, model: torch.nn.Module
```
leszko: Nit: I think `model` is not used, do we want to keep it as a param?

leszko (commented):
Looks good. Added some (probably stupid) questions/comments. Other than that looks fine (though I still don't understand (yet!) this in depth, so better if @yondonfu also takes a look).

src/scope/core/pipelines/wan2_1/lora/strategies/peft_lora.py
yondonfu: Seems to just be for easier debugging so you can eyeball if any of the keys changes by looking at the first 5.

src/scope/core/pipelines/wan2_1/lora/utils.py
yondonfu: +1

PR #172, opened 2025-11-21
Ryanontheinside/feat/independent lora merge strategies

ryanontheinside:
Depends upon #171 .

# Per-LoRA Merge Strategy Support
Enables independent merge strategy selection for each LoRA adapter instead of forcing a global strategy.

## What Changed
- Frontend: Added per-LoRA strategy dropdown in LoRA Manager
- Backend: LoRAConfig.merge_mode optional field with fallback to global lora_merge_mode
- Manager: Groups LoRAs by strategy, loads in order (permanent_merge → runtime_peft), routes runtime updates to appropriate handler

## Benefits
- Mix permanent_merge (max FPS) and runtime_peft (live adjustments) LoRAs simultaneously
- Fine-grained control per adapter without pipeline-wide tradeoffs
- Backward compatible: defaults to global strategy if not specified

## Technical Details
- LoRAManager refactored: consolidated mode resolution, generic grouping helper
- Frontend UI consistency: both Strategy and Scale labels use LabelWithTooltip
- Timeline export/import automatically preserves per-LoRA mergeMode
- Scale slider tooltip dynamically reflects effective merge mode

## Example Use Case
Some LoRAs are desirable for performance but do not require scale updates. Users that want to usually/always include these with maximum performance, but still want to runtime updates of other LoRA will need this feature.  

### Example performance LoRA for:
- [1.3b](https://civitai.com/models/1934938?modelVersionId=2192703)
- [14b](https://civitai.com/models/1678575/wan21fusionx-the-lora)
image

## Next Steps
A more explicit/elegant pattern for merge order can be revisited when moving scale updates to modular block is considered in place of Mixin.

### DIFF
--- a/frontend/src/data/parameterMetadata.ts
+++ b/frontend/src/data/parameterMetadata.ts
@@ -70,11 +70,11 @@ export const PARAMETER_METADATA: Record<string, ParameterMetadata> = {
   loraScale: {
     label: "Scale:",
     tooltip:
-      "Adjust LoRA strength. Updates automatically when you release the slider or use +/- buttons. 0.0 = no effect, 1.0 = full strength",
+      "Adjust LoRA strength. Updates automatically when you release the slider or use +/- buttons. Typical values: 0.0 = no effect, 1.0 = full strength. Full range -10.0 to 10.0 available depending on LoRA specifications.",
   },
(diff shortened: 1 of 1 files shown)


yondonfu (approved):
LGTM!

PR #177, opened 2025-11-21
Centralize Pipeline Defaults via Abstract Interface

ryanontheinside:
# Centralize Pipeline Defaults via Abstract Interface

First step toward universal input methods: establish a single source of truth for pipeline configuration defaults.

### Backend
- Add `PipelineDefaults` Pydantic model to `interface.py` defining typed default configuration
- Add abstract `get_defaults()` classmethod to `Pipeline` base class
- Implement `get_defaults()` in all pipeline classes (StreamDiffusionV2, LongLive, KreaRealtimeVideo, Passthrough)
- Add `defaults.py` module with utilities for extracting and applying defaults
- Add helper method `_initialize_state_with_defaults()` to reduce boilerplate in pipeline constructors
- Add `/api/v1/pipeline/defaults` endpoint to expose pipeline defaults to frontend
- Refactor `pipeline_manager.py` to use pipeline defaults instead of hardcoded values

### Frontend
- Add `getPipelineDefaults()` API client function
- Update `useStreamState` to fetch and apply pipeline defaults when pipeline changes
- Remove hardcoded defaults from `StreamPage.tsx` pipeline switching logic
- Remove unused `getDefaultDenoisingSteps()` utility (now backend-driven)

## Benefits

- **DRY**: Eliminated duplicate default values across frontend and backend
- **Type Safety**: Pydantic validation ensures consistent configuration
- **Maintainability**: Each pipeline owns its defaults in one place
- **Flexibility**: Frontend automatically adapts to backend default changes
- **Extensibility**: Useful in imminent work on universal input methods, could pave the way for user-defined defaults

## Next Steps

- Pipeline registry pattern to remove if/elif chains in API endpoint
- Additional universal input mode features building on this foundation

## Testing

- [x] Verify pipeline switching updates all settings correctly
- [x] Confirm each pipeline loads with appropriate defaults
- [x] Check API endpoint returns correct defaults for all pipeline IDs

### DIFF
--- a/src/scope/core/pipelines/interface.py
+++ b/src/scope/core/pipelines/interface.py
@@ -1,9 +1,11 @@
 """Base interface for all pipelines."""
 
 from abc import ABC, abstractmethod
+from typing import Any
 
 import torch
-from pydantic import BaseModel
+from diffusers.modular_pipelines import PipelineState
+from pydantic import BaseModel, Field, field_validator
 
 
 class Requirements(BaseModel):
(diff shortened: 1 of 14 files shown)


src/scope/server/app.py
```diff
@@ -308,6 +309,42 @@ async def get_pipeline_status(
         raise HTTPException(status_code=500, detail=str(e)) from e
 
 
+@app.get("/api/v1/pipeline/defaults", response_model=PipelineDefaultsResponse)
```
yondonfu, line 312: It feels like defaults are a part of pipeline metadata. This makes me think that defaults could be included alongside all pipeline metadata that is returned by an API endpoint. 

WDYT about a `GET /api/v1/pipeline/<pipeline_id>` endpoint that returns pipeline metadata which includes supported parameters and defaults? I imagine that later on (not saying this needs to happen in this PR) it'd be helpful for the UI to just query the backend for parameters so it can setup proper controls depending on the type of parameter and if we're going to be using an API endpoint to send back defaults it might make sense to just have that API endpoint be more general now (can still iterate on format over time) and send back a payload representing all pipeline metadata which includes defaults.

Something like what is described [here](https://chatgpt.com/share/6924b6f2-9800-800e-a276-f32914fff3a7)? It'd be a bonus if we can use standard conventions like OpenAPI schema that is supported by the Python libraries we're using for easy validation.

src/scope/server/app.py
```diff
@@ -373,14 +416,20 @@ def process_lora_file(file_path: Path, lora_dir: Path) -> LoRAFileInfo:
         )
 
     try:
-        lora_dir = get_models_dir() / "lora"
+        lora_dir = Path("models/lora")
```
yondonfu, line 419: Noting that I made a change in https://github.com/daydreamlive/scope/commit/cbe1cd6cd1c1122cbec45f8fa905369940191fcf because I think we want to have this be relative to the base models dir (which by default is `~/.daydream-scope` right now, but can be overridden with env var).

src/scope/server/app.py
```diff
@@ -194,8 +195,8 @@ async def lifespan(app: FastAPI):
     logs_dir = get_logs_dir()
     logger.info(f"Logs directory: {logs_dir}")
 
-    # Ensure models directory and lora subdirectory exist
-    models_dir = ensure_models_dir()
+    # Log models directory
+    models_dir = get_models_dir()
```
yondonfu, line 199: Noting I made a change in https://github.com/daydreamlive/scope/commit/cbe1cd6cd1c1122cbec45f8fa905369940191fcf to use `ensure_models_dir()` with makes sure that `models/lora` exists as opposed to forcing users to create the dir themselves if they want to put LoRA files there.

src/scope/core/pipelines/interface.py
```diff
@@ -12,9 +14,118 @@ class Requirements(BaseModel):
...
+
+    model_config = {"extra": "allow"}
+
+
 class Pipeline(ABC):
     """Abstract base class for all pipelines."""
 
+    @classmethod
+    @abstractmethod
+    def get_defaults(cls) -> PipelineDefaults:
```
yondonfu, line 66: If we consider the approach of having a general way to expose pipeline metadata then maybe this becomes something like:

```
@property
def schema(self) -> dict:
    ...
```

And a property is used to express that everything is self-contained and not dynamically computed based on an input.

This could also be a natural way to expose config information for a caller to determine the input chunk size for V2V cc @leszko as an alternative to what is currently being discussed in https://github.com/daydreamlive/scope/pull/152.

leszko: Yes, I like. But I'd also prefer to have it more "abstract" and return `dict` instead of `PipelineDefaults`. The biggest issue I see with the typed `PipelineDefaults` is that not every pipeline will have the same fields. So, what Yondon suggested will work better IMO.

src/scope/core/pipelines/interface.py
```diff
@@ -12,9 +14,118 @@ class Requirements(BaseModel):
...
+        # Noise controls from defaults
+        if defaults.noise_scale is not None:
+            state.set("noise_scale", defaults.noise_scale)
+
+        if defaults.noise_controller is not None:
+            state.set("noise_controller", defaults.noise_controller)
+
+        # Pipeline-specific parameters (optional, may not exist in all defaults)
+        if defaults.kv_cache_attention_bias is not None:
+            state.set("kv_cache_attention_bias", defaults.kv_cache_attention_bias)
```
leszko, line 127: If we decide to stick to `dict`, then this code will also simplify, because we'll only iterate over the dict and set the state for each key.

leszko (commented):
I like the change. The only thing is what Yondon mentioned in the inline comments, especially the part about returning `dict` in the interface rather than the specific fields.

ryanontheinside:

PR #178, opened 2025-11-21
Mode-Aware Pipeline Configuration Infrastructure

ryanontheinside:
## Add Mode-Aware Pipeline Configuration Infrastructure

Depends upon #177 .

### Overview
This PR introduces foundational infrastructure for generation mode support (text-to-video vs video-to-video) across all pipelines. This is **Step 2** of the Universal Input Methods project.

### What Changed

**Core Architecture:**
- Pipelines now declare a native generation mode and provide mode-specific default configurations
- Introduced `PipelineModeConfig` to encapsulate mode-specific settings (resolution, denoising steps, noise controls, etc.)
- Restructured `PipelineDefaults` to contain both text and video mode configurations

**Runtime Behavior:**
- Pipelines initialize with their native mode defaults
- Frame processor automatically resets cache when generation mode changes
- Frontend extracts and applies correct mode configuration based on pipeline's native mode

**Type Safety:**
- Added `generation_mode` parameter to backend schema with proper typing
- Frontend types updated to support mode-aware configurations

### Current State
All pipelines declare mode configurations, but currently only operate in their native mode. Full mode-switching behavior will be implemented in Step 3.

### Architecture Note
Pipelines are required to declare both text and video mode configurations (ISP consideration). This will be considered and addressed in the next PR.

### DIFF
--- a/src/scope/core/pipelines/helpers.py
+++ b/src/scope/core/pipelines/helpers.py
@@ -1,0 +1,134 @@
+"""Helper utilities for pipeline configuration and initialization.
+
+These helpers are used by Pipeline implementations to reduce boilerplate
+in configuration and state initialization.
+"""
+
+from typing import Any
+
+from diffusers.modular_pipelines import PipelineState
+
+from .schema import create_mode_config
+
+
+def create_mode_configs(
+    shared: dict[str, Any],
+    text_overrides: dict[str, Any] | None = None,
+    video_overrides: dict[str, Any] | None = None,
+) -> dict[str, dict[str, Any]]:
+    """Helper to create mode configs with shared values and mode-specific overrides.
+
+    Args:
+        shared: Common configuration values shared across both modes
+        text_overrides: Text-mode specific overrides (default: empty)
+        video_overrides: Video-mode specific overrides (default: empty)
+
+    Returns:
+        Dictionary with 'text' and 'video' mode configurations
+    """
+    text_overrides = text_overrides or {}
+    video_overrides = video_overrides or {}
+
+    from .defaults import GENERATION_MODE_TEXT, GENERATION_MODE_VIDEO
+
+    return {
+        GENERATION_MODE_TEXT: create_mode_config(**{**shared, **text_overrides}),
+        GENERATION_MODE_VIDEO: create_mode_config(**{**shared, **video_overrides}),
+    }
+
+
+def build_pipeline_schema(
+    pipeline_id: str,
+    name: str,
+    description: str,
+    native_mode: str,
+    shared: dict[str, Any],
+    text_overrides: dict[str, Any] | None = None,
+    video_overrides: dict[str, Any] | None = None,
+    version: str = "1.0.0",
+) -> dict[str, Any]:
+    """Build complete pipeline schema with native mode and mode configs.
+
+    This helper reduces boilerplate by combining mode config creation with
+    schema construction.
+
+    Args:
+        pipeline_id: Unique identifier for the pipeline
+        name: Human-readable pipeline name
+        description: Description of pipeline capabilities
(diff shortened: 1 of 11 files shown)


src/scope/core/pipelines/interface.py
```diff
@@ -58,73 +57,195 @@ def validate_resolution(cls, v: dict[str, int]) -> dict[str, int]:
...
+    native_generation_mode: str = Field(
+        ...,
+        description="Native generation mode for this pipeline (text or video)",
+    )
+    modes: dict[str, PipelineModeConfig] = Field(
+        ...,
+        description="Mode-specific defaults. Keys are 'text' and 'video'",
+    )
+
+    model_config = {"extra": "forbid"}
```
leszko: I think that when it comes to the interface, I'd think in a way that it should work now for all our pipelines, but also for the future pipelines. In that sense:
- Defaults should be returned as a `dict` instead of `PipelineDefault` with fields => IIUC they may be different for different pipelines, e.g. some pipelines may not support noise controller
- Modes can be returned as an object, but it should be pretty minimalistic, something that will fit all potential pipelines, like include only the fields related to the input type (T2V, V2V, I2V). Saying that, I think returning everything in one dict object is ok as well. Like to return key/values like `{"native_mode": "text-to-video", "modes": [ "text-to-video", "video-to-video"]}` (or something similar).

src/scope/core/pipelines/interface.py
```diff
@@ -58,73 +57,195 @@ def validate_resolution(cls, v: dict[str, int]) -> dict[str, int]:
...
+    Principle deviation, accepted because:
+    1. All pipelines will support both modes in future iterations
+    2. The temporary violation simplifies the architecture during the transition
+    3. The overhead is minimal (declaration only, no forced implementation)
+    """
+
+    NATIVE_GENERATION_MODE: str = "text"
+
+    @classmethod
+    def _create_mode_configs(
```
leszko: nit: I'd probably move helper functions to a separate file to keep the `interface.py` clean. I think the `interface.py` should be pretty minimalistic, because that is something someone will read if they want to develop their own pipelines.

src/scope/core/pipelines/helpers.py
```diff
@@ -0,0 +1,134 @@
...
+    """Build complete pipeline schema with native mode and mode configs.
+
+    This helper reduces boilerplate by combining mode config creation with
+    schema construction.
+
+    Args:
+        pipeline_id: Unique identifier for the pipeline
+        name: Human-readable pipeline name
+        description: Description of pipeline capabilities
+        native_mode: Native generation mode (text or video)
```
yondonfu, line 59: Why do we need concept of a native mode? Can a pipeline just support a list of modes and then a caller can choose a mode and construct the input based on that mode? Eg if "text-to-video" no extra param, if "video-to-video" add a `video` param, later on if `image-to-video` add a `image` param that is consumed as an one-shot event.

src/scope/core/pipelines/interface.py
```diff
@@ -14,118 +13,60 @@ class Requirements(BaseModel):
...
-class Pipeline(ABC):
-    """Abstract base class for all pipelines."""
+    See helpers.py and schema.py for utilities to simplify schema creation.
+    """
 
     @classmethod
     @abstractmethod
-    def get_defaults(cls) -> PipelineDefaults:
-        """Return typed default parameters for this pipeline.
+    def get_schema(cls) -> dict[str, Any]:
```
yondonfu, line 31: I think this probably should be a property as opposed to a method since everything is self-contained and not dependent on dynamic runtime inputs. Then, it'll read more like just accessing read-only data.

ryanontheinside:

PR #186, opened 2025-11-24
Fix: Warmup changes for krea to maximize startup performance.

BuffMcBigHuge:
# Fix Krea Realtime Video Startup Stuttering

### Problem

When starting a stream with `krea-realtime-video`, users experience ~5 seconds of severe stuttering (2 FPS) before the frame rate stabilizes. This is caused by `torch.compile` recompilation and autotuning happening at runtime instead of during initialization.

### Root Cause

Two issues prevented the warm-up from being effective:

1. **Wrong code path**: Warm-up ran with `kv_cache_attention_bias=1.0` (default), which uses the standard attention path. At runtime, the UI may send a different value (e.g., `0.99`), triggering the `flex_attention` path for the first time.

2. **Incomplete cache state**: Warm-up only ran 3 iterations but reset the cache each time, never reaching the steady-state "full cache" shape that `torch.compile` needs to pre-compile.

### Solution

1. **Trigger the correct code path**: Use `kv_cache_attention_bias=0.99` during warm-up to compile the `flex_attention` kernel.

2. **Fill the cache completely**: Calculate the number of warm-up runs needed to fill the KV cache based on config values, then run with `init_cache=False` after the first iteration to simulate cache accumulation.

3. **Prevent scalar specialization**: Convert scalar values used in `score_mod` to tensors so `torch.compile` treats them as dynamic inputs rather than compile-time constants.

### Changes

**`pipeline.py`**
- Added `WARMUP_KV_CACHE_ATTENTION_BIAS = 0.99` constant
- Calculate warm-up runs dynamically: `(local_attn_size // num_frame_per_block) + 1`
- Pass `init_cache=(i == 0)` to accumulate cache across warm-up iterations

**`causal_model.py`**
- Convert `cache_current_block_start` to tensor before use in `score_mod`

### Testing

1. Start server with `krea-realtime-video` pipeline
2. Observe warm-up logs (should show 3 runs for default config)
3. Start a stream - frame rate should be stable from the first frame
4. No `AUTOTUNE` logs should appear during stream start

### DIFF
--- a/src/scope/core/pipelines/krea_realtime_video/modules/causal_model.py
+++ b/src/scope/core/pipelines/krea_realtime_video/modules/causal_model.py
@@ -512,11 +512,7 @@ class CausalWanSelfAttention(nn.Module):
             if kv_cache_attention_bias != KV_CACHE_ATTENTION_BIAS_DISABLED:
                 # Use flex_attention with bias to mitigate error accumulation in past frames
                 # log_scale in (0, 1]: smaller values = less attention to past frame tokens
-                log_scale = torch.tensor(
-                    math.log(kv_cache_attention_bias),
-                    device=roped_query.device,
-                    dtype=torch.float32,
-                )
+                log_scale = math.log(kv_cache_attention_bias)
 
                 # Exclude first frame and current block from bias
                 cache_len = local_end_index - kv_start_idx


src/scope/core/pipelines/krea_realtime_video/pipeline.py
```diff
@@ -142,11 +144,27 @@ def __init__(
...
+        num_frame_per_block = getattr(model_config, "num_frame_per_block", 3)
+        warmup_runs = (local_attn_size // num_frame_per_block) + 1
+
         start = time.time()
-        for _ in range(WARMUP_RUNS):
-            self._generate(prompts=WARMUP_PROMPT)
+        for i in range(warmup_runs):
+            self._generate(
+                prompts=WARMUP_PROMPT,
+                kv_cache_attention_bias=WARMUP_KV_CACHE_ATTENTION_BIAS,
```
yondonfu: In practice, I don't think the actual value of `kv_cache_attention_bias` used here matters as long as it is < 1.0 which would trigger the code path during compilation that previously was missed.

If this is the case since we generally expect to use `kv_cache_attention_bias` < 1.0 for Krea and since the UI already defaults the value to 0.3 WDYT if we just set the default here to 0.3 as well and instead of injecting it here we can just update L141 to:

```
self.state.set("kv_cache_attention_bias", DEFAULT_KV_CACHE_ATTENTION_BIAS)
```

and up to with the rest of the constants we have:

```
DEFAULT_KV_CACHE_ATTENTION_BIAS = 0.3
```

src/scope/core/pipelines/krea_realtime_video/pipeline.py
```diff
@@ -142,11 +144,27 @@ def __init__(
...
         self.state.set("base_seed", getattr(config, "seed", 42))
 
+        # Warm-up: Run enough iterations to fill the KV cache completely.
+        # This ensures torch.compile compiles the flex_attention kernel at the
+        # steady-state cache size, avoiding recompilation during actual streaming.
+        #
+        # Cache fills at: num_frame_per_block frames per iteration
+        # Cache capacity: local_attn_size frames
+        # Iterations needed: ceil(local_attn_size / num_frame_per_block) + 1
+        #   (+1 to exercise the "cache full with eviction" path)
```
yondonfu, line 154: Do we actually care about this? I thought that the KV cache recomputation that Krea does means that we never do eviction from the KV cache and instead we just recompute the K/Vs in the cache every call using frames that are tracked external to the KV cache?

src/scope/core/pipelines/krea_realtime_video/modules/causal_model.py
```diff
@@ -512,7 +512,11 @@ def qkv_fn(x):
             if kv_cache_attention_bias != KV_CACHE_ATTENTION_BIAS_DISABLED:
                 # Use flex_attention with bias to mitigate error accumulation in past frames
                 # log_scale in (0, 1]: smaller values = less attention to past frame tokens
-                log_scale = math.log(kv_cache_attention_bias)
+                log_scale = torch.tensor(
```
yondonfu: Just to clarify - were these changes to fix any behavior or just for consistency to always use tensors?

src/scope/core/pipelines/krea_realtime_video/modules/causal_model.py
```diff
@@ -538,14 +542,14 @@ def qkv_fn(x):
...
-                # Convert scalars to tensors to avoid ShapeAsConstantBuffer dtype issues during compilation
+                # Convert scalars to tensors to avoid specialization during torch.compile
                 frame_seqlen_tensor = torch.tensor(
                     frame_seqlen, dtype=torch.int32, device=roped_query.device
                 )
-                # cache_current_block_start might already be a tensor([0]) based on the way local_end_index is initialized
-                # in the KV cache but we want a scalar tensor so we check for that case and convert if needed
-                cache_current_block_start_tensor = cache_current_block_start
-                if isinstance(cache_current_block_start_tensor, torch.Tensor) and cache_current_block_start_tensor.ndim > 0:
+                cache_current_block_start_tensor = torch.as_tensor(
```
yondonfu, line 546: Are there any problems if `cache_current_block_start` is already a tensor (as mentioned in previous comment here)?

src/scope/core/pipelines/krea_realtime_video/pipeline.py
BuffMcBigHuge: Given that:
1. The eviction path exists and could be exercised during streaming
2. The +1 adds minimal overhead (one extra warm-up iteration)
3. It ensures we compile both the "cache filling" and "cache full" states

I think keeping the +1 is safer, but we can remove it if you prefer. I'm not sure the best approach. It's your call (line 157):

```
warmup_runs = (local_attn_size + num_frame_per_block - 1) // num_frame_per_block  # ceil division
```

src/scope/core/pipelines/krea_realtime_video/modules/causal_model.py
BuffMcBigHuge: I believe this is a fix due to allowing them to be dynamic to avoid re-compilation. 

Converting to tensors makes them dynamic inputs, so:
1. `log_scale`: Already converted (line 515-519) - prevents recompilation if bias value changes
2. `cache_current_block_start_tensor`: Now converted (line 549-553) - prevents recompilation as cache fills

This is a fix, not just consistency. Without it, changing the bias value or cache state can cause recompilation during streaming.

Do you expect these values to change?

src/scope/core/pipelines/krea_realtime_video/modules/causal_model.py
BuffMcBigHuge: If it's already a scalar tensor, torch.as_tensor() is a no-op and squeeze() does nothing. If it's a 1-d tensor like [0], squeeze() converts it to a scalar. Sorta of defensive mechanism.

src/scope/core/pipelines/krea_realtime_video/pipeline.py
yondonfu: It's fine to keep it just making sure we understand why we're doing this!

src/scope/core/pipelines/krea_realtime_video/pipeline.py
```diff
@@ -19,7 +19,9 @@
 
 DEFAULT_DENOISING_STEP_LIST = [1000, 750, 500, 250]
 
-WARMUP_RUNS = 3
+# Warm-up configuration for torch.compile pre-compilation
+# Use a non-1.0 value to trigger the flex_attention code path during warmup
+WARMUP_KV_CACHE_ATTENTION_BIAS = 0.3
```
yondonfu:
```suggestion
DEFAULT_KV_CACHE_ATTENTION_BIAS = 0.3
```

src/scope/core/pipelines/krea_realtime_video/pipeline.py
```diff
@@ -136,17 +138,33 @@ def __init__(
...
+        num_frame_per_block = getattr(model_config, "num_frame_per_block", 3)
+        warmup_runs = (local_attn_size // num_frame_per_block) + 1
+
         start = time.time()
-        for _ in range(WARMUP_RUNS):
-            self._generate(prompts=WARMUP_PROMPT)
+        for i in range(warmup_runs):
+            self._generate(
+                prompts=WARMUP_PROMPT,
+                kv_cache_attention_bias=WARMUP_KV_CACHE_ATTENTION_BIAS,
```
yondonfu: This seems unnecessary if we're setting the state earlier.

src/scope/core/pipelines/krea_realtime_video/modules/causal_model.py
yondonfu: After looking at https://github.com/daydreamlive/scope/pull/192/files it seems that we're making changes to address the same issue here and in that PR. If so, let's just remove these changes from this PR so this PR can be merged cleanly and use the changes in the other PR?

src/scope/core/pipelines/krea_realtime_video/pipeline.py
yondonfu: Fixed in https://github.com/daydreamlive/scope/pull/186/commits/8fd7e54f3bcac4271dfc9e74e2cbf36e6c469768

src/scope/core/pipelines/krea_realtime_video/pipeline.py
yondonfu: Fixed in https://github.com/daydreamlive/scope/pull/186/commits/8fd7e54f3bcac4271dfc9e74e2cbf36e6c469768

src/scope/core/pipelines/krea_realtime_video/modules/causal_model.py
yondonfu: Rolled back these changes due to previous comment.

src/scope/core/pipelines/krea_realtime_video/modules/causal_model.py
yondonfu: Rolled back these changes because they already went in https://github.com/daydreamlive/scope/pull/192/files

PR #188, opened 2025-11-25
Ryanontheinside/feat/universal input modes/03 univeral input modes

ryanontheinside:


src/scope/core/pipelines/streamdiffusionv2/modular_blocks.py
```diff
@@ -35,7 +37,29 @@
     ]
 )
 
+# Text mode: pure text-to-video, no video-dependent blocks. We still run
+# EmbeddingBlendingBlock so that conditioning_embeds is produced for the
+# DenoiseBlock, matching the behaviour of the video path and other pipelines.
+TEXT_BLOCKS = InsertableDict(
```
yondonfu: Just leaving a quick comment before actually reviewing design:

Any particular reason why we can't use AutoPipelineBlocks from modular diffusers to have a single set of blocks and just conditionally switch on presence of a video input to determine whether to use/include V2V specific blocks eg PrepareVideoLatentsBlock, PreprocessVideo, etc.?

I'm asking about this because in my mind if possible it seems simpler to be able to to have a single set of blocks with conditional including/exclusion of certain blocks depending on the workflow being used which is determined by whether there is a video input (which I believe AutoPipelineBlocks is supposed to handle).

leszko: FWIW I _think_ I had auto blocks working at some point in my first modular PR. I can find it if needed.

frontend/src/components/SettingsPanel.tsx
```diff
@@ -68,34 +69,44 @@ export function SettingsPanel({
...
+  const modeCapabilities = getPipelineModeCapabilities(pipelineId);
+  const effectiveGenerationMode = generationMode ?? modeCapabilities.nativeMode;
+
   // Local slider state management hooks
-  const noiseScaleSlider = useLocalSliderValue(noiseScale, onNoiseScaleChange);
+  const noiseScaleSlider = useLocalSliderValue(
+    noiseScale,
+    onNoiseScaleChange,
+    2,
+    0.7
```
leszko: Shouldn't these default params be defined in the Pipeline Defaults?

frontend/src/lib/pipelineModes.ts
```diff
@@ -0,0 +1,108 @@
+import { PIPELINES } from "../data/pipelines";
+import { getCachedPipelineSchema } from "./utils";
+import { GENERATION_MODE, type GenerationMode } from "../constants/modes";
+import type { PipelineCategory, PipelineId } from "../types";
+
+export interface PipelineModeCapabilities {
```
leszko, line 6: I'd focus on the fields we have in this interface. I think a lot of properties here are either redundant or unused. E.g. 1. `requiresVideoInVideoMode`, isn't it the same as `supportsVideo`? I think we just have Video mode and Text mode
2. `category` => Isn't it the same as these `supportsVideo`, `supportsText`?

In general, maybe we could try to reduce the number of properties to what we need.

frontend/src/lib/utils.ts
```diff
@@ -1,26 +1,175 @@
...
+  const cached = pipelineSchemaCache.get(pipelineId);
+  if (!cached) {
+    return null;
+  }
+
+  // Check if expired
+  if (Date.now() - cached.timestamp > SCHEMA_CACHE_TTL) {
+    pipelineSchemaCache.delete(pipelineId);
+    return null;
+  }
```
leszko: I wonder if we even need any cache eviction. Honestly I think that if anything in the backend changes, then user will just refresh the frontend anyway.

frontend/src/lib/utils.ts
```diff
@@ -1,26 +1,175 @@
...
+  if (modeConfig.kv_cache_attention_bias) {
+    extracted.kv_cache_attention_bias =
+      modeConfig.kv_cache_attention_bias.default ?? null;
+  }
+  if (modeConfig.input_size) {
+    extracted.input_size = modeConfig.input_size.default ?? null;
+  }
+  if (modeConfig.vae_strategy) {
+    extracted.vae_strategy = modeConfig.vae_strategy.default ?? null;
+  }
```
leszko: Do we need to assign here field by field, cannot we just iterate like it's down below?

frontend/src/pages/StreamPage.tsx
```diff
@@ -143,19 +158,71 @@ export function StreamPage() {
...
+      updates.denoisingSteps = modeConfig.denoising_steps ?? undefined;
+      updates.noiseScale = modeConfig.noise_scale ?? undefined;
+      updates.noiseController = modeConfig.noise_controller ?? undefined;
+
+      // When switching to "video" mode, ensure videoSourceMode is set to "video"
+      // (unless it's already "video", in which case we don't need to change it)
+      if (videoSourceMode !== VIDEO_SOURCE_MODE.VIDEO) {
+        switchMode(VIDEO_SOURCE_MODE.VIDEO);
+      }
+    }
```
leszko: I think this should be refactored. LIke some of the code is executed for both if conditions, so it should be outside if. E.g.
```
updates.denoisingSteps = modeConfig.denoising_steps ?? undefined;
```

frontend/src/types/index.ts
```diff
@@ -30,17 +38,28 @@ export interface LoRAConfig {
...
+/**
+ * Settings state for stream configuration.
+ *
+ * Convention for optional fields:
+ * - undefined = not loaded yet / not set by user (waiting for schema)
+ * - null = explicitly not applicable / disabled for current mode
+ */
 export interface SettingsState {
   pipelineId: PipelineId;
+  generationMode?: GenerationMode;
```
leszko: It's called everywhere the `generation mode`, but isn't it actually the `input mode`. Like V2V or T2V.

src/scope/core/pipelines/base/vae/factory.py
```diff
@@ -0,0 +1,114 @@
...
+    "longlive": ("scope.core.pipelines.base.vae.longlive", "LongLiveVAE"),
+    "streamdiffusionv2": (
+        "scope.core.pipelines.base.vae.streamdiffusionv2",
+        "StreamDiffusionV2VAE",
+    ),
+    "streamdiffusionv2_longlive_scaled": (
+        "scope.core.pipelines.base.vae.streamdiffusionv2",
+        "StreamDiffusionV2VAEWithLongLiveScaling",
+    ),
+}
```
leszko, line 20: So, is the idea that we put all VAEs into `base`, but keep naming then with the pipeline name, e.g. `StreamDiffusionV2VAE`?

I think the other option would be to keep the streamdiffusionv2 vae inside pipelines/streamdiffusionv2. Have we considered it? Then, the registry would just be outside `pipelines.`

src/scope/core/pipelines/interface.py
```diff
@@ -67,6 +68,30 @@ def get_schema(cls) -> dict[str, Any]:
         """
         pass
 
+    def prepare(self, **kwargs) -> Requirements | None:
```
leszko: Do we actually need it now? Maybe we can get the requirements from the defaults and then get rid of the `prepare()` function? (yeah, I l know we're discussing this prepare function too much 🙃 )

src/scope/core/pipelines/krea_realtime_video/pipeline.py
```diff
@@ -130,52 +134,34 @@ def __init__(
...
-        components_config["device"] = device
-        components_config["dtype"] = dtype
-
-        components = ComponentsManager(components_config)
-        components.add("generator", generator)
-        components.add("scheduler", generator.get_scheduler())
-        components.add("vae", vae)
-        components.add("text_encoder", text_encoder)
-
-        embedding_blender = EmbeddingBlender(
```
leszko, line 154: Nice to see so much code removed 🚀

src/scope/core/pipelines/wan2_1/blocks/prepare_video_latents.py
```diff
@@ -77,6 +80,22 @@ def intermediate_outputs(self) -> list[OutputParam]:
...
 
+        # Align video latents with the generator's dtype and device so that downstream
+        # convolutions (e.g. patch_embedding) always receive tensors compatible with
+        # model parameters. This mirrors the behaviour of PrepareLatentsBlock.
+        generator_param = next(components.generator.model.parameters())
+
+        # If no video is provided, skip video latent preparation.
+        # This allows pipelines to share the same block graph for both
+        # text-to-video and video-to-video workflows.
+        if not hasattr(block_state, "video") or block_state.video is None:
```
leszko: Isn't it a case that when we call the PrepareVideoLatents block then we always have the `video` input param? 🤔

src/scope/core/pipelines/mode_helpers.py
```diff
@@ -0,0 +1,304 @@
...
+"""
+
+from typing import Any
+
+import torch
+
+from .interface import Requirements
+
+
+class UniversalInputModesMixin:
```
leszko: Do we actually this need to be a mixin? Cannot it just be a util class with helper functions?

src/scope/core/pipelines/streamdiffusionv2/modular_blocks.py
ryanontheinside: Just updated with a second draft that i think is more in the direction we want.

frontend/src/types/index.ts
ryanontheinside: There was some confusion here as 'camera' was also an input mode. This was meant to differentiate them, but I resolved instead by divorcing camera from this concept. Renamed!

src/scope/core/pipelines/base/vae/factory.py
ryanontheinside: This currently essentially just wraps the low level vae from the pipeline, so it's already doing what you suggest unless I am misunderstanding

ryanontheinside: @yondonfu thoughts?

tests/test_multi_mode_pipeline.py
```diff
@@ -0,0 +1,419 @@
+"""Tests for MultiModePipeline base class and self-configuring blocks.
```
leszko: I think that if we want to include the test, I'd probably keep the convention we have in the pipelines, so instead of having this mocked here, then create additional tests like: `src/scope/core/pipelines/streamdiffusionv2/test_t2v.py`. But maybe we don't need these tests at all 🙃

src/scope/core/pipelines/multi_mode.py
```diff
@@ -0,0 +1,285 @@
...
+    resolve_input_mode,
+)
+from .helpers import initialize_state_from_config
+from .interface import Pipeline, Requirements
+from .process import postprocess_chunk
+
+logger = logging.getLogger(__name__)
+
+
+class MultiModePipeline(Pipeline):
```
leszko, line 30: I wonder if we need this abstraction if we specify the other blocks as `AutoPipelineBlocks`. IFAIU if we use auto blocks, then they will be automatically selected based on the input type. 

https://huggingface.co/docs/diffusers/main/en/modular_diffusers/auto_pipeline_blocks

tests/test_multi_mode_pipeline.py
ryanontheinside: id just as soon remove it

src/scope/core/pipelines/multi_mode.py
ryanontheinside: more like [this](https://github.com/daydreamlive/scope/blob/df127c1665f1abe333df127a89f77662b7cabc6a/src/scope/core/pipelines/longlive/auto_blocks.py#L144)?

ryanontheinside:

Issue #184, opened 2025-11-24
[Bug Report]: Krea freezes on startup on H100


ericxtang:
### Summary

I ran into this error while trying to run the Krea model.  This results in the output video being SUPER slow.

https://github.com/user-attachments/assets/150d634c-f914-42b8-bf76-498d7d7d9570

Error:
```
2025-11-24T06:50:03.707696092Z 2025-11-24 06:50:03,706 - scope.server.frame_processor - ERROR - Error processing chunk: LoweringException: AttributeError: 'ShapeAsConstantBuffer' object has no attribute 'dtype'
2025-11-24T06:50:03.707700975Z   target: flex_attention
2025-11-24T06:50:03.707705647Z   args[0]: TensorBox(StorageBox(
2025-11-24T06:50:03.707712082Z     ComputedBuffer(name='buf5', layout=FlexibleLayout('cuda:0', torch.bfloat16, size=[1, 40, 128*CeilToInt(IntTrueDiv(Max(2160, -s27 + s30 + s47 - Max(0, -s27 + s30 + s47 - 2160) + 2160), 128)), 128], stride=[655360*CeilToInt(IntTrueDiv(Max(2160, -s27 + s30 + s47 - Max(0, -s27 + s30 + s47 - 2160) + 2160), 128)), 16384*CeilToInt(IntTrueDiv(Max(2160, -s27 + s30 + s47 - Max(0, -s27 + s30 + s47 - 2160) + 2160), 128)), 128, 1]), data=Pointwise(device=device(type='cuda', index=0), dtype=torch.bfloat16, inner_fn=<function BaseView.make_loader.<locals>.loader at 0x759fcdee3d00>, ranges=[1, 40, 128*CeilToInt(IntTrueDiv(Max(2160, -s27 + s30 + s47 - Max(0, -s27 + s30 + s47 - 2160) + 2160), 128)), 128]))
2025-11-24T06:50:03.707717092Z   ))
2025-11-24T06:50:03.707721767Z   args[1]: TensorBox(StorageBox(
[...]

PR #192, opened 2025-11-26
Fix to AttributeError: 'ShapeAsConstantBuffer' dtype error with compiling

BuffMcBigHuge:
## Fix: Resolve `ShapeAsConstantBuffer` dtype error when compiling flex_attention with score_mod

Reference issue:
- https://github.com/daydreamlive/scope/issues/184

> Note: This was tested on a RunPod 5090 by forcing `compile=True` in the KreaRealtimeVideoPipeline and has not been tested on an H100 Hopper as of yet.

### Problem
When `compile=True` is used with `KreaRealtimeVideoPipeline`, compilation fails with:
```
LoweringException: AttributeError: 'ShapeAsConstantBuffer' object has no attribute 'dtype'
  target: flex_attention
```

This occurs in the denoise block when `flex_attention` is called with a `score_mod` function.

### Root Cause
In `score_mod` (used for KV cache attention bias), Python scalars (`cache_current_block_start` as int, `log_scale` as float) were used directly. With `torch.compile`, these become `ShapeAsConstantBuffer` objects that lack a `dtype`, causing the error.

Related issues:
- https://github.com/pytorch/pytorch/issues/157833
- https://github.com/pytorch/pytorch/issues/129550
- https://github.com/pytorch/pytorch/issues/150056

### Solution
Convert all scalars used in `score_mod` to tensors before compilation:
1. Always convert `cache_current_block_start` to a tensor (handles both tensor and int cases)
2. Convert `log_scale` to a tensor (`log_scale_tensor`) with the correct dtype and device
3. Use tensor versions in `score_mod` instead of Python scalars

This ensures all values in `score_mod` are tensors with `dtype`, preventing the `ShapeAsConstantBuffer` error.

### Changes
- Modified `causal_model.py` lines 541-568 to convert scalars to tensors before use in `score_mod`

### Testing
- Verified compilation succeeds with `compile=True`
- No functional changes; only fixes the compilation path

### DIFF
--- a/src/scope/core/pipelines/krea_realtime_video/modules/causal_model.py
+++ b/src/scope/core/pipelines/krea_realtime_video/modules/causal_model.py
@@ -540,21 +540,13 @@ class CausalWanSelfAttention(nn.Module):
 
                 # Convert scalars to tensors to avoid ShapeAsConstantBuffer dtype issues during compilation
                 # This is critical when using torch.compile with flex_attention
-                frame_seqlen_tensor = torch.tensor(
+                frame_seqlen_tensor = torch.as_tensor(
                     frame_seqlen, dtype=torch.int32, device=roped_query.device
                 )
-                # Always convert cache_current_block_start to a tensor to avoid ShapeAsConstantBuffer issues
-                if isinstance(cache_current_block_start, torch.Tensor):
-                    if cache_current_block_start.ndim > 0:
-                        cache_current_block_start_tensor = cache_current_block_start.squeeze()
-                    else:
-                        cache_current_block_start_tensor = cache_current_block_start
-                else:
-                    cache_current_block_start_tensor = torch.tensor(
-                        cache_current_block_start, dtype=torch.int32, device=roped_query.device
-                    )
-                # Convert log_scale to a tensor to avoid ShapeAsConstantBuffer dtype issues
-                log_scale_tensor = torch.tensor(
+                cache_current_block_start_tensor = torch.as_tensor(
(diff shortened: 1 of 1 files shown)


src/scope/core/pipelines/krea_realtime_video/modules/causal_model.py
```diff
@@ -539,21 +539,31 @@ def qkv_fn(x):
...
                 frame_seqlen_tensor = torch.tensor(
                     frame_seqlen, dtype=torch.int32, device=roped_query.device
                 )
-                # cache_current_block_start might already be a tensor([0]) based on the way local_end_index is initialized
-                # in the KV cache but we want a scalar tensor so we check for that case and convert if needed
-                cache_current_block_start_tensor = cache_current_block_start
-                if isinstance(cache_current_block_start_tensor, torch.Tensor) and cache_current_block_start_tensor.ndim > 0:
-                    cache_current_block_start_tensor = cache_current_block_start_tensor.squeeze()
+                # Always convert cache_current_block_start to a tensor to avoid ShapeAsConstantBuffer issues
+                if isinstance(cache_current_block_start, torch.Tensor):
```
yondonfu: Can we use the simpler approach with `torch.as_tensor()` you used in https://github.com/daydreamlive/scope/pull/186#discussion_r2565919821 instead of the series of conditionals here?

yondonfu: Addressed in https://github.com/daydreamlive/scope/pull/192/commits/9e4e3c58dffde61f0c21e8c56422c25a23651b36

Issue #56, opened 2025-10-27
Add support for recording


yondonfu:
<img width="832" height="272" alt="Image" src="https://github.com/user-attachments/assets/9b908c47-9cc4-4099-ae33-50895e39b24a" />

The fourth button in the PlayerButton would be the recording button. By default, it would show a circle icon and when pressed recording would start switching the button to a square (stop) icon. At this point if the button is clicked recording would be stopped and ideally an mp4 of the recording would be downloaded.

An mp4 is desirable for maximum compatibility eg easy to upload anywhere else.

Given the mp4 requirement server-side recording might make more sense than in-browser recording since I think the latter doesn't handle mp4 well (worth a check). aiortc has a [MediaRecorder](https://aiortc.readthedocs.io/en/latest/helpers.html#media-sinks) class that might be helpful here.

EDIT: See below for updated proposal.

PR #193, opened 2025-11-27
Add recording

leszko:
# Env variables

### `RECORDING_ENABLED`
- **Default**: `true`
- **Description**: Enable/disable recording. When `false`, recording is disabled.

### `RECORDING_MAX_LENGTH`
- **Default**: `1h`
- **Description**: Maximum total recording length (sum of all segments). Recording stops when reached.
- **Format**: Supports `1h`, `30m`, `120s`, or plain seconds (e.g., `3600`)

### `RECORDING_STARTUP_CLEANUP_ENABLED`
- **Default**: `true`
- **Description**: Enable/disable cleanup of recording files 

--------
## Assumptions / Comments
- The recording stream is sourced from the same place we send frames to the frontend.
  - This means that if a frame is dropped, it will also be missing from the recording (sorry, Ryan).
  - The main benefit is that we can piggy-back on our existing PTS calculation mechanism, and `aiortc.contrib.media.MediaRecorder`.
- Temporary recordings are stored in the OS temp directory.
- Both stream pause/resume and exporting a recording cause the backend to segment/clip the stream. The final recording is a concatenation of these segments when downloaded. This is mainly due to how pause handling works and the behavior of MediaRecorder, which needs to finalize the stream before exporting.
- The pause/resume feature adds significant complexity to recording logic. On the backend we keep sending the same frame while paused, but we don't want repeated identical frames in the final recording.
Recording input may be tricky because of the pause/resume (we don't pause input stream!)
- We use ffmpeg (as Python library) for mp4 segments concatenation; I experimented with using AV library, but it happened to be surprisingly difficult (a lot of corner cases, etc.), so I ended up with ffmpeg. AV is still an option, but required more work and maintenance. One other option is to use the moviepy library
- MAX_FPS is decreased to `30` because `60` FPS won't work with MediaRecorder, because MediaRecorder calculates PTS/DTS on its own and if we start delivering frames at `60FPS`, it fails because PTS/DTS are duplicated

--------

## Demo with comments

1. Paused video is visible in the recording
2. When the recording is downloaded the second time, then it has the length of the whole video, but the Start Time is from the last recording time. So depending on the player you may see a blank screen at the beginning (you see in Quick Time, but VLC or ffplay respects the Start Time and starts from where the video started). e.g.:
- Play stream, after 00:10:00, you download the recording; stream keeps playing
- After another 00:10:00, you click download again,
- The results stream has Duration 00:20:00 and Start Time 00:10:00
- QuickTime does not respect Start Time, so you'll see first 00:10:00 blank black screen

https://github.com/user-attachments/assets/b546df3e-a7be-4768-8a14-8f14b2a31797

--------

fix https://github.com/daydreamlive/scope/issues/56

### DIFF
--- a/frontend/src/components/ExportDialog.tsx
+++ b/frontend/src/components/ExportDialog.tsx
@@ -39,7 +39,7 @@ export function ExportDialog({
               onClose();
             }}
             variant="outline"
-            className="w-full justify-start gap-2"
+            className="w-full justify-start gap-3 px-4 py-6"
           >
             <Download className="h-4 w-4" />
             <div className="flex flex-col items-start">
@@ -56,7 +56,7 @@ export function ExportDialog({
               onClose();
             }}
             variant="outline"
-            className="w-full justify-start gap-2"
+            className="w-full justify-start gap-3 px-4 py-6"
           >
             <Download className="h-4 w-4" />
             <div className="flex flex-col items-start">


pyproject.toml
```diff
@@ -30,6 +30,7 @@ classifiers = [
 dependencies = [
     "aiortc>=1.13.0",
     "fastapi>=0.116.1",
+    "ffmpeg-python>=0.2.0",
```
yondonfu: Can you confirm that this does not require the user to install a system copy of ffmpeg separately?

src/scope/server/app.py
```diff
@@ -331,6 +339,42 @@ async def handle_webrtc_offer(
         raise HTTPException(status_code=500, detail=str(e)) from e
 
 
+@app.get("/api/v1/recording/download")
```
yondonfu: Can we be more REST compliant (we need some cleanup elsewhere too)?

Eg

GET /api/v1/recordings/{session_id}

And since the WebRTC offer resp includes a session ID that would be used here?

src/scope/server/webrtc.py
```diff
@@ -172,7 +175,19 @@ async def handle_offer(
...
             )
             session.video_track = video_track
-            pc.addTrack(video_track)
+
+            # Create a MediaRelay to allow multiple consumers (WebRTC and recording)
+            relay = MediaRelay()
+            relayed_track = relay.subscribe(video_track)
+
+            # Set the relay on the video track so it can create a recording track
+            video_track.recording_manager.set_relay(relay)
```
yondonfu: What about a recording_track on the session that uses the RecordingManager (instead of tracking RecordingManager in the video_track) since it seems it can be largely decoupled from the video_track since the MediaRelay should just feed frames to the recording_track?

The RecordingTrack could have a pause fn and then when a pause param update is received could call pause() on both the video_track and recording_track stored with the session.

src/scope/server/recording.py
```diff
@@ -0,0 +1,753 @@
...
+
+
+def cleanup_temp_file(file_path: str):
+    """Clean up temporary file after download."""
+    if os.path.exists(file_path):
+        RecordingManager._safe_remove_file(file_path)
+        logger.info(f"Cleaned up temporary download file: {file_path}")
+
+
+async def get_recording_for_download(webrtc_manager):
```
yondonfu: If we go with the /api/v1/recordings/{session_id} API format I think the handler could take care of getting the session using the ID and then calling the finalize fn on the recording track instead of having this logic here?

src/scope/server/app.py
leszko: Updated

src/scope/server/recording.py
leszko: Updated

pyproject.toml
leszko: You **do** need the FFmpeg binary installed separately, since `ffmpeg-python` is just a wrapper. I’ve previously tried replacing it with other libraries, but it wasn’t very straightforward. That said, we only use FFmpeg for MP4 concatenation, so replacing it should be doable.

I’ll try again to remove `ffmpeg-python` and evaluate alternatives. Some options that avoid concatenation entirely (mentioning them here, since maybe we'll like them more and it'd solve the `ffmpeg-python` issue):
1. Return multiple MP4 files (listed in the UI or zipped on the backend if the stream was paused).
2. Rethink the “pause” feature, since this issue only affects that path.
3. Adopt a WYSIWYG approach: the recording matches exactly what the user sees, including pause, so some frames will be repeated.
4. Revisit recording delivery: instead of enforcing timeline timing, record frames and let the user choose the FPS when constructing the final video.

src/scope/server/webrtc.py
leszko: Yeah, that's fine. I moved RecordingManager to session.

pyproject.toml
yondonfu: I'm inclined to do 3 in this PR if it means an easy way to avoid requiring system ffmpeg installation and then take a separate pass at improving how we handle recording (4 feels promising).

leszko: > I'm inclined to do 3 in this PR if it means an easy way to avoid requiring system ffmpeg installation and then take a separate pass at improving how we handle recording (4 feels promising).

Ok, let me do some time-boxed exercise (a few hours max) and if getting rid of `ffmpeg` is not trivial, then I'll just do 3.

leszko: I've spent some time exploring MP4 concatenation and ultimately decided to go with **Option 3**.  

Some notes:
- MP4 concatenation turned out to be significantly more complex than I initially expected.  
- I do have a working pure-Python concatenation branch, but the Claude-generated implementation is very long and complex (>1k lines) and still doesn't handle certain edge cases (e.g., varying FPS, long mp4 files).  
- My suggestion is to revisit this if a real need arises. For now, maintaining AI-generated MP4 concatenation code seems unnecessary.  

See the PR Description and the attached demo to see how it works now.

leszko: I wrote it too fast. Actually I realized `MediaRecorder` depends on ffmpeg, so we need ffmpeg anyway...

yondonfu: I don't see an attached demo anywhere.

leszko: Sorry, added now with the comment. PTAL.

PR #203, opened 2025-11-28
01 feat: add shared VAE infrastructure 

ryanontheinside:
Add `base/vae` module with factory pattern for VAE selection.

- New VAE implementations: `KreaRealtimeVideoVAE`, `LongLiveVAE`, `StreamDiffusionV2VAE`
- Factory function `create_vae()` for strategy-based instantiation
- Shared constants for latent normalization

Purely additive - no existing code modified.

Will be utilized in following PRs.

### DIFF
--- a/src/scope/core/pipelines/wan2_1/vae/constants.py
+++ b/src/scope/core/pipelines/wan2_1/vae/constants.py
@@ -1,0 +1,46 @@
+"""Shared VAE normalization constants for Wan2.1 models.
+
+These constants define the latent space normalization parameters used across
+all Wan2.1-based VAE wrappers to ensure consistent encoding/decoding behavior.
+"""
+
+# Wan2.1 VAE latent space normalization parameters
+# Mean values for each of the 16 latent channels
+WAN_VAE_LATENT_MEAN = [
+    -0.7571,
+    -0.7089,
+    -0.9113,
+    0.1075,
+    -0.1745,
+    0.9653,
+    -0.1517,
+    1.5508,
+    0.4134,
+    -0.0715,
+    0.5517,
+    -0.3632,
+    -0.1922,
+    -0.9497,
+    0.2503,
+    -0.2921,
+]
+
+# Standard deviation values for each of the 16 latent channels
+WAN_VAE_LATENT_STD = [
+    2.8184,
+    1.4541,
+    2.3275,
+    2.6558,
+    1.2196,
+    1.7708,
+    2.6052,
+    2.0743,
+    3.2687,
+    2.1526,
+    2.8652,
+    1.5579,
+    1.6382,
+    1.1253,
+    2.8251,
+    1.9160,
+]
--- a/src/scope/core/pipelines/wan2_1/vae/wan.py
+++ b/src/scope/core/pipelines/wan2_1/vae/wan.py
@@ -1,0 +1,173 @@
+"""Unified Wan VAE wrapper with streaming and batch encoding/decoding."""
+
+import os
+
+import torch
+
+from .constants import WAN_VAE_LATENT_MEAN, WAN_VAE_LATENT_STD
+from .modules.vae import _video_vae
+
+# Default filename for standard Wan2.1 VAE checkpoint
+DEFAULT_VAE_FILENAME = "Wan2.1_VAE.pth"
+
+
+class WanVAEWrapper(torch.nn.Module):
+    """Unified VAE wrapper for Wan2.1 models.
+
+    This VAE supports both streaming (cached) and batch encoding/decoding modes.
+    Normalization is always applied during encoding for consistent latent distributions.
+    """
+
+    def __init__(
+        self,
+        model_dir: str = "wan_models",
+        model_name: str = "Wan2.1-T2V-1.3B",
+        vae_path: str | None = None,
+    ):
+        super().__init__()
+
+        # Determine paths with priority: explicit vae_path > model_dir/model_name default
+        if vae_path is None:
+            vae_path = os.path.join(model_dir, model_name, DEFAULT_VAE_FILENAME)
+
+        self.register_buffer(
+            "mean", torch.tensor(WAN_VAE_LATENT_MEAN, dtype=torch.float32)
+        )
+        self.register_buffer(
+            "std", torch.tensor(WAN_VAE_LATENT_STD, dtype=torch.float32)
+        )
+        self.z_dim = 16
+
+        self.model = (
+            _video_vae(
+                pretrained_path=vae_path,
+                z_dim=self.z_dim,
+            )
+            .eval()
+            .requires_grad_(False)
+        )
+
+    def _get_scale(self, device: torch.device, dtype: torch.dtype) -> list:
+        """Get normalization scale parameters on the correct device/dtype."""
+        return [
+            self.mean.to(device=device, dtype=dtype),
+            1.0 / self.std.to(device=device, dtype=dtype),
+        ]
+
+    def _apply_encoding_normalization(
+        self, latent: torch.Tensor, scale: list
+    ) -> torch.Tensor:
+        """Apply normalization to encoded latents."""
+        if isinstance(scale[0], torch.Tensor):
+            return (latent - scale[0].view(1, self.z_dim, 1, 1, 1)) * scale[1].view(
+                1, self.z_dim, 1, 1, 1
+            )
+        return (latent - scale[0]) * scale[1]
+
+    def _create_encoder_cache(self) -> list:
+        """Create a fresh encoder feature cache."""
+        return [None] * 55
+
+    def encode_to_latent(
+        self,
+        pixel: torch.Tensor,
+        use_cache: bool = True,
+        feat_cache: list | None = None,
+    ) -> torch.Tensor:
+        """Encode video pixels to latents.
+
+        Args:
+            pixel: Input video tensor [batch, channels, frames, height, width]
+            use_cache: If True, use streaming encode (maintains cache state).
+                      If False, use batch encode with a temporary cache.
+            feat_cache: Optional external cache. If provided, this cache is used
+                       instead of the internal streaming cache, allowing one-time
+                       encodes without affecting the streaming state (e.g., for
+                       re-encoding the first frame in Krea pipeline).
+
+        Returns:
+            Latent tensor [batch, frames, channels, height, width]
+        """
+        device, dtype = pixel.device, pixel.dtype
+        scale = self._get_scale(device, dtype)
+
+        if use_cache and feat_cache is None:
+            # Streaming encode - cache is maintained across calls
+            latent = self.model.stream_encode(pixel)
+            # Apply normalization (stream_encode returns unnormalized)
+            latent = self._apply_encoding_normalization(latent, scale)
+        else:
+            # Batch encode with one-time cache (does not affect streaming state)
+            # Use provided cache or create a temporary one
+            cache = (
+                feat_cache if feat_cache is not None else self._create_encoder_cache()
+            )
+            latent = self._encode_with_cache(pixel, scale, cache)
+
+        # [batch, channels, frames, h, w] -> [batch, frames, channels, h, w]
+        return latent.permute(0, 2, 1, 3, 4)
+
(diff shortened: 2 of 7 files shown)


src/scope/core/pipelines/base/vae/streamdiffusionv2.py
```diff
@@ -0,0 +1,111 @@
...
+        output = self.model.stream_decode(zs, scale).float().clamp_(-1, 1)
+        output = output.permute(0, 2, 1, 3, 4)
+        return output
+
+    def clear_cache(self):
+        """Clear decoder cache for next sequence."""
+        self.model.first_batch = True
+
+
+class StreamDiffusionV2VAEWithLongLiveScaling(StreamDiffusionV2VAE):
```
ryanontheinside: I'm not overly thrilled with this. As I work on LightVAE I find that I will have to do the same thing, subclass it for longlive's scaling. Then I imagine a world where we add additional models with further scaling requirements. A couple of alternative solutions might be:

- Callback approach: Pass a callable encoding_scaling_fn (or None). Define standard functions like longlive_encoding_scaling() in the module. 

- Enum + registry: EncodingScalingMode enum with a registry mapping to functions. More discoverable and type-safe, but adds complexity

leszko: Does it mean that `StreamDiffusionV2VAE` does not work with LongLive? If yes, then it makes me think that VAE are not interchangeable between different pipelines. If that's the case, I wonder if we couldn't "make" them interchangeable, so that each VAE is possible to apply to every pipeline. Some ideas:
1. (if possible) Detect (here or in the related block) if the normalization is required and if yes, then apply it; Then we would have only `StreamDiffusionV2VAE`
2. Have a param (here or in the related block) to decide if the normalization should happen

src/scope/core/pipelines/base/vae/krea_realtime_video.py
```diff
@@ -0,0 +1,209 @@
...
+        self.encoder = VAEEncoderWrapper(vae)
+        self.encoder.eval()
+        self.encoder.requires_grad_(False)
+
+        self.decoder = VAEDecoderWrapper()
+        state_dict = torch.load(vae_path, map_location="cpu")
+        decoder_state_dict = {}
+        for key, value in state_dict.items():
+            if "decoder." in key or "conv2" in key:
+                decoder_state_dict[key] = value
```
leszko: What is this code block doing? The variable `decoder_state_dict` does not seem to be used anywhere.

src/scope/core/pipelines/base/vae/krea_realtime_video.py
```diff
@@ -0,0 +1,209 @@
...
+    This VAE uses a specialized architecture optimized for real-time video
+    generation with cached encoding and decoding.
+    """
+
+    def __init__(
+        self,
+        model_name: str = "Wan2.1-T2V-1.3B",
+        model_dir: str | None = None,
+        vae_path: str | None = None,
+    ):
```
leszko: I wonder if we should maybe make all the VAE constructors looking the same. I see that in this one you can define `model_name`. `model_dir`, and `vae_path`. At the same time in `LongLiveVAE` you have only 1 parameter `model_dir` and the rest is hardcoded.

src/scope/core/pipelines/base/vae/krea_realtime_video.py
```diff
@@ -0,0 +1,209 @@
...
+        """
+        output, self.decoder_cache = self.decoder(latent, *self.decoder_cache)
+        return output
+
+    def clear_cache(self):
+        """Clear decoder cache for next sequence."""
+        self.decoder_cache = [None] * 55
+
+
+class VAEEncoderWrapper(nn.Module):
```
leszko: Sorry, super stupid question, why for the Krea we need to write our own wrapper while for other pipelines we use the already implemented `WanVAE_()`?

And super nit: if this wrapper is potentially reusable, I'd move it out of this file.

leszko (commented):
Very nice PR @ryanontheinside . Added some comments. My main points are:

1. I asked this elsewhere already, so apologies if you’ve responded before.) Just to confirm: we want to move all VAEs into the `pipelines/base/vae` directory instead of keeping them alongside their respective pipelines, right? I’m fine with that, but it does feel a bit counter-intuitive - especially as we continue adding more pipelines. That’s why I wanted to double-check.
2. My understanding is that the VAEs aren’t currently interchangeable across pipelines - for example, LongLive requires latent normalization, and Krea uses its own VAE wrapper. Is that correct? And if so, do we eventually want to make the VAEs interchangeable?

src/scope/core/pipelines/base/vae/streamdiffusionv2.py
ryanontheinside: > Does it mean that `StreamDiffusionV2VAE` does not work with LongLive? If yes, then it makes me think that VAE are not interchangeable between different pipelines.

Correct - StreamDiffusionV2VAE doesn't work directly with LongLive because it doesn't apply the latent normalization that LongLive expects. The goal is to make VAEs interchangeable when possible - this strategy pattern is a first step, and the subclass for LongLive scaling is a stopgap until we decide on a better approach. I prefer the callback approach I mentioned earlier, since other models may have different scaling requirements. I expect we'll see other common patterns across VAE implementations as we add new models and VAEs.

src/scope/core/pipelines/base/vae/krea_realtime_video.py
ryanontheinside: > Sorry, super stupid question, why for the Krea we need to write our own wrapper while for other pipelines we use the already implemented `WanVAE_()`?
> 
> And super nit: if this wrapper is potentially reusable, I'd move it out of this file.

We need the wrappers because unlike the other models Krea splits the encoder and decoder with separate caches. They aren't reusable elsewhere yet, so we could leave them for now.

src/scope/core/pipelines/base/vae/krea_realtime_video.py
ryanontheinside: > I wonder if we should maybe make all the VAE constructors looking the same.

I agree we should standardize the constructors

src/scope/core/pipelines/base/vae/krea_realtime_video.py
ryanontheinside: > What is this code block doing? The variable `decoder_state_dict` does not seem to be used anywhere.

this is a bug - the variable is vestigial and never used. I'll fix it.

src/scope/core/pipelines/wan2_1/vae/constants.py
```diff
@@ -0,0 +1,46 @@
+"""Shared VAE normalization constants for Wan2.1 models.
```
yondonfu, line 1: The fact that this is specific to Wan2.1 makes me feel like this should be under `wan2_1`.

src/scope/core/pipelines/base/vae/krea_realtime_video.py
```diff
@@ -0,0 +1,209 @@
+"""KreaRealtimeVideo VAE wrapper with encoder/decoder separation."""
+
+import os
+
+import torch
+import torch.nn as nn
+
+from ...krea_realtime_video.modules.vae import CausalConv3d, WanVAE
```
yondonfu: As a convention, IMO we should avoid ... imports and only do .. imports. If a ... is needed an absolute import could be clearer eg `from scope.core.pipelines.krea_realtime_video...`. WDYT about this convention? We should just decide on one and stick with it.

src/scope/core/pipelines/wan2_1/vae/constants.py
ryanontheinside: i agree

src/scope/core/pipelines/base/vae/krea_realtime_video.py
ryanontheinside: works for me

ryanontheinside:
> I'm wondering whether we can avoid having to maintain all of these VAE comp/module impls.
> 
> AFAICT (not the most thorough review yet) the impls are very similar minus a few highlights:
> 
> * krea-realtime-video uses an external cache that is always expected to be passed in and the impl returns the updated cache along with results
> * streamdiffusionv2 is missing a normalization step during encoding that the other impls have - should that just be there always?
> 
> I created this [spike](https://github.com/daydreamlive/scope/commit/ae772be997ac7cda07966dac252179f0ddeb74e5#r171740317) where every pipeline uses a modified version of the streamdiffusionv2 VAE comp/module impls and while I haven't tested thoroughly/benchmarked properly it does like all the pipelines work without any major noticeable regression. And streamdiffusionv2 output looks...different (possibly better?) with the added normalization step during encoding.

Your spike validates consolidating the Wan VAE implementations - agree we should do that, much more simple and elegant! I thought the normalization being missing from sd was intentional but this simplifies things further. Excited to try it as I would assume it will improve the quality of the output too.

The factory pattern will still be useful for LightVAE which is a fundamentally different model but same interface. See the WIP branch: #208 
Im thinking plan:
This PR: 
- Consolidate Wan VAEs into one implementation (per your spike)
- Move to wan directory
- Keep factory
Follow-up PR: Add LightVAE as a factory option

WDYT?

src/scope/core/pipelines/wan2_1/vae/wan.py
```diff
@@ -0,0 +1,125 @@
...
+import os
+
+import torch
+
+from scope.core.pipelines.streamdiffusionv2.modules.vae import _video_vae
+
+from .constants import WAN_VAE_LATENT_MEAN, WAN_VAE_LATENT_STD
+
+
+class WanVAE(torch.nn.Module):
```
yondonfu: Let's keep the *Wrapper naming convention for now?

src/scope/core/pipelines/wan2_1/vae/wan.py
```diff
@@ -0,0 +1,125 @@
...
+        device, dtype = pixel.device, pixel.dtype
+        scale = self._get_scale(device, dtype)
+
+        if use_cache:
+            # Streaming encode - cache is maintained across calls
+            latent = self.model.stream_encode(pixel)
+            # Apply normalization (stream_encode returns unnormalized)
+            latent = self._apply_encoding_normalization(latent, scale)
+        else:
+            # Batch encode - cache is cleared, normalization applied internally
```
yondonfu: We don't want to clear the cache for Krea because the first frame re-encode shouldn't affect the cache used for streaming encodes. The way I handled this in the spike branch I shared was to manually pass in a one-time use cache which should work, but open to other solutions.

src/scope/core/pipelines/wan2_1/vae/wan.py
```diff
@@ -0,0 +1,127 @@
+"""Unified Wan VAE wrapper with streaming and batch encoding/decoding."""
+
+import os
+
+import torch
+
+from scope.core.pipelines.streamdiffusionv2.modules.vae import _video_vae
```
yondonfu: Can we copy the module under `wan2_1/vae/modules`? Then, we can remove the per pipeline VAE modules in a later PR.

src/scope/core/pipelines/wan2_1/vae/wan.py
yondonfu: Bump.

src/scope/core/pipelines/wan2_1/vae/wan.py
yondonfu: Bump - do you intend to handle this separately? If so, just confirm.

yondonfu (changes requested):

src/scope/core/pipelines/wan2_1/vae/wan.py
ryanontheinside: i had,  but I am rebuilding branch 3 so better to do it here

src/scope/core/pipelines/wan2_1/vae/wan.py
ryanontheinside: My bad - i totally missed this one

PR #204, opened 2025-11-28
02 feat: schema infrastructure

ryanontheinside:
Depends upon #203 
Add infrastructure for declarative pipeline configuration.

- `MultiModePipeline` base class with text/video mode support
- Schema helpers (`build_pipeline_schema`, `defaults.py`)
- Auto-routing blocks (`AutoPrepareLatentsBlock`, `AutoPreprocessVideoBlock`)
- Non-breaking: `get_schema()` has default implementation

Existing pipelines unchanged.

### DIFF
--- a/src/scope/core/pipelines/__init__.py
+++ b/src/scope/core/pipelines/__init__.py
@@ -2,7 +2,8 @@
 
 
 def __getattr__(name):
-    """Lazy import for pipeline classes to avoid triggering heavy imports."""
+    """Lazy import for pipeline and config classes to avoid triggering heavy imports."""
+    # Pipeline classes
     if name == "LongLivePipeline":
         from .longlive.pipeline import LongLivePipeline
 
@@ -19,12 +20,40 @@ def __getattr__(name):
         from .passthrough.pipeline import PassthroughPipeline
 
         return PassthroughPipeline
+    # Config classes
+    elif name == "BasePipelineConfig":
+        from .schema import BasePipelineConfig
+
+        return BasePipelineConfig
+    elif name == "LongLiveConfig":
+        from .schema import LongLiveConfig
+
+        return LongLiveConfig
+    elif name == "StreamDiffusionV2Config":
+        from .schema import StreamDiffusionV2Config
+
+        return StreamDiffusionV2Config
+    elif name == "KreaRealtimeVideoConfig":
+        from .schema import KreaRealtimeVideoConfig
+
+        return KreaRealtimeVideoConfig
+    elif name == "PassthroughConfig":
+        from .schema import PassthroughConfig
+
+        return PassthroughConfig
     raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
 
 
 __all__ = [
+    # Pipeline classes
     "LongLivePipeline",
     "KreaRealtimeVideoPipeline",
     "StreamDiffusionV2Pipeline",
     "PassthroughPipeline",
+    # Config classes
+    "BasePipelineConfig",
+    "LongLiveConfig",
+    "StreamDiffusionV2Config",
+    "KreaRealtimeVideoConfig",
+    "PassthroughConfig",
 ]
--- a/src/scope/core/pipelines/interface.py
+++ b/src/scope/core/pipelines/interface.py
@@ -1,9 +1,13 @@
 """Base interface for all pipelines."""
 
 from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING
 
 import torch
 from pydantic import BaseModel
+
+if TYPE_CHECKING:
+    from .schema import BasePipelineConfig
 
 
 class Requirements(BaseModel):
@@ -13,7 +17,48 @@ class Requirements(BaseModel):
 
 
 class Pipeline(ABC):
-    """Abstract base class for all pipelines."""
+    """Abstract base class for all pipelines.
+
+    Pipelines must implement get_config_class() to return their Pydantic config model.
+    This enables:
+    - Validation via model_validate() / model_validate_json()
+    - JSON Schema generation via model_json_schema()
+    - Type-safe configuration access
+    - API introspection and automatic UI generation
+
+    See schema.py for the BasePipelineConfig model and pipeline-specific configs.
+    For multi-mode pipeline support (text/video), use multi_mode.MultiModePipeline
+    as the base class which provides declarative mode configuration.
+    """
+
+    @classmethod
+    def get_config_class(cls) -> type["BasePipelineConfig"]:
+        """Return the Pydantic config class for this pipeline.
+
+        The config class should inherit from BasePipelineConfig and define:
+        - pipeline_id: Unique identifier
+        - pipeline_name: Human-readable name
+        - pipeline_description: Capabilities description
+        - pipeline_version: Version string
+        - Default parameter values for the pipeline
+
+        Returns:
+            Pydantic config model class
+
+        Note:
+            Subclasses should override this method to return their config class.
+            The default implementation returns BasePipelineConfig.
+
+        Example:
+            from .schema import LongLiveConfig
+
+            @classmethod
+            def get_config_class(cls) -> type[BasePipelineConfig]:
+                return LongLiveConfig
+        """
+        from .schema import BasePipelineConfig
+
+        return BasePipelineConfig
 
     @abstractmethod
     def __call__(
--- a/src/scope/core/pipelines/multi_mode.py
+++ b/src/scope/core/pipelines/multi_mode.py
@@ -1,0 +1,241 @@
+"""Multi-mode pipeline base class for declarative pipeline definitions.
+
+This module provides the MultiModePipeline base class that centralizes
+orchestration logic for pipelines supporting multiple input modes (text, video, etc.).
+Pipelines declare their capabilities via class methods, and the base class handles
+all execution logic.
+"""
+
+import logging
+from abc import abstractmethod
+from typing import Any
+
+import torch
+from diffusers.modular_pipelines import PipelineState
+
+from .blending import EmbeddingBlender
+from .components import ComponentsManager
+from .defaults import (
+    apply_mode_defaults_to_state,
+    get_pipeline_config,
+    resolve_input_mode,
+)
+from .helpers import initialize_state_from_config
+from .interface import Pipeline, Requirements
+from .process import postprocess_chunk
+
+logger = logging.getLogger(__name__)
+
+
+class MultiModePipeline(Pipeline):
+    """Base class for multi-mode pipelines with declarative configuration.
+
+    This class centralizes orchestration logic for pipelines supporting multiple
+    input modes (text-to-video, video-to-video). Pipelines declare their
+    capabilities via class methods:
+
+    - get_config_class(): Returns Pydantic config model class
+    - get_blocks(): Returns single workflow with nested AutoPipelineBlocks
+    - get_components(): Declares component requirements
+
+    Architecture Pattern:
+    Uses nested AutoPipelineBlocks for input-based routing (e.g., AutoPrepareLatentsBlock
+    routes based on presence of 'video' input). MultiModePipeline complements this by:
+    - Resolving mode from inputs and applying mode-specific defaults
+    - Handling mode transitions and cache initialization
+    - Providing configuration and schema management
+
+    This eliminates duplicate workflows while maintaining clear separation of concerns.
+
+    The unified WanVAE handles both text-to-video and video-to-video modes via
+    the use_cache parameter at encode/decode time, eliminating the need for
+    mode-specific VAE strategies.
+
+    Example:
+        class MyPipeline(MultiModePipeline):
+            @classmethod
+            def get_config_class(cls) -> type[BasePipelineConfig]:
+                return MyPipelineConfig
+
+            @classmethod
+            def get_blocks(cls):
+                return MyWorkflow()
+
+            @classmethod
+            def get_components(cls) -> dict:
+                return {
+                    "generator": MyGenerator,
+                    "text_encoder": MyTextEncoder,
+                }
+    """
+
+    @classmethod
+    @abstractmethod
+    def get_blocks(cls):
+        """Return block graph for pipeline execution.
+
+        Can return either:
+        - SequentialPipelineBlocks: Single unified workflow with conditional execution
+        - AutoPipelineBlocks: Multiple workflows with automatic routing
+
+        Returns:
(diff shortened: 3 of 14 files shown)


src/scope/core/pipelines/wan2_1/blocks/auto_prepare_latents.py
```diff
@@ -0,0 +1,139 @@
...
+from .prepare_latents import PrepareLatentsBlock
+from .prepare_video_latents import PrepareVideoLatentsBlock
+from .preprocess_video import PreprocessVideoBlock
+
+
+class EmptyBlock(ModularPipelineBlocks):
+    """Empty block that does nothing. Used as a no-op placeholder in routing."""
+
+    def __call__(self, components, state):
+        return components, state
```
leszko: Why do we actually need this no-op block?

src/scope/core/pipelines/__init__.py
```diff
@@ -19,6 +19,10 @@ def __getattr__(name):
         from .passthrough.pipeline import PassthroughPipeline
 
         return PassthroughPipeline
+    elif name == "MultiModePipeline":
+        from .multi_mode import MultiModePipeline
+
+        return MultiModePipeline
```
leszko: Will this `MultiModePipeline` be a separate pipeline, in addition to the existing sdv2, longlive, and krea pipeline? I always somehow thought we would have the same sdv2, longlive, krea pipeline, but in their internals the blocks will change to accept video/no-video latents. So, now I'm confused 🙃 If this is correct and we'll have an additional pipeline, then could you explain it better in the PR description?

ryanontheinside: MultiModePipeline will not be a separate pipeline - it's an abstract base class that existing pipelines (LongLive, StreamDiffusionV2, Krea) will extend. The class name stays the same (e.g., `LongLivePipeline`), it will just inherit from `MultiModePipeline` instead of `Pipeline`. 

You're right that the routing happens in the blocks - the mode routing happens at two levels:

1. **Block-level routing (as you expected)**: The workflow uses diffusers' `AutoPipelineBlocks` which route internally based on input presence. For example:
   - `AutoPreprocessVideoBlock`: if 'video' input exists → preprocess it; otherwise → no-op
   - `AutoPrepareLatentsBlock`: if 'video' input exists → encode video to latents; otherwise → generate noise latents

2. **Pipeline-level orchestration**: `MultiModePipeline` handles the surrounding concerns:
   - Resolving which mode is active from inputs
   - Applying mode-specific defaults (different denoising steps, resolution, noise_scale per mode)
   - Lazy-loading mode-specific components (e.g., different VAE strategies for T2V vs V2V)
   - Managing mode transitions and cache resets

The thinking is that blocks are stateless execution units - they should focus on "how do I do this operation?", not "what mode am I in?" or "which VAE should be loaded?". Pushing orchestration into blocks would mean every block needs mode awareness, component management logic, and transition handling - violating SR, creatign duplication, and making blocks harder to reuse.

See #205  for the concrete example: `LongLivePipeline` extends `MultiModePipeline` and uses a single `LongLiveWorkflow` that contains the `AutoPipelineBlocks` for routing. One pipeline class, one workflow, automatic T2V/V2V switching based on whether you pass `video` input.

src/scope/core/pipelines/wan2_1/blocks/auto_prepare_latents.py
ryanontheinside: You raise a valid point AutoPipelineBlocks will automatically skip when no trigger matches. So EmptyBlock isn't strictly required. I kept it explicit for easier mental model - that said, we can remove it if the preference is relying on the automatic skip.

src/scope/core/pipelines/wan2_1/blocks/auto_prepare_latents.py
```diff
@@ -0,0 +1,139 @@
...
+    @property
+    def description(self):
+        return (
+            "AutoPreprocessVideoBlock: Routes video preprocessing before cache setup:\n"
+            " - Routes to video preprocessing when 'video' input is provided\n"
+            " - Skips preprocessing when no 'video' input is provided\n"
+        )
+
+
+class TextToVideoLatentWorkflow(SequentialPipelineBlocks):
```
yondonfu: This and VideoToVideoLatentWorkflow just contain a single block. Can we omit them altogether and use PrepareLatentsBLock and PrepareVideoLatentsBlock directly in AutoPrepareLatentsBlock?

src/scope/core/pipelines/wan2_1/blocks/auto_prepare_latents.py
yondonfu: Slight pref not to include EmptyBlock mainly because if we stay consistent with that pattern then every time we need to do a AutoPipelineBlock where there is a no-op we have to explicitly include this which feels a bit cumbersome.

src/scope/core/pipelines/interface.py
```diff
@@ -13,7 +14,64 @@ class Requirements(BaseModel):
...
+    def get_schema(cls) -> dict[str, Any]:
+        """Return complete pipeline schema with metadata and mode configurations.
+
+        The schema should be a dictionary compatible with JSON Schema and OpenAPI
+        conventions, containing:
+        - id: Unique pipeline identifier
+        - name: Human-readable pipeline name
+        - description: Pipeline capabilities description
+        - version: Pipeline version string
+        - native_mode: Native input mode ("text" or "video")
```
yondonfu: Why do we need a native_mode? It seems that a pipeline just needs to be able to signal the modes it supports.

src/scope/core/pipelines/schema.py
```diff
@@ -0,0 +1,268 @@
...
+                default=value,
+            )
+        else:
+            # For unknown types, store as-is
+            config[key] = value
+
+    return config
+
+
+def compute_capabilities(
```
yondonfu: Do we need to introduce capabilities right now? I think everything here can be derived or assumed for the time being.

hasInputModeControl -> if the pipeline supports > 1 mode then this is supported
hasNoiseControls -> if the pipeline supports video input this is supported
showNoiseControlsInText ->if the pipeline supports text input then show all relevant controls for that
showNoiseControlsInVideo -> if the pipeline supports video input then show all relevant controls for that
hasCacheManagement -> besides passthrough which is really a dummy pipeline everything has a KV cache right now and if we need to add something without it then we can address when we have a better sense of what we're designing for
requiresVideoInVideoMode -> not sure what this is for?

src/scope/core/pipelines/interface.py
```diff
@@ -13,7 +14,64 @@ class Requirements(BaseModel):
...
+        - supported_modes: List of supported input modes
```
yondonfu: If there is already a dict with mode names can supported_modes be skipped since the info on which modes are supported is indicated by the keys of the dict already?

src/scope/core/pipelines/multi_mode.py
```diff
@@ -0,0 +1,260 @@
...
+        This checks for explicit input_mode parameter first, then falls back
+        to the native mode from schema.
+
+        Args:
+            kwargs: Generation parameters
+
+        Returns:
+            Inferred mode name (e.g., "text", "video")
+        """
+        return resolve_input_mode(kwargs.get("input_mode"), kwargs, self.__class__)
```
yondonfu: I feel like this can be simplified right now where there is no need for the input mode to be passed and just based on whether video is passed or not you can infer that you are in video vs. text mode.

src/scope/core/pipelines/multi_mode.py
```diff
@@ -0,0 +1,260 @@
...
+
+        Returns:
+            Post-processed output tensor
+        """
+        mode = resolve_input_mode(self.state.get("input_mode"), kwargs, self.__class__)
+
+        apply_mode_defaults_to_state(self.state, self.__class__, mode, kwargs)
+
+        # Store resolved mode in state for blocks that need mode-specific behavior
+        self.state.set("_resolved_mode", mode)
```
yondonfu: Can we avoid using this? Seems like with AutoPipelineBlocks and the ability to check the presence of the video param that should be sufficient for mode specific behavior.

src/scope/core/pipelines/multi_mode_blocks.py
```diff
@@ -0,0 +1,185 @@
...
+                "ConfigureForModeBlock: No _resolved_mode in state. "
+                "This should be set by MultiModePipeline._execute()"
+            )
+            mode = "text"
+            state.set("_resolved_mode", mode)
+
+        return components, state
+
+
+class LoadComponentsBlock(ModularPipelineBlocks):
```
yondonfu: Given our current simplification of VAEs eg no VAE strategies I think this can be removed. In the PR adding LightVAE we would just need a load param indicating the VAE filename to use.

src/scope/core/pipelines/schema.py
```diff
@@ -0,0 +1,268 @@
...
+        "hasInputModeControl": has_input_mode_control,
+        "hasNoiseControls": has_noise_controls,
+        "showNoiseControlsInText": show_noise_controls_in_text,
+        "showNoiseControlsInVideo": show_noise_controls_in_video,
+        "hasCacheManagement": has_cache_management,
+        "requiresVideoInVideoMode": requires_video_in_video_mode,
+    }
+
+
+def create_pipeline_schema(
```
yondonfu: A few downsides I see with this current approach of dynamically constructing a dict:

1. There is no way to look at a single place to clearly list + see what is supported
2. This forgoes the ability to get validation for free using Pydantic

Another way that could streamline things:

Each pipeline has a:

```
@property
def get_model_config(self) -> BasePipelineConfig
    return ...
```

Then, each pipeline has it's own config Pydantic model:

```
class LongLiveModelConfig(BasePipelineConfig):
    ...
```

And `get_model_config()` just returns the pipeline specific Pydantic model.

This can then be used to validate like this:

```
# JSON string input
json_str = '{"width": 1920, "height": 1080}'
cfg = LongLiveModelConfig.model_validate_json(json_str)
```

And can also be converted to JSON schema (easier to go in this direction then the other way around):

```
pipeline.get_model_config().model_json_schema()
```

There is also good support for converting from Pydantic to OpenAPI schema so that would also avoid needing to add all that custom create_parameter() boilerplate code.

src/scope/core/pipelines/interface.py
ryanontheinside: native_mode is intended to drive UI defaults - when a pipeline supports multiple modes, the UI needs to know which mode to show by default and which defaults to apply on initial load. Without it, we'd rely on implicit list ordering in supported_modes or push that decision to the client, but part of my effort here was single source of truth

src/scope/core/pipelines/schema.py
ryanontheinside: my original approach here was with Pydantic [177](https://github.com/daydreamlive/scope/pull/177#discussion_r2558961790) - so im happy to switch back just want to lock it in to avoid churn

src/scope/core/pipelines/interface.py
ryanontheinside: agreed - will remove and derive it from mode_configs.keys() where needed

src/scope/core/pipelines/schema.py
ryanontheinside: front end uses them, but they could be derived. what's your pref?

src/scope/core/pipelines/interface.py
yondonfu: I think the UI can just default to the same mode per pipeline eg Text which would remove the need for having to define native_mode.

src/scope/core/pipelines/schema.py
yondonfu: I'd prefer to just derive and avoid introducing all of this right now if possible.

The rationale here is that I think the pipeline should just focus on signaling what parameters and modes it offers and details about what controls should be showed should stay in the UI as opposed to leaking into the pipeline itself.

src/scope/core/pipelines/schema.py
yondonfu: My pref is for using Pydantic classes as described in the previous comment!

src/scope/core/pipelines/__init__.py
leszko: ok, I think I understand now. It's ok as it is. Thanks for the explanation.

src/scope/core/pipelines/interface.py
leszko: Yondon, so you mean that each pipeline would start from text-to-video? Don't we want to default to the mode that is announced together with the pipeline by their creators (meaning T2V for LongLive/Krea and V2V for SDv2)?

yondonfu: @leszko I don't see a strong reason for enshrining the mode that the pipeline was created with in the metadata if the pipeline can handle both.

src/scope/core/pipelines/__init__.py
```diff
@@ -19,12 +20,45 @@ def __getattr__(name):
...
     raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
 
 
 __all__ = [
+    # Pipeline classes
     "LongLivePipeline",
     "KreaRealtimeVideoPipeline",
     "StreamDiffusionV2Pipeline",
     "PassthroughPipeline",
+    "MultiModePipeline",
```
yondonfu: Does this actually need to be exported from the package? This seems like it should be a part of the implementation detail of the actual pipelines and users of the package wouldn't need to know about it.

src/scope/core/pipelines/interface.py
```diff
@@ -13,7 +17,65 @@ class Requirements(BaseModel):
...
+            @classmethod
+            def get_config_class(cls) -> type[BasePipelineConfig]:
+                return LongLiveConfig
+        """
+        from .schema import BasePipelineConfig
+
+        return BasePipelineConfig
+
+    @classmethod
+    def get_schema(cls) -> dict[str, Any]:
```
yondonfu: The caller can just get this information from the value returned by `get_config_class()` so seems that we can simplify and not introduce an additional required method here.

src/scope/core/pipelines/registry.py
```diff
@@ -0,0 +1,103 @@
...
+        Returns:
+            Pydantic config class if found, None otherwise
+        """
+        pipeline_class = cls.get(pipeline_id)
+        if pipeline_class is None:
+            return None
+        return pipeline_class.get_config_class()
+
+    @classmethod
+    def get_schema(cls, pipeline_id: str) -> dict[str, Any] | None:
```
yondonfu: If a caller can get the schema from the return value of get_config_class() we can omit this.

PR #205, opened 2025-11-28
03 feat: migrate pipelines to MultiModePipeline architecture

ryanontheinside:
Convert all pipelines to declarative `MultiModePipeline` with schema support.

- Enables v2v and t2v for all pipelines
- Migrate `LongLivePipeline`, `StreamDiffusionV2Pipeline`, `KreaRealtimeVideoPipeline`, `PassthroughPipeline`
- Remove old VAE wrappers (now using shared `base/vae`)
- Update server to use pipeline schemas for mode routing
- Update `prepare_video_latents` for new VAE format

Testing: 
- ran all three pipelines in their native mode
- created and executed test scripts for each pipelines non-native mode. 

Next steps:

- Front end support for multi mode
- Light vae
- Possible test script PR


src/scope/server/app.py
```diff
@@ -361,7 +408,13 @@ async def list_lora_files():
     def process_lora_file(file_path: Path, lora_dir: Path) -> LoRAFileInfo:
         """Extract LoRA file metadata."""
         size_mb = file_path.stat().st_size / (1024 * 1024)
-        relative_path = file_path.relative_to(lora_dir)
+        # Calculate relative path for folder structure
```
yondonfu: Can this ever actually happen or is this AI being overly defensive? If it's the latter can we omit this change as it feels noisy amidst everything else. Same point below.

src/scope/core/pipelines/longlive/test.py
```diff
@@ -21,60 +28,67 @@
...
 print(
-    f"Latency - Avg: {sum(latency_measures) / len(latency_measures):.2f}s, Max: {max(latency_measures):.2f}s, Min: {min(latency_measures):.2f}s"
+    f"test.py: Generated {num_frames_video} frames in {latency_video:.2f}s ({fps_video:.2f} fps)"
 )
+print(f"test.py: Output shape: {output_video.shape}")
+
+# Save video mode output
+output_video_np = output_video.detach().cpu().contiguous().numpy()
+export_to_video(output_video_np, Path(__file__).parent / "output_video.mp4", fps=16)
+print("test.py: Saved video-to-video output to output_video.mp4")
```
yondonfu: Need to add both output_video.mp4 and output_text.mp4 to .gitignore.

src/scope/core/pipelines/longlive/test.py
```diff
@@ -1,3 +1,8 @@
+"""Quick test for LongLivePipeline with new MultiModePipeline architecture.
+
```
yondonfu: For clarity can we keep the existing test script and reduce the prompts as is needed to make it run faster and then add a separate test script for video-to-video?

Eg

`test_text_to_video.py`
`test_video_to_video.py`

And for both retain the same format for printing perf stats which includes min/max/avg latency + fps.

yondonfu: Actually, let's just omit this change altogether from this specific PR because I'd prefer to do a separate pass to update all the test scripts if we are going to make modifications to keep them consistent and to also update docs that contain references to the current naming structure eg `uv run -m scope.core.pipelines.longlive.test`.

src/scope/core/pipelines/wan2_1/blocks/prepare_latents.py
```diff
@@ -93,11 +93,18 @@ def __call__(self, components, state: PipelineState) -> tuple[Any, PipelineState
         block_seed = base_seed + block_state.current_start_frame
         rng = torch.Generator(device=generator_param.device).manual_seed(block_seed)
 
+        # Determine number of latent frames to generate
+        num_latent_frames = components.config.num_frame_per_block
+        # VAE stream_decode requires at least 2 latent frames on first batch
+        # because it splits into x[:,:,:1,:,:] and x[:,:,1:,:,:] (second part must be non-empty)
+        if block_state.current_start_frame == 0 and num_latent_frames < 2:
```
yondonfu, line 100: Is this required for streamdiffusionv2 since it uses num_frame_per_block = 1?

src/scope/core/pipelines/streamdiffusionv2/pipeline.py
```diff
@@ -24,8 +30,8 @@
 
 DEFAULT_DENOISING_STEP_LIST = [750, 250]
 
-# Chunk size for streamdiffusionv2
-CHUNK_SIZE = 4
+# Default chunk size for video mode
+DEFAULT_VIDEO_CHUNK_SIZE = 4
```
yondonfu: I think we can remove this as well as the `video_input_size` arg in `prepare_for_mode()` if we already have logic that calculates the right value for us.

src/scope/core/pipelines/wan2_1/blocks/prepare_latents.py
ryanontheinside: Yes, for t2v mode

src/scope/core/pipelines/streamdiffusionv2/pipeline.py
ryanontheinside: removed

ryanontheinside:
> This is looking good and just had a few small comments.
> 
> Also, how are you thinking about using the MultiModePipeline? I see that you removed its usage in the latest version of this branch. Do you still intend to use it just in a separate PR?

Upon rewriting this branch on the simplified infrastructure, doing this lifting in helper methods seemed simpler. Rafal had suggested this at one point anyways, so I went with it. I think the file was copied by Fork and I didn't notice - I've now deleted it.

PR #206, opened 2025-11-28
04 feat: add frontend UI for input mode switching

ryanontheinside:
Depends upon #205 

Add UI for text/video mode selection driven by pipeline schemas.

- New `constants/modes.ts` and `lib/pipelineModes.ts`
- Update components to handle mode-specific settings
- Remove deprecated `useInputMode` hook
- Mode selection integrates with backend schema API


frontend/src/data/pipelines.ts
```diff
@@ -1,69 +1,58 @@
...
-    name: "Krea Realtime Video",
     docsUrl:
       "https://github.com/daydreamlive/scope/blob/main/pipelines/krea_realtime_video/docs/usage.md",
     about:
       "A streaming pipeline and autoregressive video diffusion model from Krea. The model is trained using Self-Forcing on Wan2.1 14b.",
     modified: true,
-    category: "no-video-input",
-    defaultPrompt:
-      "A 3D animated scene. A **panda** walks along a path towards the camera in a park on a spring day.",
+    category: "video-input",
```
yondonfu, line 50: We should be able to remove this field since the pipeline metadata will include modes right?

PR #212, opened 2025-12-02
Add Electron Desktop App

leszko:
# Electron App Release

This PR adds the Electron desktop application and related GitHub Actions workflow for building and distributing Scope as a native desktop app for Windows.

## Build & Release Pipeline

**Automated Releases:**
- Pushing a version tag (e.g., `v0.1.0-alpha.8`) triggers the Electron build workflow
- Built artifacts are automatically attached to the GitHub release
- Currently building: Windows x64 only

**User Experience:**
1. User downloads the installer from the GitHub release page (or `daydream.live`)
2. User installs and launches the app
3. On subsequent launches, the app automatically checks for updates
4. When an update is available, user is prompted to download and install

## Technical Architecture

**Startup Flow:**
- Electron app bundles the entire Python project in its resources during build
- In the Windows Installer (and on first launch as plan B), the app copies Python files to the user's data directory and installs `uv` (Python package manager)
- Main process spawns the Python server via `uv run daydream-scope` with automatic port allocation (finds next available if busy)
- Renderer shows loading screens (Setup → ServerLoading) while dependencies install and server starts
- Once server health check passes, the app loads the frontend from `http://localhost:PORT`
- Frontend runs as a web app inside Electron's BrowserView, communicating with the Python backend via HTTP

## Important Notes

**Versioning:**
- Electron app uses separate versioning from Python packages
- Mapping example: Python `0.1.0a7` → Electron `0.1.0-alpha.7`
- Version is managed in `app/package.json`

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,7 +1,7 @@
 {
   "name": "daydream-scope-desktop",
   "productName": "Daydream Scope",
-  "version": "0.1.0-alpha.7",
+  "version": "0.1.0-rc1",
   "description": "A tool for running and customizing real-time, interactive generative AI pipelines and models",
   "main": ".vite/build/main.js",
   "scripts": {
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "daydream-scope-desktop",
-  "version": "0.1.0-alpha.7",
+  "version": "0.1.0-rc1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "daydream-scope-desktop",
-      "version": "0.1.0-alpha.7",
+      "version": "0.1.0-rc1",
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
@@ -88,7 +88,6 @@
       "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.5",
@@ -458,7 +457,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18"
       },
@@ -482,7 +480,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=18"
       }
@@ -1153,18 +1150,6 @@
         "node": ">=6.0.0"
       }
     },
-    "node_modules/@jridgewell/source-map": {
-      "version": "0.3.11",
-      "resolved": "https://registry.npmjs.org/@jridgewell/source-map/-/source-map-0.3.11.tgz",
-      "integrity": "sha512-ZMp1V8ZFcPG5dIWnQLr3NSI1MiCU7UETdS/A0G8V/XWHvJv3ZsFqutJn1Y5RPmAPX6F3BiE397OqveU/9NCuIA==",
-      "dev": true,
-      "license": "MIT",
-      "optional": true,
-      "dependencies": {
-        "@jridgewell/gen-mapping": "^0.3.5",
-        "@jridgewell/trace-mapping": "^0.3.25"
-      }
-    },
     "node_modules/@jridgewell/sourcemap-codec": {
       "version": "1.5.5",
       "resolved": "https://registry.npmjs.org/@jridgewell/sourcemap-codec/-/sourcemap-codec-1.5.5.tgz",
@@ -1690,7 +1675,8 @@
       "resolved": "https://registry.npmjs.org/@types/aria-query/-/aria-query-5.0.4.tgz",
       "integrity": "sha512-rfT93uj5s0PRL7EzccGMs3brplhcrghnDoV26NqKhCAS1hVo+WdNsPvE/yb6ilfr5hi2MEk6d5EWJTKdxg8jVw==",
       "dev": true,
-      "license": "MIT"
+      "license": "MIT",
+      "peer": true
     },
     "node_modules/@types/babel__core": {
       "version": "7.20.5",
@@ -1829,7 +1815,6 @@
       "integrity": "sha512-MWtvHrGZLFttgeEj28VXHxpmwYbor/ATPYbBfSFZEIRK0ecCFLl2Qo55z52Hss+UV9CRN7trSeq1zbgx7YDWWg==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "csstype": "^3.2.2"
       }
@@ -1840,7 +1825,6 @@
       "integrity": "sha512-jp2L/eY6fn+KgVVQAOqYItbF0VY/YApe5Mz2F0aykSO8gx31bYCZyvSeYxCHKvzHG5eZjc+zyaS5BrBWya2+kQ==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "peerDependencies": {
         "@types/react": "^19.2.0"
       }
@@ -1972,7 +1956,6 @@
       "integrity": "sha512-izzd2zmnk8Nl5ECYkW27328RbQ1nKvkm6Bb5DAaz1Gk59EbLkiCMa6OLT0NoaAYTjOFS6N+SMYW1nh4/9ljPiw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@vitest/utils": "2.1.9",
         "fflate": "^0.8.2",
@@ -2074,7 +2057,6 @@
       "integrity": "sha512-j3fVLgvTo527anyYyJOGTYJbG+vnnQYvE0m5mmkc1TK+nxAppkCLMIL0aZ4dblVCNoGShhm+kzE4ZUykBoMg4g==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "fast-deep-equal": "^3.1.1",
         "fast-json-stable-stringify": "^2.0.0",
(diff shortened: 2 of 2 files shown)


.github/workflows/build-electron.yml
```diff
@@ -0,0 +1,106 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # Required for electron-builder to generate proper version info
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '23'
```
yondonfu: If we can, let's pin the same Node version everywhere in CI since we already specify 22.19.0 [here](https://github.com/daydreamlive/scope/blob/3f8e3df59abe4672c9b6d012fa57664cdbf3f8fa/.github/workflows/test.yml#L21)?

app/package.json
```diff
@@ -0,0 +1,66 @@
+{
+  "name": "daydream-scope-desktop",
+  "productName": "Daydream Scope",
+  "version": "0.1.0-alpha.7",
+  "description": "Daydream Scope - Real-time AI video generation and streaming",
```
yondonfu:
```suggestion
  "description": "A tool for running and customizing real-time, interactive generative AI pipelines and models",
```

to match description used in pyproject.toml.

app/package.json
```diff
@@ -0,0 +1,66 @@
+{
+  "name": "daydream-scope-desktop",
+  "productName": "Daydream Scope",
+  "version": "0.1.0-alpha.7",
```
yondonfu: I noticed that the PR description noted that the Electron app is versioned separately from the Python package.

Could you elaborate on the rationale for this? I would've imagined that it'd be simpler to use the same version for everything since we already do so for the frontend and the Python project. But, curious if you see a benefit in specifically versioning the Electron app differently.

app/electron-builder.yml
```diff
@@ -0,0 +1,158 @@
...
+
+# Windows configuration - NSIS installer (best practice)
+win:
+  icon: assets/icon.ico
+  target:
+    - target: nsis
+      arch:
+        - x64
+  publisherName: Livepeer Inc.
+  verifyUpdateCodeSignature: false # Disable code signature verification
```
yondonfu: Shouldn't this be true? IIUC this enables verifying publisher signature during updates?

app/electron-builder.yml
```diff
@@ -0,0 +1,158 @@
...
+  sign: false # Skip code signing (disable to avoid privilege errors on Windows)
+  signDlls: false # Skip DLL signing
+  signAndEditExecutable: false # Prevent signing and editing executables
+
+nsis:
+  oneClick: false
+  perMachine: false # Per-user installation (best practice - no admin required)
+  allowToChangeInstallationDirectory: true
+  createDesktopShortcut: always
```
yondonfu: Let's make this optional.

app/electron-builder.yml
```diff
@@ -0,0 +1,158 @@
...
+linux:
+  icon: assets/icon.png
+  category: Utility
+  target:
+    - target: deb
+      arch:
+        - x64
+  executableName: daydream-scope
+  desktop: build/linux/daydream-scope.desktop
+  synopsis: Real-time AI video generation and streaming
```
yondonfu: Let's just use the same description as Python project so that it is consistent everywhere and we can find + replace if we ever need to update.

app/electron-builder.yml
```diff
@@ -0,0 +1,158 @@
...
+  description: >
```
yondonfu, line 120: Same comment as above.

app/src/services/setup.ts
```diff
@@ -0,0 +1,380 @@
...
+import https from 'https';
+import { spawn, execSync } from 'child_process';
+import { SetupService } from '../types/services';
+import { getPaths, UV_DOWNLOAD_URLS, getEnhancedPath } from '../utils/config';
+import { logger } from '../utils/logger';
+
+export class ScopeSetupService implements SetupService {
+  isSetupNeeded(): boolean {
+    const paths = getPaths();
+    const uvExists = fs.existsSync(paths.uvBin);
```
yondonfu, line 13: Just want to confirm that if the user already has uv installed it will just re-use that installation right?

app/src/services/autoUpdater.ts
yondonfu, line 1: Is this file needed if whitespace only changes?

app/src/components/styles.ts
yondonfu, line 1: Is this file needed if whitespace only changes?

app/build/linux/daydream-scope.desktop
```diff
@@ -0,0 +1,17 @@
+[Desktop Entry]
+Name=Daydream Scope
+Comment=Real-time AI video generation and streaming
```
yondonfu, line 3: Same comment as elsewhere about using same description as Python project.

app/src/services/electronApp.ts
```diff
@@ -0,0 +1,506 @@
...
+    if (!app.isPackaged) return;
+
+    window.webContents.closeDevTools();
+    window.webContents.on('devtools-opened', () => {
+      logger.warn('DevTools attempted to open in production - closing');
+      window.webContents.closeDevTools();
+    });
+  }
+
+  /**
```
yondonfu, line 94: https://github.com/user-attachments/assets/439c44ba-fc20-4910-975f-1164b1014cc0

There is a bug where the entire UI shifts up/down when I click the Mode or Pipeline ID dropdown - see attached video. I would expect clicking the Mode or Pipeline ID dropdown to not shift the UI at all like this.

app/src/components/LogViewer.js
```diff
@@ -0,0 +1,65 @@
+// LogViewer script - externalized for CSP compliance
+(function() {
+  'use strict';
+
+  const params = new URLSearchParams(window.location.search);
+  const logPath = decodeURIComponent(params.get('path') || 'Unknown');
+  const logContent = decodeURIComponent(params.get('content') || '');
+
+  // Display log path
+  const logPathElement = document.getElementById('logPath');
```
yondonfu, line 10: I noticed that this log viewer is reading from C:\Users\<username>\AppData\Roaming\Daydream Scope\logs\main\...

Could you briefly explain why we need to copy server log files from the default location eg ~/.daydream-scope/logs into this directory? Not against it just want to understand what is going on.

app/src/services/setup.ts
leszko: No, it will always download the `uv` tool into the app directory and use that version. But I think it's better, because it's more controllable, we know the uv version used.

We could add a check if `uv` is available at the system path, but then we risk some issues with uv versions. And `uv` tool is super small comparing to the other dependencies we download.

app/package.json
leszko: The problem is that `0.1.0a7` is not a valid sem version format. Sem versioning has a specific format `MAJOR.MINOR.PATCH[-PRERELEASE][+BUILD]`. 

Technically we could use a non-sem-version, but it would make our lives more difficult, because we'll not stick to the common conventions and, e.g., electron-update which handles auto-updating would need some special treatment.

app/src/services/electronApp.ts
leszko: Fixed

leszko:
> * If I exit the window, the system tray icon disappears and the server is stopped. This is fine for now, but separately from this PR I'd like to consider keeping the server running in the background (for API access) even if the window is closed unless the user clicks "Quit" in the system tray menu.

Technically, it's not difficult to change. The reason why I implemented it that way (closing the app closes the server as well) is that I thought it is more intuitive. Especially these 2 scenarios:
- You close the app, you think everything is closed. You start some Jupyter notebook or some game and it crashes because t he VRAM is occupied by Scope
- Scope crashes (e.g. OOM), so you want to close it and open again; it does not help, because you closed only the frontend, not the server.

Saying that, it's all about UX, so we can definitely discuss it and change. My opinion is that it will be worth changing when we have good scenarios for backend-only apps. Spout integration is good, but AGAIU right now you anyway need to have the frontend open, so that it works.

> * I think it'll be useful to have the server URL accessible somewhere in the UI especially since the port across runs if there is an initial conflict. At the moment, I think it is only visible if you toggle the Logs window. This is fine for now, but would like to consider an improvement here separately.

Yes, we can add it later. Not difficult, it's just about the UX where to show it.

app/src/components/LogViewer.js
leszko: We don't need to. Three (not so strong) reasons we may want to copy the logs:
1. Then, all your Electron App logs (Scope Serve, but also: updater, installer) are in one place: `C:\Users\user\AppData\Roaming\Daydream Scope\logs` (otherwise we have Electron App logs in different directories)
2. If you run both Electron App and the github repo mode on the same machine, then `C:\Users\user\AppData\Roaming\Daydream Scope\logs` includes only logs from the Electron App
3. Related to Point 2. The logic in Log Viewer is simpler, because it shows only the latest file with the logs from the given directory. Otherwise, it would need to know somehow which log files are related to Electron App and which are not.

None of these is a very "strong" reason, so I have a preference to not copy the logs, then we can change it.

app/src/services/setup.ts
yondonfu: SGTM thanks for explanation.

app/src/components/LogViewer.js
yondonfu: Hmm seems simpler to have a single source of truth re: logs and then have the LogViewer open the latest log file from wherever the server keeps logs. This assumes that if you install from source and use the Electron app that by default they share the same ~/.daydream-scope dir which feels fine because you can always configure this to be elsewhere too.

So, I'd say just to simplify to read from ~/.daydream-scope. WDYT?

app/package.json
yondonfu: Discussed offline to switch everything to proper sem ver.

app/src/components/LogViewer.js
yondonfu: This is going to be done separately from this PR.

yondonfu (approved):
LGTM

PR #213, opened 2025-12-03
Spout Send & Receive

BuffMcBigHuge:
# Spout Send & Receive

This PR adds bidirectional Spout support to Scope, enabling real-time video texture sharing with Spout-compatible applications like TouchDesigner, OBS, and other Windows-based media tools.

## Overview

Spout is a Windows-only protocol for sharing OpenGL textures between applications with minimal latency. This implementation adds both **Spout input** (receiving frames from external apps) and **Spout output** (sending processed frames to external apps).

### Spout Input (Receiver)
- Receive video frames from Spout senders
- Configurable sender name matching (empty string connects to any active sender)
- Automatic buffer resizing when sender resolution changes
- Frame rate limiting (30 FPS target) to match pipeline processing rate (Needs review)
- Integrated as a new input mode alongside "Video" and "Camera"

### Spout Output (Sender)
- Send processed frames to Spout receivers
- Customizable sender name (default: "Scope")
- Independent of WebRTC output queue
- Automatic format conversion (RGB/RGBA, float/uint8)

## Technical Details

- **Platform**: Windows only (SpoutGL requires Windows)
- **Dependencies**: `SpoutGL` and `pyopengl` (optional, gracefully handled if missing)
- **Frame Format**: RGB/RGBA uint8 arrays `[0, 255]` or float `[0, 1]`
- **Threading**: Spout input runs in a dedicated daemon thread
- **Error Handling**: Comprehensive error handling with fallback behavior on non-Windows platforms

### Receiving from Spout (Input)
1. Select "Spout" as the input mode
2. Optionally specify a sender name (leave empty to connect to any active sender)
3. Start streaming - frames will be received from the Spout sender

### Sending to Spout (Output)
1. Open Settings panel
2. Enable "Spout Output" toggle
3. Optionally customize the sender name (default: "Scope")
4. Processed frames will be available to Spout receivers

https://github.com/user-attachments/assets/41290acd-9809-4c34-b401-3ae6738e5520

### DIFF
--- a/frontend/src/components/InputAndControlsPanel.tsx
+++ b/frontend/src/components/InputAndControlsPanel.tsx
@@ -11,6 +11,7 @@ import { ToggleGroup, ToggleGroupItem } from "./ui/toggle-group";
 import { Badge } from "./ui/badge";
 import { Input } from "./ui/input";
 import { Upload, Radio } from "lucide-react";
+import { LabelWithTooltip } from "./ui/label-with-tooltip";
 import type { VideoSourceMode } from "../hooks/useVideoSource";
 import type { PromptItem, PromptTransition } from "../lib/api";
 import type { InputMode } from "../types";
@@ -210,9 +211,11 @@ export function InputAndControlsPanel({
                   <span className="text-sm font-medium">Spout Receiver</span>
                 </div>
                 <div className="space-y-2">
-                  <label className="text-xs text-muted-foreground">
-                    Sender Name (leave empty for any)
-                  </label>
+                  <LabelWithTooltip
+                    label="Sender Name"
+                    tooltip="Receive video from Spout-compatible apps like TouchDesigner, Resolume, OBS. Leave empty to receive from any sender."
+                    className="text-xs text-muted-foreground"
+                  />
                   <Input
                     type="text"
                     value={spoutReceiverName}
@@ -222,10 +225,6 @@ export function InputAndControlsPanel({
                     placeholder="TDSyphonSpoutOut"
                   />
                 </div>
-                <p className="text-xs text-muted-foreground">
-                  Receive video from Spout-compatible apps like TouchDesigner,
-                  Resolume, OBS.
-                </p>
               </div>
             ) : (
               /* Video/Camera Input Preview */
--- a/frontend/src/components/SettingsPanel.tsx
+++ b/frontend/src/components/SettingsPanel.tsx
@@ -651,12 +651,12 @@ export function SettingsPanel({
             <div className="space-y-3">
               <div className="flex items-center gap-2 text-muted-foreground">
                 <Radio className="h-4 w-4" />
-                <span className="text-sm font-medium">Spout Sender</span>
+                <LabelWithTooltip
+                  label="Spout Sender"
+                  tooltip="Send processed frames to external apps like TouchDesigner, Resolume, OBS."
+                  className="text-sm font-medium"
+                />
               </div>
-              <p className="text-xs text-muted-foreground">
-                Send processed frames to external apps like TouchDesigner,
-                Resolume, OBS.
-              </p>
               <div className="space-y-2 p-2 border rounded-md">
                 <div className="flex items-center justify-between gap-2">
                   <LabelWithTooltip


src/scope/spout/sender.py
```diff
@@ -0,0 +1,242 @@
...
+        Returns:
+            True if send was successful
+        """
+        if self.sender is None:
+            return False
+
+        try:
+            # Convert torch tensor to numpy if needed
+            if isinstance(frame, torch.Tensor):
+                frame = frame.detach().cpu().numpy()
```
yondonfu: My best guess on what is causing the perf hit/slowdown is that it is this movement of each frame to host CPU mem. SpoutGL might ultimately convert the numpy array into a texture in GPU mem that is shared to the receiver (I haven't looked into what it does under the hood yet), but if we are doing a CPU mem transfer here I suspect that would cause non-negligible overhead as that is typically slow especially in the scenario where we have many frames per second.

Ideally, we'd be able to keep the tensor (frame) in GPU mem (CUDA) the entire time. With SpoutGL, perhaps what we need is to convert that frame into a OpenGL texture while keeping it in GPU mem the entire time. I searched around and while it doesn't seem trivial it seems doable. A couple links I encountered that seem relevant:

https://gist.github.com/victor-shepardson/5b3d3087dc2b4817b9bffdb8e87a57c4

https://dev-discuss.pytorch.org/t/opengl-interoperability/2696

https://documen.tician.de/pycuda/gl.html

But...

Should first confirm that this is actually the cause by profiling how much time is getting eaten up by the steps taken her eg transfer to CPU mem + the send call that takes the numpy array.

After a second thought, it does seem a bit weird that the memory transfer on its own cause such a noticeable perf hit (as shown in the video attached in my other comment) because the WebRTC flow also involves a GPU -> CPU mem transfer in order for the frames to be encoded into a video stream. And the WebRTC video is much smoother than the Spout output in TD right now. So, the memory transfer might still be slowing things down, but there might be a bigger issue somewhere else that is the primary cause of the Spout output in TD being more choppy.

BuffMcBigHuge: Great find - slipped over this!

src/scope/server/frame_processor.py
```diff
@@ -329,6 +562,22 @@ def process_chunk(self):
                         break
 
             for frame in output:
+                # Send to Spout first (independent of WebRTC queue)
+                if self.spout_output_enabled and self.spout_sender is not None:
```
yondonfu: Ah so the actual cause of the jittery playback seems to be here. IIUC based on [this](https://spout.discourse.group/t/setframesync-and-waitframesync/531) and [this](https://spout.discourse.group/t/full-screen-compositor/475) the Spout receiver will:

- By default duplicate textures if the sender is behind and drop textures if the sender is ahead because they both rely on shared texture handle that is constantly being updated
- The receiver will run at the rate that the application chooses to run it

TD's default [timeline](https://derivative.ca/UserGuide/Timeline) is 60. If that means the receiver runs at 60fps then what might've been happening might've been:

- A pipeline generates in "bursts" eg 12 frames every say 1s and then immediately sends each of the frames via Spout
- The receiver gets all of these and renders them
- While the pipeline is generating the next batch, the receiver runs again many times and each time it just duplicates the current frame for the shared texture handle. This results in a freeze.
- The video unfreezes when the receivers gets the next batch

If the receiver instead gets a steady rate of frames then even if the sending FPS is lower than 60 the receiver will duplicate frames at a constant rate as opposed to duplicating every 1s which results in that freezing/jitter effect.

This [commit](https://github.com/daydreamlive/scope/commit/681f204b300077ed02840fc3fd91d67f3e8250c7) (vibe coded) ensures a smooth, constant frame rate going into Spout by sending into Spout at the same rate that frames are being pulled for the WebRTC stream. The latter already contains a bunch of logic for maintaining a smooth, constant rate so the commit just tries to piggyback off of that. I don't think this is the cleanest design, but just sharing for now to illustrate that smooth, constant frame rate for the Spout sender is how to solve for jittery playback.

src/scope/spout/sender.py
yondonfu: Ah so there mem transfer is still incurring overhead, but I believe the real cause of the perf hit is described in [this comment](https://github.com/daydreamlive/scope/pull/213/files#r2590954862).

src/scope/server/spout/sender.py
```diff
@@ -0,0 +1,308 @@
...
+# Check if SpoutGL is available
+try:
+    import SpoutGL
+
+    SPOUT_AVAILABLE = True
+    # Log available methods for debugging
+    logger.debug(f"SpoutGL sender methods: {dir(SpoutGL.SpoutSender())}")
+except ImportError:
+    SPOUT_AVAILABLE = False
+    logger.warning("SpoutGL not available. Install with: pip install SpoutGL pyopengl")
```
yondonfu: Let's add the `SpoutGL` and `pyopengl` deps (conditional on the platform being Windows) to `pyproject.toml` so they can be installed via `uv`.

src/scope/server/spout/__init__.py
```diff
@@ -0,0 +1,15 @@
+"""
```
yondonfu, line 1: Let's move this module into `server` since at the moment it is logically tied with the server?

Eg.

```
/server
    /spout
```

frontend/src/components/InputAndControlsPanel.tsx
```diff
@@ -188,60 +197,88 @@ export function InputAndControlsPanel({
...
-                <>
-                  <input
-                    type="file"
-                    accept="video/*"
-                    onChange={handleFileUpload}
-                    className="hidden"
-                    id="video-upload"
-                    disabled={isConnecting}
+                <p className="text-xs text-muted-foreground">
+                  Receive video from Spout-compatible apps like TouchDesigner,
```
yondonfu: Let's move this into a tooltip when hovering over "Sender Name" - see other tooltips in the frontend for the convention used elsewhere.

frontend/src/components/InputAndControlsPanel.tsx
```diff
@@ -180,6 +186,9 @@ export function InputAndControlsPanel({
               <ToggleGroupItem value="camera" aria-label="Camera">
                 Camera
               </ToggleGroupItem>
+              <ToggleGroupItem value="spout" aria-label="Spout">
```
yondonfu: Let's only show this source type on Windows. 

There is already a hardware info check elsewhere that might be useful here if it can propagate OS info too. Might make sense to rely on this as source of truth as ultimately what matters is what OS the server is running.

frontend/src/components/SettingsPanel.tsx
```diff
@@ -635,6 +641,62 @@ export function SettingsPanel({
             </div>
           </div>
         )}
+
+        {/* Spout Output Settings (Windows only) */}
```
yondonfu: Let's only show this on Windows - see my other comment about only showing Spout as an input on Windows as well.

frontend/src/components/SettingsPanel.tsx
```diff
@@ -635,6 +641,62 @@ export function SettingsPanel({
...
+        <div className="space-y-4">
+          <div className="space-y-3">
+            <h3 className="text-sm font-medium">Spout Output (Windows)</h3>
+            <p className="text-xs text-muted-foreground">
+              Send processed frames to external apps like TouchDesigner,
```
yondonfu: Let's move this into a tooltip when hovering over Spout Sender.

frontend/src/components/SettingsPanel.tsx
```diff
@@ -635,6 +641,62 @@ export function SettingsPanel({
             </div>
           </div>
         )}
+
+        {/* Spout Output Settings (Windows only) */}
+        <div className="space-y-4">
+          <div className="space-y-3">
+            <h3 className="text-sm font-medium">Spout Output (Windows)</h3>
```
yondonfu:
```suggestion
            <h3 className="text-sm font-medium">Spout Sender</h3>
```

as that seems to be the naming convention in other apps eg TD, Unity, etc.

frontend/src/components/SettingsPanel.tsx
```diff
@@ -635,6 +641,62 @@ export function SettingsPanel({
...
+                  }}
+                  variant="outline"
+                  size="sm"
+                  className="h-7"
+                >
+                  {spoutOutput?.enabled ? "ON" : "OFF"}
+                </Toggle>
+              </div>
+
+              {spoutOutput?.enabled && (
```
yondonfu: Image

Image

Could we use this layout? IMO this aligns better with the existing layout of the other controls in the Settings panel

frontend/src/components/InputAndControlsPanel.tsx
yondonfu:
```suggestion
              <ToggleGroupItem value="spout" aria-label="Spout Receiver">
```

if we just call the source type this then we could reduce the amount of visual space we need to take up below.

frontend/src/components/InputAndControlsPanel.tsx
```diff
@@ -188,60 +197,88 @@ export function InputAndControlsPanel({
...
-                      ? "Camera access failed:"
-                      : "Video error:"}
-                  </p>
-                  <p className="text-xs mt-1">{error}</p>
+            {mode === "spout" ? (
+              /* Spout Input Configuration */
+              <div className="space-y-3 p-3 rounded-lg bg-muted/10">
+                <div className="flex items-center gap-2 text-muted-foreground">
+                  <Radio className="h-4 w-4" />
+                  <span className="text-sm font-medium">Spout Receiver</span>
```
yondonfu, line 211: If we just call the mode Spout Receiver we could remove this.

frontend/src/components/InputAndControlsPanel.tsx
```diff
@@ -188,60 +197,88 @@ export function InputAndControlsPanel({
...
-                  playsInline
-                />
-              ) : (
-                <div className="text-center text-muted-foreground text-sm">
-                  {mode === "camera" ? "Camera Preview" : "Video Preview"}
+                <div className="space-y-2">
+                  <label className="text-xs text-muted-foreground">
+                    Sender Name (leave empty for any)
+                  </label>
+                  <Input
```
yondonfu, line 219: Image

Can we make the input here align with the left edge of the Prompts input?

frontend/src/components/InputAndControlsPanel.tsx
yondonfu: And the Radio could be put on the left side of Sender Name below

frontend/src/hooks/useWebRTC.ts
```diff
@@ -295,6 +295,8 @@ export function useWebRTC(options?: UseWebRTCOptions) {
       reset_cache?: boolean;
       kv_cache_attention_bias?: number;
       paused?: boolean;
+      spout_output?: { enabled: boolean; senderName: string };
```
yondonfu: spout_receiver and spout_sender?

src/scope/server/frame_processor.py
```diff
@@ -12,6 +12,15 @@
 
 logger = logging.getLogger(__name__)
 
+# Try to import Spout support (Windows only)
```
yondonfu: I think the right behavior here might be:

1. The Spout modules raise an import error when trying to import SpoutGL on a non-Windows platform.
2. Use a lazy import for Spout module so that it is only imported *if* the Spout I/O is requested. If we add Windows check for the Spout controls in the frontend then when the frontend should generally not request Spout at all.

WDYT?

src/scope/server/spout/receiver.py
```diff
@@ -0,0 +1,265 @@
...
+
+import numpy as np
+
+logger = logging.getLogger(__name__)
+
+# Check if SpoutGL is available
+try:
+    import SpoutGL
+
+    SPOUT_AVAILABLE = True
```
yondonfu: As mentioned in another comment feels like better behavior is raising import error here and having FrameProcessor take care of it as opposed to trying to silently continue with just a warning. This could result in a cleaner code eg not having to sprinkle all those checks on whether SpoutGL is available.

src/scope/server/frame_processor.py
```diff
@@ -179,6 +274,221 @@ def update_parameters(self, parameters: dict[str, Any]):
...
+                time.sleep(0.01)
+
+        logger.info(f"Spout output thread stopped after {frame_count} frames")
+
+    def _spout_input_loop(self):
+        """Background thread that receives frames from Spout and adds to buffer."""
+        logger.info("Spout input thread started")
+
+        # Target frame rate for Spout input (match typical pipeline processing rate)
+        target_fps = 30.0
```
yondonfu: Can we use actual pipeline FPS which is already tracked by FrameProcessor?

pyproject.toml
```diff
@@ -53,6 +53,8 @@ dependencies = [
     "kernels>=0.10.4",
     "triton==3.4.0; sys_platform == 'linux'",
     "triton-windows==3.4.0.post21; sys_platform == 'win32'",
+    "SpoutGL>=0.1.1; sys_platform == 'win32'",
```
yondonfu, line 56: Looks like we need a local `uv sync` to update uv.lock?

frontend/src/components/InputAndControlsPanel.tsx
```diff
@@ -18,6 +19,12 @@ import { PromptInput } from "./PromptInput";
 import { TimelinePromptEditor } from "./TimelinePromptEditor";
 import type { TimelinePrompt } from "./PromptTimeline";
 
+// Detect if user is on Windows (Spout is Windows-only)
```
yondonfu: I don't think we want to do a frontend check like this because Spout support is a server side concern. If the server is running on Linux and the browser is on Windows (which wouldn't be uncommon if the user uses Runpod) then this check means that the frontend would show Spout even though the server does not support it.

I think this can be avoided if the frontend checks if the server is on Windows perhaps including that information in the hardware info request/response that is used elsewhere.

src/scope/server/spout/receiver.py
yondonfu: Bump on this as it looks like we're just logging a warning here and proceeding. WDYT?

src/scope/server/spout/sender.py
yondonfu: Similar comment here - bump on raising import error vs. just proceeding with a warning.

frontend/src/components/InputAndControlsPanel.tsx
yondonfu: Bump

frontend/src/components/SettingsPanel.tsx
yondonfu: Bump

frontend/src/components/SettingsPanel.tsx
yondonfu: Bump

frontend/src/components/InputAndControlsPanel.tsx
yondonfu: Bump

frontend/src/components/InputAndControlsPanel.tsx
yondonfu: Bump

src/scope/server/frame_processor.py
yondonfu: FWIW in response to an offline comment:

The main motivation here is just to avoid hardcoding 30 here because ultimately I think what we care about is making sure that the input FPS of Spout matches pipeline FPS as indicated by the comment and that happens to be what is already returned by get_current_pipeline_fps().

yondonfu (approved):
There are still some visual changes that didn't make it in, but LGTM for now and going to fixup on `main`.

PR #214, opened 2025-12-03
04 Universal Input Modes Front End

ryanontheinside:
Depends upon #205 

Front end support for universal input modes. 

Next steps:
- Additional VAE implementation with front end support

### DIFF
--- a/frontend/src/pages/StreamPage.tsx
+++ b/frontend/src/pages/StreamPage.tsx
@@ -12,7 +12,11 @@ import { useVideoSource } from "../hooks/useVideoSource";
 import { useWebRTCStats } from "../hooks/useWebRTCStats";
 import { usePipeline } from "../hooks/usePipeline";
 import { useStreamState } from "../hooks/useStreamState";
-import { PIPELINES, getPipelineDefaultMode } from "../data/pipelines";
+import {
+  PIPELINES,
+  getPipelineDefaultMode,
+  getDefaultPromptForMode,
+} from "../data/pipelines";
 import type {
   InputMode,
   PipelineId,
@@ -49,9 +53,11 @@ export function StreamPage() {
   const { settings, updateSettings, getDefaults, supportsNoiseControls } =
     useStreamState();
 
-  // Prompt state
+  // Prompt state - use unified default prompts based on mode
+  const initialMode =
+    settings.inputMode || getPipelineDefaultMode(settings.pipelineId);
   const [promptItems, setPromptItems] = useState<PromptItem[]>([
-    { text: PIPELINES[settings.pipelineId]?.defaultPrompt || "", weight: 100 },
+    { text: getDefaultPromptForMode(initialMode), weight: 100 },
   ]);
   const [interpolationMethod, setInterpolationMethod] = useState<
     "linear" | "slerp"
@@ -62,6 +68,12 @@ export function StreamPage() {
 
   // Track when we need to reinitialize video source
   const [shouldReinitializeVideo, setShouldReinitializeVideo] = useState(false);
+
+  // Store custom video resolution from user uploads - persists across mode/pipeline changes
+  const [customVideoResolution, setCustomVideoResolution] = useState<{
+    width: number;
+    height: number;
+  } | null>(null);
 
   const [isLive, setIsLive] = useState(false);
   const [isTimelineCollapsed, setIsTimelineCollapsed] = useState(false);
@@ -139,6 +151,14 @@ export function StreamPage() {
     onStopStream: stopStream,
     shouldReinitialize: shouldReinitializeVideo,
     enabled: settings.inputMode === "video",
+    // Sync output resolution when user uploads a custom video
+    // Store the custom resolution so it persists across mode/pipeline changes
+    onCustomVideoResolution: resolution => {
+      setCustomVideoResolution(resolution);
+      updateSettings({
+        resolution: { height: resolution.height, width: resolution.width },
+      });
+    },
   });
 
   const handlePromptsSubmit = (prompts: PromptItem[]) => {
@@ -169,28 +189,24 @@ export function StreamPage() {
     // Get mode-specific defaults from backend schema
     const modeDefaults = getDefaults(settings.pipelineId, newMode);
 
+    // Use custom video resolution if switching to video mode and one exists
+    // This preserves the user's uploaded video resolution across mode switches
+    const resolution =
+      newMode === "video" && customVideoResolution
+        ? customVideoResolution
+        : { height: modeDefaults.height, width: modeDefaults.width };
+
     // Update settings with new mode and ALL mode-specific defaults including resolution
     updateSettings({
       inputMode: newMode,
-      resolution: { height: modeDefaults.height, width: modeDefaults.width },
+      resolution,
       denoisingSteps: modeDefaults.denoisingSteps,
       noiseScale: modeDefaults.noiseScale,
       noiseController: modeDefaults.noiseController,
     });
 
-    // Update prompts to mode-specific defaults if available
-    if (
-      "defaultPrompts" in modeDefaults &&
-      modeDefaults.defaultPrompts &&
-      modeDefaults.defaultPrompts.length > 0
-    ) {
-      setPromptItems(
-        modeDefaults.defaultPrompts.map((text: string) => ({
-          text,
-          weight: 100,
-        }))
-      );
-    }
+    // Update prompts to mode-specific defaults (unified per mode, not per pipeline)
+    setPromptItems([{ text: getDefaultPromptForMode(newMode), weight: 100 }]);
 
     // Handle video source based on mode
     if (newMode === "video") {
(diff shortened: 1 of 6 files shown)


src/scope/server/frame_processor.py
```diff
@@ -252,8 +256,13 @@ def process_chunk(self):
 
         requirements = None
         if hasattr(pipeline, "prepare"):
+            prepare_params = dict(self.parameters.items())
```
yondonfu, line 259: *No immediate action requested here*

Noting that we should be able to remove the prepare() fn altogether separately.

src/scope/server/app.py
```diff
@@ -308,6 +309,33 @@ async def get_pipeline_status(
         raise HTTPException(status_code=500, detail=str(e)) from e
 
 
+@app.get("/api/v1/pipelines/schemas", response_model=PipelineSchemasResponse)
+async def get_pipeline_schemas():
```
yondonfu, line 313: *No immediate action requested here*

It's weird to have `/pipeline` and also `/pipelines` in the REST API. I think a more standard REST convention would look like this:

```
GET /api/v1/pipelines -> returns all pipelines with their metadata
GET /api/v1/pipelines/status -> gets status of specific pipeline
POST /api/v1/pipelines/load -> loads a specific pipeline
```

Let's consider this in a separate PR.

src/scope/core/pipelines/schema.py
```diff
@@ -103,38 +193,81 @@ def is_video_mode(self) -> bool:
...
+                    "A 3D animated scene. A **panda** walks along a path towards the camera in a park on a spring day."
+                ],
+            ),
+            "video": ModeDefaults(
+                # Video mode: requires input frames, noise controls active
+                height=512,
+                width=512,
+                noise_scale=0.7,
+                noise_controller=True,
+                input_size=4,
```
yondonfu: This doesn't seem right? If longlive uses num_frame_per_block = 3 then with VAE temporal factor = 4 that should be an input size of 12?

src/scope/core/pipelines/schema.py
```diff
@@ -154,40 +287,105 @@ class StreamDiffusionV2Config(BasePipelineConfig):
...
+                    "A 3D animated scene. A **panda** walks along a path towards the camera in a park on a spring day."
+                ],
+            ),
+            "video": ModeDefaults(
+                # Video mode: requires input frames, noise controls active
+                height=256,
+                width=256,
+                noise_scale=0.7,
+                noise_controller=True,
+                input_size=4,
```
yondonfu: This doesn't seem right? If krea uses num_frame_per_block = 3 then with VAE temporal factor = 4 that should be an input size of 12?

src/scope/core/pipelines/schema.py
```diff
@@ -154,40 +287,105 @@ class StreamDiffusionV2Config(BasePipelineConfig):
...
 
-    # Passthrough defaults - requires video input
-    height: int = Field(default=512, ge=1, description="Output height in pixels")
-    width: int = Field(default=512, ge=1, description="Output width in pixels")
+    # Mode support - video only
+    supported_modes: ClassVar[list[InputMode]] = ["video"]
+    default_mode: ClassVar[InputMode] = "video"
+
+    # Passthrough defaults - requires video input (distinct from StreamDiffusionV2)
+    height: int = Field(default=480, ge=1, description="Output height in pixels")
```
yondonfu: Can we just keep the 512x512 default? A change doesn't seem required here.

src/scope/core/pipelines/schema.py
```diff
@@ -7,18 +7,53 @@
...
+    # Core parameters
+    denoising_steps: list[int] | None = None
+
+    # Video mode parameters
+    noise_scale: float | None = None
+    noise_controller: bool | None = None
+    input_size: int | None = None
+
+    # Default prompts for this mode (list of prompt text strings, will be converted to PromptItems with weight 1.0)
+    default_prompts: list[str] | None = None
```
yondonfu: This doesn't seem like the right place to define default prompts because for V2V we're defining the default prompt based on a specific default input video that is used in the frontend. But, only the frontend knows what the default input video is. So, it seems strange to define the default prompt here since its making an assumption about what the frontend does. IMO would be better not to make assumptions about how is using the pipeline - the frontend just happens to be the primary consumer right now.

src/scope/core/pipelines/schema.py
```diff
@@ -154,40 +287,105 @@ class StreamDiffusionV2Config(BasePipelineConfig):
...
+        return {
+            "text": ModeDefaults(
+                # Text mode: distinct resolution, no video input, no noise controls
+                height=512,
+                width=512,
+                noise_scale=None,
+                noise_controller=None,
+                input_size=None,
+                denoising_steps=[1000, 750],
+                default_prompts=["A dog in the grass looking around, photorealistic"],
```
yondonfu: A request to use the same default prompts for V2V + T2V across pipelines if possible.

src/scope/core/pipelines/defaults.py
```diff
@@ -162,19 +165,15 @@ def prepare_for_mode(
...
-    if video_input_size is None:
-        video_input_size = calculate_video_input_size(components_config)
-
-    # If video is explicitly provided, use video mode
+    # Video mode is indicated by presence of 'video' in kwargs
+    # (FrameProcessor sets this based on frontend's input_mode)
     if kwargs.get("video") is not None:
+        # Calculate video input size if not provided
+        if video_input_size is None:
+            video_input_size = calculate_video_input_size(components_config)
```
yondonfu, line 173: It's a bit confusing to me to see that we're defining a default input size elsewhere but we're basically always calculating the required input size here. If we always calculate the required input size can we just have that be the source of the truth as opposed to also defining a default value elsewhere that is never going to be used? In this case the default value is the calculated value.

frontend/src/pages/StreamPage.tsx
```diff
@@ -443,21 +529,11 @@ export function StreamPage() {
...
-      updateSettings({
-        resolution: {
-          height: videoResolution.height,
-          width: videoResolution.width,
-        },
-      });
-    }
-  }, [videoResolution, isStreaming, settings.pipelineId, updateSettings]);
+
+  // Note: We intentionally do NOT auto-sync videoResolution to settings.resolution.
```
yondonfu, line 538: I don't think this behavior is right for V2V. As a user, if I upload a custom video with NxM resolution I would expect the default behavior to be for the output to be at NxM resolution. If there is going to be an issue with that resolution and the app happens to know ahead of time then would expect to be informed of that as opposed to silently defaulting to a different resolution than the custom input I provide.

yondonfu (changes requested):
Looking pretty good! Almost there...

src/scope/core/pipelines/schema.py
ryanontheinside: Youre right. Another advantage to moving it to the front end is that we can easily have default prompts per input source as well. (Panda looking around for video, Panda at a desk for camera....or whatever). Will move

frontend/src/pages/StreamPage.tsx
ryanontheinside: I changed this for a few practical reasons - Even with the default video, a resolution of 512 will explode most machines with Krea. I found myself changing the resolution from 512x512 over and over again when testing and I imagined a user having the same experience. Most input videos probably will be larger than 512x512, so this same problem will apply in more scenarios. 

In ComfyUI, the size of the input latent _does_ control the resolution in the way you are suggesting, but in ComfyUI, the user is provided tools to resize the input video on the fly. 

Just letting you know my thoughts - happy to change it

src/scope/core/pipelines/schema.py
ryanontheinside: vestigial from debugging. will fix

src/scope/core/pipelines/defaults.py
ryanontheinside: yes it is confusing and good eye - the front end value is not used and I have removed it

src/scope/core/pipelines/schema.py
ryanontheinside: youre correct here - also this value is not used so removing

frontend/src/pages/StreamPage.tsx
yondonfu: What about this?

1. Keep the default output res of Krea at a value that should generally work.
2. If the user uploads a custom input video sync the output res to that for all pipelines.

This ensures that the defaults work, but if the user uploads something custom that the output res is adjusted to reflect that.

The scenario that I would find painful UX wise is if I upload a custom video for streamdiffusionv2 or longlive that has a portrait aspect ratio let's say 832x480 then I would have to manually update the output res each time.

yondonfu (changes requested):
I'm getting a default output res of 512x512 in the UI for everything now eg all modes, all pipelines.

I think we want the defaults to be:

- 512x512 for V2V
- 320x576 for T2V
- (Exception) 256x256 for V2V with Krea

This is just to be consistent with current behavior on `main` (with the additional scenario for Krea V2V) for now.

src/scope/server/app.py
yondonfu: Tracking in https://github.com/daydreamlive/scope/issues/217

src/scope/server/frame_processor.py
yondonfu: Tracking in https://github.com/daydreamlive/scope/issues/218

PR #219, opened 2025-12-05
fix: use input frame rate for v2v consumption instead of production rate

ryanontheinside:
## Summary
Fix v2v streaming choppiness when using fast VAEs (LightVAE/TAE) by basing frame consumption rate on measured input FPS rather than pipeline production throughput. This fix is required to introduce faster VAEs and likely other performance enhancements.

## Problem
LightVAE and TAE produced choppy playback during v2v streaming, where the boundaries of the chunks are visible. The slower Wan VAE worked perfectly. Symptoms:
- Choppiness only with fast VAEs
- Test scripts produced smooth output
- Choppiness went away after ~45 seconds or when using more denoise steps
- Choppiness _sometimes_ went away upon updating the prompt
- Lightvae and Wan vae are virtually identical apart from processing speed. Tae is faster still

## Root Cause
The frame consumption rate (how fast WebRTC sends frames to the client) was calculated from **production throughput** (how fast the GPU produces frames) rather than **content temporal rate** (how fast frames should be played).

Eg, when a fast VAE produces 12 frames in 0.3s, the code calculated FPS=40 and sent frames to the client at 40fps. But the video content should maintain its original temporal rate for correct motion - playing it faster causes choppy/jerky appearance.

The test scripts worked because they export with a fixed FPS value, not the production rate.

## Solution
Measure the actual input video frame rate by tracking timestamps of incoming frames, then use that rate for consumption:
- Track timestamps of last 30 incoming frames in `input_loop()`
- Calculate input FPS from frame intervals
- Use input FPS when available (>=5 samples)
- Fall back to existing pipeline FPS calculation otherwise (for t2v mode or during warm-up)

## Tested
- [x] no regression with Wan VAE

### DIFF
--- a/src/scope/server/tracks.py
+++ b/src/scope/server/tracks.py
@@ -3,7 +3,6 @@ import fractions
 import logging
 import threading
 import time
-from collections import deque
 
 from aiortc import MediaStreamTrack
 from aiortc.mediastreams import VIDEO_CLOCK_RATE, VIDEO_TIME_BASE, MediaStreamError
@@ -13,10 +12,6 @@ from .frame_processor import FrameProcessor
 from .pipeline_manager import PipelineManager
 
 logger = logging.getLogger(__name__)
-
-# Input FPS measurement constants
-INPUT_FPS_SAMPLE_SIZE = 30  # Number of frame intervals to track
-INPUT_FPS_MIN_SAMPLES = 5  # Minimum samples needed before using input FPS
 
 
 class VideoProcessingTrack(MediaStreamTrack):
@@ -44,21 +39,13 @@ class VideoProcessingTrack(MediaStreamTrack):
         self._paused_lock = threading.Lock()
         self._last_frame = None
 
-        # Input FPS measurement: track timestamps of incoming frames
-        self._input_frame_times = deque(maxlen=INPUT_FPS_SAMPLE_SIZE)
-        self._input_fps_lock = threading.Lock()
-
     async def input_loop(self):
         """Background loop that continuously feeds frames to the processor"""
         while self.input_task_running:
             try:
                 input_frame = await self.track.recv()
 
-                # Track input frame timestamp for FPS measurement
-                with self._input_fps_lock:
-                    self._input_frame_times.append(time.time())
-
-                # Store raw VideoFrame for later processing
+                # Store raw VideoFrame for later processing (tracks input FPS internally)
                 self.frame_processor.put(input_frame)
 
             except asyncio.CancelledError:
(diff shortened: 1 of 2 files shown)


src/scope/server/tracks.py
```diff
@@ -39,12 +44,20 @@ def __init__(
...
+        self._input_fps_lock = threading.Lock()
+
     async def input_loop(self):
         """Background loop that continuously feeds frames to the processor"""
         while self.input_task_running:
             try:
                 input_frame = await self.track.recv()
 
+                # Track input frame timestamp for FPS measurement
+                with self._input_fps_lock:
```
yondonfu: Can we encapsulate the input FPS tracking logic in FrameProcessor since that class already contains the pipeline FPS tracking logic? We could structure the code similarly to how the pipeline FPS is calculated there for consistency and even use the same [logic](https://github.com/daydreamlive/scope/blob/9f80316809e9598480376eef456215ec83b02b7a/src/scope/server/frame_processor.py#L150) for deciding when to update FPS state vars.

The small tweak we can make is for the FrameProcessor to expose a `get_output_fps()` fn that always returns the output FPS eg the constant FPS that frames should be sent out at. The code would look something like:

```
def get_output_fps(self) -> float:
    input_fps = self._get_input_fps()
    pipeline_fps = self._get_current_pipeline_fps()
    return pipeline_fps if pipeline_fps > input_fps else input_fps

# Refactor from get_current_pipeline_fps() to indicate private
def _get_current_pipeline_fps(self):
    ...

def _get_input_fps(self):
    ...
```

src/scope/server/tracks.py
```diff
@@ -104,10 +144,13 @@ async def recv(self) -> VideoFrame:
         self.initialize_output_processing()
         while self.input_task_running:
             try:
-                # Update FPS from FrameProcessor
-                if self.frame_processor:
+                # Update FPS: prefer input FPS measurement, fall back to FrameProcessor
+                input_fps = self._get_input_fps()
```
yondonfu: I don't think we want to always prefer input FPS because the output FPS would be too high for the pipeline when input FPS > pipeline FPS. We don't actually handle that scenario right now (typically either try to throttle the sender so they reduce their input FPS if possible or drop input frames if former is not possible), but generally I think the safe conditional here would be to only use the input FPS as the output FPS if it is lower than the pipeline FPS. See my other comment for a short code example.

yondonfu (changes requested):
Jotting down my understanding:

The VAE speed should only matter in that it contributes to the overall generation speed of the pipeline eg faster VAE speed results in faster overall generation speed when all other components are held constant. We currently calculate the FPS of the pipeline based on its [overall generation speed](https://github.com/daydreamlive/scope/blob/9f80316809e9598480376eef456215ec83b02b7a/src/scope/server/frame_processor.py#L138) and then [use that FPS](https://github.com/daydreamlive/scope/blob/9f80316809e9598480376eef456215ec83b02b7a/src/scope/server/tracks.py#L110) to ensure that we send out frames [at a constant rate](https://github.com/daydreamlive/scope/blob/9f80316809e9598480376eef456215ec83b02b7a/src/scope/server/tracks.py#L70).

>  But the video content should maintain its original temporal rate for correct motion - playing it faster causes choppy/jerky appearance.

The root problem is when input FPS < pipeline FPS = output FPS results in choppy/jerky appearance in the output? The default input FPS hardcoded in the frontend [is 15](https://github.com/daydreamlive/scope/blob/9f80316809e9598480376eef456215ec83b02b7a/frontend/src/hooks/useVideoSource.ts#L18) so if the VAE speed boost increased the pipeline FPS higher than the input FPS then we could get to this situation.

These changes address the root problem by ensuring that we cannot end up with output FPS > input FPS. We could increase the FPS used in the frontend, but that is a separate concern because regardless of what value is used there we would want to have logic in the backend that handles the scenario where the input FPS < pipeline FPS. And see my other comments about the actual conditional that I think we want for determining the output FPS.

Sound right?

ryanontheinside:
cherry picked and tested in #221

yondonfu (approved):
LGTM!

PR #221, opened 2025-12-05
Add LightVAE support

ryanontheinside:
Rebase after merging #219

### DIFF
--- a/frontend/src/components/SettingsPanel.tsx
+++ b/frontend/src/components/SettingsPanel.tsx
@@ -399,7 +399,7 @@ export function SettingsPanel({
         )}
 
         {/* VAE Type Selection */}
-        {pipelines?.[pipelineId]?.supportsVaeType && (
+        {vaeTypes && vaeTypes.length > 0 && (
           <div className="space-y-2">
             <div className="flex items-center justify-between gap-2">
               <LabelWithTooltip
--- a/src/scope/core/pipelines/base_schema.py
+++ b/src/scope/core/pipelines/base_schema.py
@@ -160,7 +160,6 @@ class BasePipelineConfig(BaseModel):
     supports_cache_management: ClassVar[bool] = False
     supports_kv_cache_bias: ClassVar[bool] = False
     supports_quantization: ClassVar[bool] = False
-    supports_vae_type: ClassVar[bool] = False
     min_dimension: ClassVar[int] = 1
     # Whether this pipeline contains modifications based on the original project
     modified: ClassVar[bool] = False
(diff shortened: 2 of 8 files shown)


frontend/src/components/SettingsPanel.tsx
```diff
@@ -717,6 +726,32 @@ export function SettingsPanel({
                     </SelectContent>
                   </Select>
                 </div>
+
+                <div className="flex items-center justify-between gap-2">
```
yondonfu: Can we move the VAE dropdown below VACE and above LoRAs in the Settings panel so that model related fields are grouped together with pipeline parameters are listed after that?

frontend/src/hooks/useStreamState.ts
```diff
@@ -160,14 +161,19 @@ export function useStreamState() {
...
       try {
-        const [schemasResult, hardwareResult] = await Promise.allSettled([
-          getPipelineSchemas(),
-          getHardwareInfo(),
-        ]);
+        const [schemasResult, hardwareResult, vaeTypesResult] =
+          await Promise.allSettled([
+            getPipelineSchemas(),
+            getHardwareInfo(),
+            getVaeTypes(),
```
yondonfu: Since we've begun the move towards pipelines defining their own schemas can we fold the VAE types into the pipeline schema instead of treat them as separate things?

Eg.

A `vae_type` field in the Pydantic class. AFAICT a Pydantic class can have enums such that a user of the class can inspect it to determine the allowed values - see [this](https://chatgpt.com/share/69582b52-3ef0-800e-961b-6c4b509e91b2). 

Could we do something like the above so that the pipeline schema defines allowable values for the VAE type and the frontend gets all this information as a part of the pipeline schema and can render components based on that?

src/scope/core/pipelines/wan2_1/vae/modules/vae.py
yondonfu, line 1: Let's add a link at the top of the file (eg. "Modified from ...") to indicate where the pruning related changes are from.

src/scope/core/pipelines/base_schema.py
```diff
@@ -160,6 +160,7 @@ class BasePipelineConfig(BaseModel):
     supports_cache_management: ClassVar[bool] = False
     supports_kv_cache_bias: ClassVar[bool] = False
     supports_quantization: ClassVar[bool] = False
+    supports_vae_type: ClassVar[bool] = False
```
yondonfu: I know that we already have a few of these "supports_" fields, but we're also gradually cleaning up the schema.

Given that, is it possible to avoid introducing an extra "supports_" field and instead declare the `vae_type` field as an enum in the Pydantic class and if it exists in the schema then the pipeline implicitly supports it? See [this](https://chatgpt.com/share/69582b52-3ef0-800e-961b-6c4b509e91b2) for a starting point. Might just need to handle the parsing in the frontend [here](https://github.com/daydreamlive/scope/blob/main/frontend/src/hooks/usePipelines.ts).

yondonfu (approved):
LGTM

PR #222, opened 2025-12-05
feat: tae

ryanontheinside:
depends upon #221


src/scope/core/pipelines/wan2_1/vae/tae.py
```diff
@@ -0,0 +1,625 @@
+"""Tiny AutoEncoder (TAE) wrapper for Wan2.1 models.
```
yondonfu, line 2: Let's include a link to where this file came form.

yondonfu (changes requested):

PR #223, opened 2025-12-05
feat: lighttae support

ryanontheinside:
depends upon #222

### DIFF
--- a/src/scope/core/pipelines/wan2_1/vae/tae.py
+++ b/src/scope/core/pipelines/wan2_1/vae/tae.py
@@ -632,9 +632,6 @@ class TAEWrapper(nn.Module):
             self.std.to(device=device, dtype=dtype),
         ]
 
-        # Return in [batch, frames, channels, h, w] format
-        return latent
-
     def decode_to_pixel(
         self, latent: torch.Tensor, use_cache: bool = True
     ) -> torch.Tensor:


yondonfu (approved):

frontend/src/components/SettingsPanel.tsx
```diff
@@ -757,6 +757,32 @@ export function SettingsPanel({
                     </SelectContent>
                   </Select>
                 </div>
+
+                <div className="flex items-center justify-between gap-2">
```
yondonfu: Seems like this can be removed given that https://github.com/daydreamlive/scope/pull/221 already added the Select for VAE types?

src/scope/core/pipelines/wan2_1/vae/tae.py
```diff
@@ -505,21 +506,37 @@ class TAEWrapper(nn.Module):
...
 
-        # Determine checkpoint path
+        # Determine checkpoint path with priority: explicit vae_path > model_dir/model_name default
+        # Both TAE and LightTAE downloaded from lightx2v/Autoencoders
         if vae_path is None:
-            vae_path = os.path.join(model_dir, model_name, DEFAULT_TAE_FILENAME)
+            default_filename = (
+                LIGHTTAE_FILENAME if use_lighttae else DEFAULT_TAE_FILENAME
+            )
+            vae_path = os.path.join(model_dir, "Autoencoders", default_filename)
```
yondonfu: Since the default is Autoencoders what if we just update L515 to:

```
model_name: str = "Autoencoders"
```

and this could be:

```
vae_path = os.path.join(model_dir, model_name, default_filename)
```

which would also let the caller pass in a custom `model_name`.

src/scope/server/app.py
```diff
@@ -786,6 +787,14 @@ async def get_hardware_info():
         raise HTTPException(status_code=500, detail=str(e)) from e
 
 
+@app.get("/api/v1/vae/types", response_model=VaeTypesResponse)
+async def get_vae_types():
```
yondonfu: I don't think we need this anymore since https://github.com/daydreamlive/scope/pull/221 uses the inferred enum values for `vae_type` in the pipeline schema returned?

src/scope/server/schema.py
```diff
@@ -19,6 +19,12 @@ class HealthResponse(BaseModel):
     timestamp: str
 
 
+class VaeTypesResponse(BaseModel):
```
yondonfu: I don't think we need this anymore?

Issue #4, opened 2025-09-25
Fix WebRTC ICE negotiation failure for Docker + specific ISP


leszko:
There is some problem when running inside docker (e.g. RunPod) with some ISP blocking the connection.

```
# Scenario 1: Docker Host Network => Working
2025-09-25 10:46:58,363 - aioice.ice - INFO - Connection(0) Check CandidatePair(('206.168.83.188', 56985) -> ('185.45.245.1', 59436)) State.IN_PROGRESS -> State.SUCCEEDED

# Scenario 2: Docker Bridge Network (Expose 8000) => Not Working
2025-09-25 10:45:47,058 - aioice.ice - INFO - Connection(0) Check CandidatePair(('172.17.0.2', 44992) -> ('185.45.245.1', 54336)) State.IN_PROGRESS -> State.FAILED

# Scenario 3: Docker Bridge Network (Expose 8000), Using Mobile Network => Working
2025-09-25 11:00:48,205 - aioice.ice - INFO - Connection(0) Check CandidatePair(('172.17.0.2', 49454) -> ('37.31.42.149', 13154)) State.IN_PROGRESS -> State.SUCCEEDED
```

yondonfu:
> pull the full ICE config (STUN + TURN) directly from the backend

This is included in [this PR](https://github.com/daydreamlive/scope/pull/224) along with support for trickle ICE.

Needs some more testing (welcome help esp on restrictive networks :)) but will likely merge Monday.

PR #224, opened 2025-12-05
Support fetching ICE servers + trickle ICE

yondonfu:
May fix https://github.com/daydreamlive/scope/issues/4.

This PR introduces 2 changes:

1. The frontend fetches ICE servers from the backend (https://github.com/daydreamlive/scope/pull/224/commits/c2df5f59674dfa3e3fa20758381c9da0549127a6)

This allows the user to set HF_TOKEN in order to configure TURN servers for both the frontend and backend. As a result, the frontend can also use TURN relays in the scenario where its local host candidates are insufficient for establishing a connection with the backend.

2. The frontend and backend use [Trickle ICE](https://bloggeek.me/webrtcglossary/trickle-ice/) (https://github.com/daydreamlive/scope/pull/224/commits/d1b98be668b29f18ee6e3e5668628ba5efbafdad) to speed up the connection process.

Previously, the frontend would wait until ICE gathering completed to send an offer to the backend. The benefit of this approach is that it maximizes the probability of a successful connection because it gathers every possible ICE candidate and then advertises all of them to the backend. The downside of this approach is that it can be slow. I noticed on my local Windows machine that it could take ~30s for ICE gathering to complete (I'm not totally sure why -  maybe additional local network interfaces to wait on or something). I suspect that this approach also would also result in a longer time till connection when combined with 1 because the frontend will have even more ICE candidates to gather based on the ICE servers returned from the backend. I think we might've actually observed that longer time till connection awhile go in a draft PR that attempted to implement 1?

An alternative to waiting until ICE gathering completes is to hardcoded a timeout and send the offer to the backend if ICE gathering completes *or* if that timeout is hit. This could work, but it has a few downsides a) when the frontend/backend are both local we should be able to connect quickly but this approach would take longer in many scenarios (in my case since complete ICE gathering takes a long time on my machine I would have to wait until the timeout say 3-5s) and b) dealing with a hardcoded timeout value is not ideal.

The best solution seems to be to use Trickle ICE. The basic idea is that the frontend will immediately send an offer to the backend and then as it finds ICE candidates it will send them to the backend ("trickling" them to the backend). The backend will then add those candidates to the session so they can be potentially used to establish a connection. The benefit of this approach is that as soon as an ICE candidate is found that can be used for the connection it is used. This means in the fully local scenario we can establish a connection ASAP. And in the scenario where the connection establishment is more complex eg the use of TURN servers on both sides the trickling process will proceed until the right candidates are discovered.

The Trickle ICE impl is based off of:

- [MediaMTX JS code](https://github.com/bluenviron/mediamtx/blob/2ce5fde6e96ee1c4587592d8faddbf14a42ac8de/internal/servers/webrtc/reader.js#L529)
- [MediaMTX WHIP code](https://github.com/bluenviron/mediamtx/blob/2ce5fde6e96ee1c4587592d8faddbf14a42ac8de/internal/servers/webrtc/http_server.go#L230)
- [MediaMTX ICE code](https://github.com/bluenviron/mediamtx/blob/2ce5fde6e96ee1c4587592d8faddbf14a42ac8de/internal/protocols/whip/ice_fragment.go)
- [fastrtc trickle ICE code](https://github.com/gradio-app/fastrtc/blob/main/backend/fastrtc/webrtc_connection_mixin.py)

**Testing**

- [x] Tested on my Windows machine fully local
- [x] Tested with backend on Runpod

I think we should also test:

- The specific Docker situation mentioned in the linked issue.
- Try to use the server.md docs to connect to the server API just to make sure that still works given these changes.

### DIFF
--- a/docs/server.md
+++ b/docs/server.md
@@ -126,12 +126,20 @@ The `/api/v1/pipeline/status` endpoint returns:
 ### For Video-Input Pipelines (e.g., streamdiffusionv2)
 
 ```javascript
-// 1. Create peer connection
+// 1. Fetch ICE servers from backend (includes TURN servers for firewall traversal)
+const iceServersResponse = await fetch("http://localhost:8000/api/v1/webrtc/ice-servers");
+const { iceServers } = await iceServersResponse.json();
+
+// 2. Create peer connection
 const pc = new RTCPeerConnection({
-  iceServers: [{ urls: "stun:stun.l.google.com:19302" }],
+  iceServers: iceServers,
 });
 
-// 2. Create data channel
+// Store session ID for sending ICE candidates
+let sessionId = null;
+const queuedCandidates = [];
+
+// 3. Create data channel
 const dataChannel = pc.createDataChannel("parameters", { ordered: true });
 
 dataChannel.onopen = () => {
@@ -146,14 +154,14 @@ dataChannel.onmessage = (event) => {
   }
 };
 
-// 3. Set up video element
+// 4. Set up video element
 const videoElement = document.createElement("video");
 videoElement.autoplay = true;
 videoElement.muted = true;
 videoElement.playsInline = true;
 document.body.appendChild(videoElement);
 
-// 4. Add local video track
+// 5. Add local video track
 const localStream = await navigator.mediaDevices.getUserMedia({
   video: { width: 512, height: 512 },
 });
@@ -164,7 +172,7 @@ localStream.getTracks().forEach((track) => {
   }
 });
 
-// 5. Set up event handlers
+// 6. Set up event handlers
 const onTrack = (event) => {
   if (event.streams && event.streams[0]) {
     videoElement.srcObject = event.streams[0];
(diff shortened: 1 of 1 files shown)


src/scope/server/app.py
```diff
@@ -336,6 +339,25 @@ async def get_pipeline_schemas():
     return PipelineSchemasResponse(pipelines=pipelines)
 
 
+@app.get("/api/v1/webrtc/ice-servers", response_model=IceServersResponse)
```
leszko, line 342: You may check what's the convention for returning the ICE servers (with credentials). I think in go-livepeer we return it as headers to all requests (or maybe just to OPTIONS req). You may check [here](https://github.com/livepeer/go-livepeer/blob/83ea085f96c0313c4a13bba6cb54941b1937e857/media/whip_server.go#L547).

yondonfu: I see that the headers in that link are the format used for WHIP/WHEP. We don't support WHIP/WHEP right now, but we could so I created [this issue](https://github.com/daydreamlive/scope/issues/228) to track that and if/when we implement it we could use the same header format.

PR #226, opened 2025-12-05
Add RewardForcing pipeline

yondonfu:
A few notes:

- We use SetTransformerBlocksLocalAttnSizeBlock to set the local_attn_size param in the attention blocks similar to what is done for LongLive. This is required because without it the causal model logic won't run the KV cache eviction logic which depends on local_attn_size != -1 in the attention block.
- Initial observation is ~12-13fps @480x832 with 4 denoising steps on a 5090 with Sage Attention 2.
- There is some weird temporal consistency issues during V2V, but T2V looks fine relative to the video samples from the project page

# Example Outputs

https://github.com/user-attachments/assets/28158dd1-294a-423a-b766-8637678b772e

https://github.com/user-attachments/assets/b47afba2-9689-48a5-81cc-3b7b944aca0f

https://github.com/user-attachments/assets/c19a3d17-2cde-40f5-9166-a2cb077a09c3

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ Scope is a tool for running and customizing real-time, interactive generative AI
   - [StreamDiffusionV2](./src/scope/core/pipelines/streamdiffusionv2/docs/usage.md) (text-to-video, video-to-video)
   - [LongLive](./src/scope/core/pipelines/longlive/docs/usage.md) (text-to-video, video-to-video)
   - [Krea Realtime Video](./src/scope/core/pipelines/krea_realtime_video/docs/usage.md) (text-to-video)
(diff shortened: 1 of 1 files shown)


src/scope/core/pipelines/reward_forcing/docs/usage.md
```diff
@@ -0,0 +1,73 @@
...
+```
+# Run from scope directory
+uv run download_models --pipeline reward_forcing
+```
+
+Then:
+
+```
+# Run from scope directory
+uv run -m score.core.pipelines.reward_forcing.test
```
ryanontheinside: typo
`uv run -m scope.core.pipelines.reward_forcing.test`

PR #229, opened 2025-12-09
Fix Temporal Consistency Bug in reward_forcing v2v Mode

ryanontheinside:
### Problem
The reward_forcing pipeline had a temporal consistency bug in video-to-video (v2v) mode where `current_start_frame` tracked latent frames instead of pixel frames, apparently causing misalignment in temporal tracking.

### Root Cause
The shared `PrepareNextBlock` increments `current_start_frame` by latent frame count, but the VAE temporally expands latents to pixels (e.g., 4 latent frames → 13 pixel frames on first batch, 3 latent frames → 12 pixel frames on subsequent batches).

While `current_start_frame` is used internally by the generator/KV cache in latent space, proper v2v temporal consistency requires tracking the actual number of generated pixel frames for seed generation and state management.

**Note:** longlive may have the same underlying issue but uses `RecacheFramesBlock` which maintains a separate latent buffer that could mask the symptoms.

### Solution
Created reward_forcing-specific `PrepareNextPixelFrameBlock` that:
- Takes `output_video` from `DecodeBlock` as input
- Increments `current_start_frame` by actual pixel frames generated
- Ensures proper temporal tracking aligned with VAE output
- Maintains compatibility with existing generator/KV cache usage

NOTE: A more architectural solution would separate latent vs pixel frame tracking globally, but that requires changes across shared blocks and all pipelines. This could be considered more seriously if we implement another pipeline requiring pixel frame tracking. This pipeline-specific fix achieves the same result with minimal impact.

### DIFF
--- a/src/scope/core/pipelines/reward_forcing/blocks/__init__.py
+++ b/src/scope/core/pipelines/reward_forcing/blocks/__init__.py
@@ -1,0 +1,5 @@
+from .prepare_next_pixel import PrepareNextPixelFrameBlock
+
+__all__ = [
+    "PrepareNextPixelFrameBlock",
+]
--- a/src/scope/core/pipelines/reward_forcing/blocks/prepare_next_pixel.py
+++ b/src/scope/core/pipelines/reward_forcing/blocks/prepare_next_pixel.py
@@ -1,0 +1,78 @@
+from typing import Any
+
+import torch
+from diffusers.modular_pipelines import (
+    ModularPipelineBlocks,
+    PipelineState,
+)
+from diffusers.modular_pipelines.modular_pipeline_utils import (
+    ComponentSpec,
+    ConfigSpec,
+    InputParam,
+    OutputParam,
+)
+
+
+class PrepareNextPixelFrameBlock(ModularPipelineBlocks):
+    @property
+    def expected_components(self) -> list[ComponentSpec]:
+        return [
+            ComponentSpec("generator", torch.nn.Module),
+        ]
+
+    @property
+    def expected_configs(self) -> list[ConfigSpec]:
+        return [
+            ConfigSpec("num_frame_per_block", 3),
+        ]
+
+    @property
+    def description(self) -> str:
+        return "Prepare Next block for reward_forcing that tracks pixel frames instead of latent frames for proper v2v temporal alignment"
+
+    @property
+    def inputs(self) -> list[InputParam]:
+        return [
+            InputParam(
+                "current_start_frame",
+                required=True,
+                type_hint=int,
+                description="Current starting frame index of current block",
+            ),
+            InputParam(
+                "latents",
+                required=True,
+                type_hint=torch.Tensor,
+                description="Denoised latents",
+            ),
+            InputParam(
+                "output_video",
+                required=True,
+                type_hint=torch.Tensor,
+                description="Decoded video frames",
+            ),
+        ]
+
+    @property
+    def intermediate_outputs(self) -> list[OutputParam]:
+        return [
+            OutputParam(
+                "current_start_frame",
+                type_hint=int,
+                description="Current starting frame index of current block",
+            ),
+        ]
+
+    @torch.no_grad()
+    def __call__(self, components, state: PipelineState) -> tuple[Any, PipelineState]:
+        block_state = self.get_block_state(state)
+
+        _, _, num_output_frames, _, _ = block_state.output_video.shape
+        # Track pixel frames (not latent frames) for proper v2v temporal alignment.
+        # VAE expands latents to pixels (4 latent → 13 pixel on first batch,
+        # 3 latent → 12 pixel on subsequent batches), so we must track pixel space
+        # to align with input video chunk extraction.
+        block_state.current_start_frame += num_output_frames
+
+        self.set_block_state(state, block_state)
+        return components, state
--- a/src/scope/core/pipelines/reward_forcing/modular_blocks.py
+++ b/src/scope/core/pipelines/reward_forcing/modular_blocks.py
@@ -9,18 +9,19 @@ from ..wan2_1.blocks import (
     DecodeBlock,
     DenoiseBlock,
     EmbeddingBlendingBlock,
-    PrepareNextBlock,
     SetTimestepsBlock,
     SetTransformerBlocksLocalAttnSizeBlock,
     SetupCachesBlock,
     TextConditioningBlock,
 )
+from .blocks import PrepareNextPixelFrameBlock
 
 logger = diffusers_logging.get_logger(__name__)
 
 # Main pipeline blocks with multi-mode support (text-to-video and video-to-video)
 # AutoPreprocessVideoBlock: Routes to video preprocessing when 'video' input provided
 # AutoPrepareLatentsBlock: Routes to PrepareVideoLatentsBlock or PrepareLatentsBlock
+# PrepareNextPixelFrameBlock: reward_forcing-specific block that tracks pixel frames for proper v2v alignment
 ALL_BLOCKS = InsertableDict(
     [
         ("text_conditioning", TextConditioningBlock),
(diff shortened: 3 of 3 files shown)


yondonfu (commented):
I'm not sure I understand these changes so a question...

IIUC `current_start_frame` is supposed to be incremented by the # of latent frames generated because the generator internally is uses the # of latent frames generated thus far multiplied by the # of tokens per latent frame (eg `frame_seq_length`) when indexing into the KV cache. 

So, why do we need to all of sudden increment by the # of pixel frames when this is not required for other models? And why was there not a problem with T2V, but only V2V?

ryanontheinside:
I think I may have gotten to the bottom of this, but I am not sure I understand. Is this making sense to you? Claude used the words "empirical proof" in the below report, but I am considerably less confident than that. There are some additional notes at the very bottom including the additional branch in which I manually hacked in the fix, "proving" this.

## TL;DR

The name "PrepareNextPixelFrameBlock" is somewhat misleading. The fix doesn't actually track "pixel frames" - it tracks the **effective frame stride** (consistently 3 frames) that keeps `current_start_frame` aligned across all four places where it's used. The V2V first-batch special case creates a 0→4→7→10 pattern with latent tracking, but the system needs 0→3→6→9 for temporal consistency. T2V doesn't have this issue because it doesn't have the first-batch special case.

---

## The Root Cause

### Where `current_start_frame` is Actually Used

I investigated all usages of `current_start_frame` in the pipeline. It's used in **four critical places within each iteration**:

1. **PrepareVideoLatentsBlock** (line 119): `block_seed = base_seed + current_start_frame`
   - Seeds the RNG for noise generation
   - Misalignment → inconsistent noise patterns → temporal discontinuities

2. **DenoiseBlock** (line 140): `current_start = current_start_frame * frame_seq_length`
   - Passed to generator for KV cache write positioning
   - Misalignment → cache writes at wrong indices

3. **CleanKVCacheBlock** (line 120): `current_start = current_start_frame * frame_seq_length`
   - Runs generator at timestep=0 to clean cache with final denoised latents
   - Misalignment → cache cleanup at wrong positions

4. **CausalWanSelfAttention** (causal_model.py line 145): `current_start_frame = current_start // frame_seqlen`
   - Used to calculate RoPE positional encoding offsets
   - Misalignment → incorrect temporal positional embeddings

**Critical insight**: All four usages within a single iteration must use the **same** `current_start_frame` value. If they're misaligned, the system breaks.

### Empirical Proof

I implemented manual corrections in all four places (subtracting 1 when current_start_frame > 0) to convert the broken pattern (0, 4, 7, 10, 13...) to the working pattern (0, 3, 6, 9, 12...). **Result: Temporal consistency was fully restored**, proving that alignment across these four usage points is the complete root cause.

### Why V2V But Not T2V?

**T2V mode**: All chunks follow the same pattern
- Chunk 1: 3 latent frames → current_start_frame: 0 + 3 = 3
- Chunk 2: 3 latent frames → current_start_frame: 3 + 3 = 6
- Chunk 3: 3 latent frames → current_start_frame: 6 + 3 = 9
- Pattern: 0, 3, 6, 9, 12... ✓ **Aligned**

**V2V mode with PrepareNextBlock**: First batch has special case
- Chunk 1: **4** latent frames (VAE needs extra frame for temporal overlap) → current_start_frame: 0 + 4 = **4**
- Chunk 2: 3 latent frames → current_start_frame: 4 + 3 = 7
- Chunk 3: 3 latent frames → current_start_frame: 7 + 3 = 10
- Pattern: 0, **4**, 7, 10, 13... ✗ **Misaligned**

The V2V first-batch special case (requiring 4 latent frames) breaks the alignment because:
- Seeds become: base_seed+0, base_seed+**4**, base_seed+7... (should be +0, +3, +6...)
- Cache positions become: 0, **4096**, 7168... (should be 0, 3072, 6144...)
- RoPE offsets become misaligned with actual temporal progression

### Why "PrepareNextPixelFrameBlock" Works (And Why The Name Is Misleading)

From the logs:
```

DecodeBlock: VAE decoded 4 latent frames -> 13 pixel frames
PrepareNextPixelFrameBlock: incrementing current_start_frame from 0 by 3 pixel frames
```

It doesn't increment by 13! It increments by **3** - the **effective frame stride**, which is the number of new non-overlapping frames produced per iteration.

Looking at the actual code:
```python
_, _, num_output_frames, _, _ = block_state.output_video.shape
block_state.current_start_frame += num_output_frames
```

The `output_video` at this point contains only 3 frames (the non-overlapping portion), not the full 13 decoded frames. The VAE's temporal overlaps are already handled earlier in the pipeline.

**The fix works because it maintains the effective stride of 3**, giving us:
- Pattern: 0, 3, 6, 9, 12... ✓ **Aligned**
- Seeds: base_seed+0, +3, +6, +9... ✓ **Consistent**
- Cache positions: 0, 3072, 6144, 9216... ✓ **Aligned**
- RoPE offsets: 0, 3, 6, 9... ✓ **Correct**

### Why Not Required For Other Models?

Other models (longlive) have additional buffering/recaching mechanisms (like `RecacheFramesBlock`) that mask the symptom, so it seems. I remove these and there is a 

The reward_forcing pipeline exposed this issue because it:
- Uses V2V mode with the first-batch special case
- Lacks additional buffering that would hide the misalignment
- Has tight temporal dependencies (reward forcing mechanism) that amplify inconsistencies

### Alternative Considered

A more architectural solution would be to separate latent vs pixel frame tracking globally with something like:
- `current_start_latent_frame` for KV cache indexing
- `current_start_pixel_frame` for seed derivation and temporal tracking

However, this requires changes across all shared blocks and pipelines. The pipeline-specific fix achieves the same result with minimal impact while maintaining full compatibility with the existing architecture.

---

## Summary

The fix doesn't actually track "pixel frames" - it tracks the **effective frame stride** that keeps `current_start_frame` values aligned (0, 3, 6, 9...) across:
1. Seed derivation (PrepareVideoLatentsBlock)
2. KV cache write positions (DenoiseBlock)
3. KV cache cleanup positions (CleanKVCacheBlock)
4. RoPE positional encoding (CausalWanSelfAttention)

The V2V first-batch special case (+4 latent frames) breaks this alignment, and the fix restores it by tracking the effective stride (+3 frames consistently).

A better name might be `PrepareNextWithEffectiveStrideBlock`, but `PrepareNextPixelFrameBlock` was chosen to indicate it operates on the decoded output rather than the latent space.

# Addnl Notes
- ["Proof" branch](https://github.com/daydreamlive/scope/tree/ryanontheinside/fix/reward-forcing-tempoal-fix-proof)
- Does removing prepare_recache_frames and recache_frames blocks from Longlive replicate the issue as seen in reward forcing? kind of. It is worse, and the symptoms do seem similar, but this is not apples to apples. 
- Does adding the PrepareNextPixelFrameBlock to longlive fix the symptoms introduced by removing the recache blocks? It seems so. 
- Something that threw me off quite a bit, is that Longlive worked with all of its normal blocks, but using the new PrepareNextPixelFrameBlock instead of PrepareNextBlock. 

If this *does* make sense to you, I can use it as a basis to think about a potentially more elegant solution.

yondonfu:
Haha so the current solution works by accident.

The shape of `output_video` is BTCHW where T is # of pixel frames and C is # of channels.

```
_, _, num_output_frames, _, _ = block_state.output_video.shape
```

This line is taking the channel count as `num_output_frames`. The channel count is 3 so it is incrementing `current_start_frame` by 3 every time. This seems to fix the behavior. But, only because the channel count is always 3.

yondonfu:
> Isn't it related to this code where we consume one more additional frame at the beginning?

I think it is.

For V2V this means we end up with a current_start_frame progression that looks like this:

0, num_frame_per_block + 1, (num_frame_per_block * 2) + 1...

eg.

0, 4, 7, 10

As opposed to T2V where we have:

0, num_frame_per_block, (num_frame_per_block * 2)...

eg.

0, 3, 6, 9

But, it is not clear to me why the right behavior for V2V should be any different than what we've defined right now. It happens to be the case where if we do 0, 3, 6, 9 instead of 0, 4, 7, 10 for V2V with RewardForcing the behavior improves. 

But, even with that change RewardForcing V2V doesn't feel any better than LongLive (has some more lingering visual issues) so I'm inclined to just make the change in https://github.com/daydreamlive/scope/pull/235 as a special case for now and move on for the time being.

yondonfu:
Closing in favor of https://github.com/daydreamlive/scope/pull/235

PR #239, opened 2025-12-16
Update frontend icon

leszko:


leszko:
Updated to the black background. But THB I like it more when it's transparent. Here's the comparison.
image

PR #241, opened 2025-12-17
Add initial support for plugins

yondonfu:
Adds plugins feature behind a feature flag `DAYDREAM_SCOPE_PREVIEW`.

## Demo

https://github.com/user-attachments/assets/de8d6c09-2784-4d0e-92d5-4439d4d0c158

## Architecture

This is inspired by the [llm](https://github.com/simonw/llm) [plugin system](https://llm.datasette.io/en/stable/plugins/index.html).

A plugin is just a Python package that uses the hook system from [pluggy](https://pluggy.readthedocs.io/en/stable/) that will be installed into the same virtual env as Scope (there is a single virtual env). It can be installed via `uv run daydream-scope install` and uninstalled via `uv run daydream-scope uninstall`. These commands are thin wrappers around `uv pip install` and `uv pip uninstall` allowing us to just rely on typical Python package management and we inherit all the features of `uv pip`.  We use `uv pip` instead of `uv add` or `uv remove` because the latter would modify the project deps, but we want plugins to be installed into the virtual env without changing the project deps. 

A plugin adheres to `ScopeHookSpec`. At the moment, it has a single `hookspec` which is `register_pipelines()`. A plugin must implement this method and decorate it with `@scope.core.hookimpl` ([example](https://github.com/yondonfu/scope-test-generator/blob/main/scope_test_generator/__init__.py#L8)). The `register_pipelines(0` method is responsible for registering one or more pipelines that should be available in Scope.

In addition to the `install` and `uninstall` commands we also introduce:

- `uv run daydream-scope plugins` to list installed plugins
- `uv run daydream-scope pipelines` to list available pipelines

These commands are all implemented using [click](https://click.palletsprojects.com/en/stable/).

The `PipelineRegistry` includes all built-in *and* plugin pipelines.

I've added a simple example of a plugin [here](https://github.com/yondonfu/scope-test-generator/blob/main/scope_test_generator/__init__.py#L8) that can be installed via:

```
uv run daydream-scope install git+https://github.com/yondonfu/scope-test-generator.git
```

https://github.com/daydreamlive/scope/pull/241/commits/22d559aa5a98afb4a2bd6005f2796441056c235c also includes frontend changes to:

- Hide Prompts box if the pipeline does not support prompts eg passthrough
- Fixes an issue where you couldn't press the rewind button if the pipeline does not support prompts
- Moves all the pipeline metadata that was hardcoded in the frontend into the backend as pipeline configs so that that becomes the single source of truth

## Missing

- There is no easy, streamlined path for tying model downloads with pipelines registered by a plugin. This is to be addressed separately.


pyproject.toml
```diff
@@ -67,6 +69,9 @@ Homepage = "https://github.com/daydreamlive/scope"
 Repository = "https://github.com/daydreamlive/scope"
 Issues = "https://github.com/daydreamlive/scope/issues"
 
+[tool.uv]
+preview = true
```
yondonfu, line 72: This silences a warning from uv about extra-build-dependencies usage. I added this because I wanted the CLI output when using commands to be cleaner.

src/scope/server/app.py
```diff
@@ -688,12 +688,133 @@ def main():
...
+    # If no subcommand was invoked, run the server
+    if ctx.invoked_subcommand is None:
+        run_server(reload, host, port, no_browser)
+
+
+@main.command()
+def plugins():
+    """List all installed plugins."""
+
+    @suppress_init_output
```
yondonfu, line 969: The use of this decorator is to suppress logs when using these commands to maintain cleaner CLI output.

src/scope/server/app.py
```diff
@@ -615,40 +643,12 @@ def open_browser_when_ready(host: str, port: int, server):
...
-    args = parser.parse_args()
+def run_server(reload: bool, host: str, port: int, no_browser: bool):
+    """Run the Daydream Scope server."""
 
-    # Handle version flag
-    if args.version:
-        print_version_info()
-        sys.exit(0)
+    from scope.core.pipelines.registry import (
+        PipelineRegistry,  # noqa: F401 - imported for side effects (registry initialization)
```
yondonfu, line 880: The PipelineRegistry registers pipelines at import time right now so this is just to avoid unused import errors/warnings while still ensuring that the pipeline registration happens on server startup.

frontend/src/pages/StreamPage.tsx
```diff
@@ -551,18 +556,23 @@ export function StreamPage() {
...
     if (pipeline) {
       const defaultMethod =
         pipeline.defaultTemporalInterpolationMethod || "slerp";
       const defaultSteps = pipeline.defaultTemporalInterpolationSteps ?? 4;
 
       setTemporalInterpolationMethod(defaultMethod);
       setTransitionSteps(defaultSteps);
+
+      // Clear prompts if pipeline doesn't support them
+      if (pipeline.supportsPrompts === false) {
```
yondonfu, line 630: This is a bit of a hack. If the pipeline does not support prompts we just set the prompt to an empty string for now so that the timeline still works, but it doesn't show any text. This felt like the minimal change to get things working without requiring a larger refactor right now.

frontend/src/components/SettingsPanel.tsx
```diff
@@ -277,26 +278,6 @@ export function SettingsPanel({
                         </Tooltip>
                       </TooltipProvider>
                     )}
-                    {currentPipeline.modified && (
```
yondonfu, line 310: I don't think the `modified` attribute was adding any value so opted to just remove to simplify since we were moving a bunch of attributes to the backend.

src/scope/core/pipelines/registry.py
```diff
@@ -86,5 +89,21 @@ def _register_pipelines():
         PipelineRegistry.register(config_class.pipeline_id, pipeline_class)
 
 
+def _initialize_registry():
+    """Initialize registry with built-in pipelines and plugins."""
+    # Register built-in pipelines first
+    _register_pipelines()
```
leszko, line 178: This will fail on the non-gpu devices, because it will try to import the pipelines which have dependencies on CUDA. We could either try/catch for each import, or make the CUDA import optional in each pipelines.But I thought that maybe even better is to treat this plugin stuff as an opportunity to use only those default pipelines which fit your hardware requirements. In particular:
1. Display Krea only if you run > 5090 (don't display it on 4090)
2. Display only Passthrough if you run on a non-gpu machine (in future, we'll also display Decart API)

I've drafted a PR and would be interested in your thoughts: https://github.com/daydreamlive/scope/pull/243

frontend/src/pages/StreamPage.tsx
leszko: How about hiding the timeline? I think that the timeline does not bring any value if you don't have prompts.

src/scope/core/plugins/hookspecs.py
```diff
@@ -0,0 +1,24 @@
...
+
+hookspec = pluggy.HookspecMarker("scope")
+hookimpl = pluggy.HookimplMarker("scope")
+
+
+class ScopeHookSpec:
+    """Hook specifications for Scope plugins."""
+
+    @hookspec
+    def register_pipelines(self, register):
```
leszko, line 13: Do we need to force people to annotate `@scope.core.hookimpl` in their plugins? Maybe we should just not annotate anything, just when someone installs a plugin, then look into their code and import everything that implements our `Pipeline` interface.

leszko (commented):
Great work @yondonfu — this will significantly improve Scope’s extensibility for both us and the community 🚀

**Thoughts / suggestions (ordered by priority):**

1. **Non-GPU support**  
   This currently doesn’t work on non-GPU setups. I added an [inline comment](https://github.com/daydreamlive/scope/pull/241#discussion_r2626312759) with suggested fix(es).

2. **Separate dependencies per pipeline**  
   We should consider running each pipeline in a separate process (similar to what I started in [#173](https://github.com/daydreamlive/scope/pull/173)). This would allow each pipeline to have its own dependency set and completely avoid dependency conflicts.  
   This would be a killer feature - e.g., work like @BuffMcBigHuge’s “PersonaLive!” seems largely about resolving dependency clashes. Isolating pipelines would dramatically speed up adding new ones and aligns with where we want to go anyway.

3. **Model downloads inside pipelines**  
   Since you already mentioned this: it might be a good time to move model downloading into the pipeline itself. A pipeline plugin without model management feels incomplete and less useful in practice.

4. **Existing pipelines as plugins**  
   What do you think about migrating our existing pipelines to plugins as well? They could be preinstalled, but this would give us consistent behavior across all pipelines and allow users to uninstall default ones if desired.

5. **Plugins vs. pipelines**  
   Do we need both concepts? Perhaps we could simplify by having a single concept (e.g., “plugins”), with each plugin containing exactly one pipeline. No strong opinion - just something to consider.

frontend/src/pages/StreamPage.tsx
yondonfu: I'm inclined to leave it for now to minimize drastic visual layout changes here and there are other things that having an indication of time progression may be useful for.

src/scope/core/plugins/hookspecs.py
yondonfu: This is required for using pluggy which relies on the decorator for detecting the hook impl. I'm inclined to just start with that unless we feel like we have a strong reason to start handrolling something ourselves. But, we can always make improvements on this if we need to.

src/scope/core/pipelines/registry.py
yondonfu: I merged the changes in with some tweaks in https://github.com/daydreamlive/scope/pull/241/commits/2f9fc3373cb39cf74c75b3c50f035794cf85a69e

I opted to relax the check so that if the pipeline has a VRAM req *and* there is no GPU then we skip registration. The reason for this is that I don't think the VRAM req is actually accurate right now so don't think it makes sense to be so restrictive just quite yet. If we're confident in that number then I think the more restrictive check makes sense.

I also removed the extra useEffect and moved the default pipeline logic into an existing useEffect mainly because I found it easier to understand without needing to jump into another useEffect.

yondonfu:
A couple additional changes:

https://github.com/daydreamlive/scope/pull/241/changes/7654e8620e2d868f21c90f64757d8f761a7643e6 moves heavy model related imports into __init__ for built-in pipelines. I found this to be the simplest way to avoid triggering those imports when the pipelines are registered - those imports can be a bit slow so better to trigger them as lazily as possible.

https://github.com/daydreamlive/scope/pull/241/changes/6ab629fcc62b79ee1aa584497d2ff39222f754dd I noticed a frontend bug where I couldn't change the Input mode after switching to a pipeline. I think the logic flaw was that whenever the input mode changed it would check if the input mode was different than the default and then set the default. I'm not sure when this bug was introduced, but figured would push a quick fix.

yondonfu:
@leszko 

> Non-GPU support

Addressed.

> Separate dependencies per pipeline

I'm leaning towards starting off simpler here because 1) separate venvs per plugin can become complex when we do pipeline composition across plugins as we'd need to do something like CUDA IPC to keep data in VRAM (exiting just to move to the other pipeline would be pretty bad for throughput/latency) and because I suspect it would be more complex to ensure a smooth DX from the get go and 2) I'm curious about first evaluating the DX with this and how much dep headaches we can avoid by adding an additional dependency conflict check (probably outside of this PR) [1]. Will follow up offline as well.

[1] I think one of (not the only) reasons that handling deps for plugins is a pain is they are incremental installs (eg using `uv pip install`) into the same shared venv. As a result, you can end up in a situation where you install plugin A (eg package) and then plugin B but end up with a dep version that works with B but not with A. But, it is possible (with a reasonable # of LOC) to detect this incompatibility if you aggregate plugins as they are installed and before finalizing an install you can use `uv sync` or `uv pip compile` to make sure that all the plugins + deps are compatible with the deps in the project. This doesn't make dep conflicts go away, but it does avoid broken deps.

> Model downloads inside pipelines

I'll treat that as something to be addressed next (probably outside of this PR) and want to decide on the shared venv vs separate venv first and test on that. 

> Existing pipelines as plugins

I'm inclined to address this later and focus on plugin UX/DX first.

> Plugins vs. pipelines

I see them as distinct concepts. A plugin can register one or many pipelines and this design leaves room for it introducing other extensions.

PR #256, opened 2025-12-18
01 VACE feat: add image control

ryanontheinside:
This control will be required to support VACE reference conditioning. This PR can be merged independently.

### DIFF
--- a/src/scope/server/app.py
+++ b/src/scope/server/app.py
@@ -32,7 +32,7 @@ from .logs_config import (
 )
 from .models_config import (
     ensure_models_dir,
-    get_images_dir,
+    get_assets_dir,
     get_models_dir,
     models_are_downloaded,
 )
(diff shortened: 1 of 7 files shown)


frontend/src/components/ui/file-picker.tsx
yondonfu, line 1: I think we can revert these changes since the MediaPicker is being used now.

src/scope/server/app.py
```diff
@@ -487,6 +497,191 @@ def process_lora_file(file_path: Path, lora_dir: Path) -> LoRAFileInfo:
         raise HTTPException(status_code=500, detail=str(e)) from e
 
 
+class AssetFileInfo(BaseModel):
```
yondonfu, line 500: Can we put these classes in `schema.py` which is where similar classes live right now?

src/scope/server/schema.py
```diff
@@ -105,6 +105,16 @@ class Parameters(BaseModel):
         default=None,
         description="Spout input configuration for receiving frames from external apps",
     )
+    ref_images: list[str] | None = Field(
```
yondonfu, line 108: Mentioned in another PR but should this be vace_ref_images? Feel free to address here or just lmk if you plan on addressing (or responding) in other PR.

frontend/src/components/ui/file-picker.tsx
yondonfu: Fixed in https://github.com/daydreamlive/scope/pull/269/changes/17cd1ceafff58e85cd5b6f5c21fd89bbb3493a94

src/scope/server/app.py
yondonfu: Fixed in https://github.com/daydreamlive/scope/pull/269/changes/17cd1ceafff58e85cd5b6f5c21fd89bbb3493a94

src/scope/server/schema.py
yondonfu: Fixes in later PR

PR #257, opened 2025-12-18
02 VACE feat: Infrastructure

ryanontheinside:
This PR adds VACE Infrastructure. This can be merged safely.

### DIFF
--- a/src/scope/core/pipelines/wan2_1/blocks/decode.py
+++ b/src/scope/core/pipelines/wan2_1/blocks/decode.py
@@ -32,11 +32,6 @@ class DecodeBlock(ModularPipelineBlocks):
                 type_hint=torch.Tensor,
                 description="Denoised latents",
             ),
-            InputParam(
-                "vace_ref_images",
-                default=None,
-                description="VACE reference images (not used for frame stripping in this implementation)",
-            ),
         ]
 
     @property
@@ -54,12 +49,7 @@ class DecodeBlock(ModularPipelineBlocks):
         block_state = self.get_block_state(state)
 
         # Decode to pixel space
-        latents = block_state.latents
-        video = components.vae.decode_to_pixel(latents, use_cache=True)
-
-        # Note: VACE reference frames are NOT prepended to latents in this implementation.
-        # They are only used for VACE context conditioning, so we don't need to strip them.
-        # The vace_ref_images in state is used for VACE context preparation, not for frame stripping.
+        video = components.vae.decode_to_pixel(block_state.latents, use_cache=True)
 
         block_state.output_video = video
 
--- a/src/scope/core/pipelines/wan2_1/blocks/denoise.py
+++ b/src/scope/core/pipelines/wan2_1/blocks/denoise.py
@@ -109,7 +109,20 @@ class DenoiseBlock(ModularPipelineBlocks):
             InputParam(
                 "vace_context",
                 default=None,
-                description="VACE context for reference image conditioning",
+                type_hint=torch.Tensor | None,
+                description="VACE context that provides visual conditioning",
+            ),
+            InputParam(
+                "input_frames",
+                default=None,
+                type_hint=torch.Tensor | None,
+                description="Input frames for video generation",
+            ),
+            InputParam(
+                "vace_input_frames",
+                default=None,
+                type_hint=torch.Tensor | None,
+                description="VACE conditioning input frames (depth, flow, pose, etc.) for per-chunk guidance",
             ),
             InputParam(
                 "vace_context_scale",
@@ -171,17 +184,6 @@ class DenoiseBlock(ModularPipelineBlocks):
             )
 
             if index < len(denoising_step_list) - 1:
-                # Determine if we should regenerate VACE hints:
-                # - If input_frames exists: per-chunk conditioning -> regenerate every chunk
-                # - If vace_context exists: reference images provided -> regenerate to apply them
-                # This allows ref_images to be sent at any time, not just first chunk
-                has_input_frames = (
-                    getattr(block_state, "input_frames", None) is not None
-                )
-                vace_regenerate_hints = has_input_frames or (
-                    block_state.vace_context is not None
-                )
-
                 _, denoised_pred = components.generator(
                     noisy_image_or_video=noise,
                     conditional_dict=conditional_dict,
@@ -193,7 +195,6 @@ class DenoiseBlock(ModularPipelineBlocks):
                     kv_cache_attention_bias=block_state.kv_cache_attention_bias,
                     vace_context=block_state.vace_context,
                     vace_context_scale=block_state.vace_context_scale,
-                    vace_regenerate_hints=vace_regenerate_hints,
                 )
 
                 next_timestep = denoising_step_list[index + 1]
@@ -216,17 +217,6 @@ class DenoiseBlock(ModularPipelineBlocks):
                     ),
                 ).unflatten(0, denoised_pred.shape[:2])
             else:
-                # Determine if we should regenerate VACE hints (same logic as above)
-                # - If input_frames exists: per-chunk conditioning -> regenerate every chunk
-                # - If vace_context exists: reference images provided -> regenerate to apply them
-                # This allows ref_images to be sent at any time, not just first chunk
-                has_input_frames = (
-                    getattr(block_state, "input_frames", None) is not None
-                )
-                vace_regenerate_hints = has_input_frames or (
-                    block_state.vace_context is not None
-                )
-
                 _, denoised_pred = components.generator(
                     noisy_image_or_video=noise,
                     conditional_dict=conditional_dict,
@@ -238,7 +228,6 @@ class DenoiseBlock(ModularPipelineBlocks):
                     kv_cache_attention_bias=block_state.kv_cache_attention_bias,
                     vace_context=block_state.vace_context,
                     vace_context_scale=block_state.vace_context_scale,
-                    vace_regenerate_hints=vace_regenerate_hints,
                 )
 
         block_state.latents = denoised_pred
--- a/src/scope/core/pipelines/wan2_1/components/generator.py
+++ b/src/scope/core/pipelines/wan2_1/components/generator.py
@@ -215,7 +215,6 @@ class WanDiffusionWrapper(torch.nn.Module):
         kv_cache_attention_bias: float = 1.0,
         vace_context: torch.Tensor | None = None,
         vace_context_scale: float = 1.0,
-        vace_regenerate_hints: bool = True,
     ) -> torch.Tensor:
         prompt_embeds = conditional_dict["prompt_embeds"]
 
@@ -241,7 +240,6 @@ class WanDiffusionWrapper(torch.nn.Module):
                 kv_cache_attention_bias=kv_cache_attention_bias,
                 vace_context=vace_context,
                 vace_context_scale=vace_context_scale,
-                vace_regenerate_hints=vace_regenerate_hints,
             ).permute(0, 2, 1, 3, 4)
         else:
             if clean_x is not None:
@@ -255,7 +253,6 @@ class WanDiffusionWrapper(torch.nn.Module):
                     aug_t=aug_t,
                     vace_context=vace_context,
                     vace_context_scale=vace_context_scale,
-                    vace_regenerate_hints=vace_regenerate_hints,
                 ).permute(0, 2, 1, 3, 4)
             else:
                 if classify_mode:
--- a/src/scope/core/pipelines/wan2_1/vace/__init__.py
+++ b/src/scope/core/pipelines/wan2_1/vace/__init__.py
@@ -5,34 +5,10 @@ Provides VACE model, utilities, and blocks for reference image conditioning
 and structural guidance (depth, flow, pose, etc.) across all Wan2.1 pipelines.
 """
 
-from .blocks.vace_encoding import VaceEncodingBlock
-from .models.attention_blocks import (
-    create_base_attention_block_class,
-    create_vace_attention_block_class,
-)
 from .models.causal_vace_model import CausalVaceWanModel
-from .utils.encoding import (
-    decode_vace_latent,
-    extract_depth_chunk,
-    load_and_prepare_reference_images,
-    preprocess_depth_frames,
-    vace_encode_frames,
-    vace_encode_masks,
-    vace_latent,
-)
 from .utils.weight_loader import load_vace_weights_only
 
 __all__ = [
     "CausalVaceWanModel",
-    "create_vace_attention_block_class",
-    "create_base_attention_block_class",
-    "vace_encode_frames",
-    "vace_encode_masks",
-    "vace_latent",
-    "load_and_prepare_reference_images",
-    "decode_vace_latent",
-    "preprocess_depth_frames",
-    "extract_depth_chunk",
     "load_vace_weights_only",
-    "VaceEncodingBlock",
 ]
--- a/src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
+++ b/src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
@@ -10,8 +10,8 @@ the denoising block. Supports flexible combinations:
 The mode is implicit based on what inputs are provided - no explicit mode parameter needed.
 
 For conditioning inputs (depth, flow, etc.), follows original VACE architecture:
-- input_frames = conditioning maps (3-channel RGB from annotators)
-- input_masks = spatial control masks (ones for full-frame, regional for masked areas)
+- vace_input_frames = source RGB video frames OR conditioning maps (3-channel RGB from annotators)
+- vace_input_masks = spatial control masks (ones for full-frame, regional for masked areas)
 - Standard path: vace_encode_frames -> vace_encode_masks -> vace_latent
 """
 
@@ -45,7 +45,7 @@ class VaceEncodingBlock(ModularPipelineBlocks):
     1. Single Operation: All paths perform the same conceptual operation (VACE encoding)
        with shared logic (VAE selection, reference image loading, validation).
 
-    2. OR Condition: Block should run if EITHER ref_images OR input_frames is provided.
+    2. OR Condition: Block should run if EITHER vace_ref_images OR vace_input_frames is provided.
        AutoPipelineBlock uses sequential first-match logic, making OR conditions awkward
        (would require duplicate block instances with different triggers).
 
@@ -61,7 +61,6 @@ class VaceEncodingBlock(ModularPipelineBlocks):
     def expected_components(self) -> list[ComponentSpec]:
         return [
             ComponentSpec("vae", torch.nn.Module),
-            ComponentSpec("vace_vae", torch.nn.Module, required=False),
         ]
 
     @property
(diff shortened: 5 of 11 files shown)


src/scope/core/pipelines/wan2_1/blocks/decode.py
```diff
@@ -49,7 +54,12 @@ def __call__(self, components, state: PipelineState) -> tuple[Any, PipelineState
         block_state = self.get_block_state(state)
 
         # Decode to pixel space
-        video = components.vae.decode_to_pixel(block_state.latents, use_cache=True)
+        latents = block_state.latents
+        video = components.vae.decode_to_pixel(latents, use_cache=True)
+
+        # Note: VACE reference frames are NOT prepended to latents in this implementation.
```
yondonfu: If we're not doing frame stripping and `vace_ref_images` isn't actually needed as an input can we just revert all changes in this file then?

src/scope/core/pipelines/wan2_1/vace/utils/weight_loader.py
```diff
@@ -0,0 +1,175 @@
+"""
+Utility for selectively loading VACE-specific weights into LongLive model.
```
yondonfu:
```suggestion
Utility for selectively loading VACE-specific weights into a Wan2.1 based causal model.
```

src/scope/core/pipelines/wan2_1/vace/utils/weight_loader.py
```diff
@@ -0,0 +1,175 @@
...
+
+This module extracts only VACE components (vace_blocks, vace_patch_embedding)
+from a VACE checkpoint and loads them into a CausalVaceWanModel that's already
+been initialized with LongLive weights.
```
yondonfu:
```suggestion
been initialized with Wan2.1 based causal model weights.
```

src/scope/core/pipelines/wan2_1/vace/utils/weight_loader.py
```diff
@@ -0,0 +1,175 @@
...
+
+def load_vace_weights_only(model, vace_checkpoint_path: str) -> None:
+    """
+    load_vace_weights_only: Load only VACE-specific weights from checkpoint into model.
+
+    Extracts and loads:
+    - vace_blocks.* (VACE attention blocks for hint generation)
+    - vace_patch_embedding.* (Conv3D for encoding reference images)
+
+    Skips all base model weights since those come from LongLive.
```
yondonfu:
```suggestion
    Skips all base model weights since those come from the causal model.
```

src/scope/core/pipelines/wan2_1/vace/utils/weight_loader.py
```diff
@@ -0,0 +1,175 @@
...
+
+    Args:
+        model: CausalVaceWanModel instance (already has LongLive base weights)
```
yondonfu:
```suggestion
        model: CausalVaceWanModel instance (already has causal model base weights)
```

src/scope/core/pipelines/wan2_1/vace/utils/weight_loader.py
```diff
@@ -0,0 +1,175 @@
...
+    )
+
+    # Check shapes before loading
+    if "vace_patch_embedding.weight" in vace_state_dict:
+        ckpt_shape = vace_state_dict["vace_patch_embedding.weight"].shape
+        model_shape = actual_model.vace_patch_embedding.weight.shape
+        logger.debug(
+            f"load_vace_weights_only: Checkpoint shape: {ckpt_shape}, Model shape: {model_shape}"
+        )
+        if ckpt_shape != model_shape:
```
yondonfu, line 130: Should exit early (i.e. raise an error to be caught by caller) here if the shape mismatch means that we don't have compatible embedding weights.

src/scope/core/pipelines/wan2_1/vace/utils/weight_loader.py
```diff
@@ -0,0 +1,175 @@
...
+    missing_keys, unexpected_keys = actual_model.load_state_dict(
+        vace_state_dict, strict=False
+    )
+
+    # Filter out expected missing keys (all the base model weights)
+    actual_missing = [
+        k for k in missing_keys if any(k.startswith(prefix) for prefix in vace_keys)
+    ]
+
+    if actual_missing:
```
yondonfu, line 149: Should we exit early if we are missing VACE keys as likely means there is a problem?

src/scope/core/pipelines/wan2_1/blocks/denoise.py
```diff
@@ -106,6 +106,17 @@ def inputs(self) -> list[InputParam]:
                 type_hint=float | None,
                 description="Amount of noise added to video",
             ),
+            InputParam(
+                "vace_context",
+                default=None,
```
yondonfu, line 111: Let's add `type_hint=torch.Tensor`

src/scope/core/pipelines/wan2_1/blocks/denoise.py
```diff
@@ -106,6 +106,17 @@ def inputs(self) -> list[InputParam]:
...
+                description="VACE context for reference image conditioning",
```
yondonfu:
```suggestion
                description="VACE context that provides visual conditioning",
```

since IIUC the context consists of concatenated latents encoded from all possible visual inputs eg input frames, ref images, masks, etc.

src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
```diff
@@ -0,0 +1,462 @@
...
+This block handles encoding of VACE conditioning inputs and prepares vace_context for
+the denoising block. Supports flexible combinations:
+- Reference images only (R2V): Static reference images for style/character consistency
+- Conditioning input only: Per-chunk guidance (depth, flow, pose, scribble, etc.)
+- Both combined: Reference images + conditioning input for style + structural guidance
+
+The mode is implicit based on what inputs are provided - no explicit mode parameter needed.
+
+For conditioning inputs (depth, flow, etc.), follows original VACE architecture:
+- input_frames = conditioning maps (3-channel RGB from annotators)
```
yondonfu: The `input_frames` can be conditioning maps, but can be just source RGB video right?

src/scope/core/pipelines/wan2_1/blocks/denoise.py
```diff
@@ -160,6 +171,17 @@ def __call__(self, components, state: PipelineState) -> tuple[Any, PipelineState
             )
 
             if index < len(denoising_step_list) - 1:
+                # Determine if we should regenerate VACE hints:
+                # - If input_frames exists: per-chunk conditioning -> regenerate every chunk
+                # - If vace_context exists: reference images provided -> regenerate to apply them
+                # This allows ref_images to be sent at any time, not just first chunk
+                has_input_frames = (
```
yondonfu: Let's add `input_frames` as an Input (non-required) so that it is easy to see that it as an (optional) input for this block? I don't think anything needs to change with this line though since we IIUC we have the problem where the `default` field of an Input doesn't seem to ensure that the attribute is available on `block_state`.

src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
```diff
@@ -0,0 +1,462 @@
...
+    @property
+    def inputs(self) -> list[InputParam]:
+        return [
+            InputParam(
+                "ref_images",
+                default=None,
+                description="List of reference image paths for style/character consistency (can be combined with input_frames)",
+            ),
+            InputParam(
+                "input_frames",
```
yondonfu: WDYT about adding `vace_` prefix to `ref_images`, `input_frames` and `input_masks` since they are all VACE specific and given that we already use the prefix for `vace_context` and `vace_context_scale`?

I'm keeping in mind that these will be params that can be included in a pipeline call and I think it would be clearer to have `pipeline(vace_input_frames=...)` vs. `pipeline(input_frames=...)` where the latter doesn't feel as clear eg might lead to the question of whether `input_frames` just the input video for regular V2V.

src/scope/core/pipelines/wan2_1/components/generator.py
```diff
@@ -250,6 +269,8 @@ def forward(
                         cls_pred_branch=self._cls_pred_branch,
                         gan_ca_blocks=self._gan_ca_blocks,
                         concat_time_embeddings=concat_time_embeddings,
+                        vace_context=vace_context,
+                        vace_context_scale=vace_context_scale,
```
yondonfu, line 270: Do we need to pass in `vace_regenerate_hints` here and on L283?

src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
yondonfu: Can we add type_hint for `ref_images`, `input_frames` and `input_masks`?

src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
```diff
@@ -0,0 +1,462 @@
...
+            )
+
+        self.set_block_state(state, block_state)
+        return components, state
+
+    def _encode_reference_only(self, components, block_state, current_start):
+        """
+        Encode reference images only (R2V mode) following original VACE pattern.
+
+        R2V characteristics (from notes/VACE/vace/models/wan/wan_vace.py lines 127-155):
```
yondonfu: For comments like this can if we're pointing to a file from another repo can we use a permalink (in this case something like https://github.com/ali-vilab/VACE/blob/48eb44f1c4be87cc65a98bff985a26976841e9f3/vace/models/wan/wan_vace.py#L127)

src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
```diff
@@ -0,0 +1,462 @@
...
+            # - masks=None (no masking for R2V)
+            # - pad_to_96=True (pads to 96 channels for VACE compatibility)
+            vace_context = vace_encode_frames(
+                vace_vae,
+                dummy_frames,
+                ref_images=[prepared_refs],
+                masks=None,
+                pad_to_96=True,
+            )
+        else:
```
yondonfu: IIUC the original VACE always uses zero/dummy frames if input_frames is None. Is the following an optimization? When would one want to use this path vs. the other one?

src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
```diff
@@ -0,0 +1,462 @@
...
+
+        # Load and prepare reference images
+        prepared_refs = load_and_prepare_reference_images(
+            ref_image_paths,
+            block_state.height,
+            block_state.width,
+            components.config.device,
+        )
+
+        # CRITICAL: Use vace_vae for encoding reference images.
```
yondonfu: Does passing use_cache=False to encode_with_latent with the existing VAE work since it just creates a temp one-off cache for a normal encode without invoking the streaming encode impl? If so, that would avoid this extra dep which seems to just point to the same WanVideoWrapper in the next PR + the conditionals.

src/scope/core/pipelines/wan2_1/vace/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+    load_and_prepare_reference_images,
+    preprocess_depth_frames,
+    vace_encode_frames,
+    vace_encode_masks,
+    vace_latent,
+)
+from .utils.weight_loader import load_vace_weights_only
+
+__all__ = [
+    "CausalVaceWanModel",
```
yondonfu, line 12: Do all of these need to be exported right now? Would err on side of keeping exports minimal and selectively export when needed just to start with smaller surface area - easier to add then remove if other packages import.

src/scope/core/pipelines/wan2_1/vace/models/attention_blocks.py
```diff
@@ -0,0 +1,189 @@
+# Modified from notes/VACE/vace/models/wan/modules/model.py
```
yondonfu: Use permalink URL as mentioned elsewhere for easier referencing

src/scope/core/pipelines/wan2_1/vace/models/causal_vace_model.py
```diff
@@ -0,0 +1,433 @@
...
+            self._original_block_class
+        )
+
+        # Replace blocks with hint-injection-enabled versions
+        self._replace_blocks_with_hint_injection_support()
+
+        # Create VACE blocks (parallel processing path for reference images)
+        self._create_vace_blocks()
+
+        # VACE patch embedding (separate encoder for reference images)
```
yondonfu:
```suggestion
        # VACE patch embedding
```

a slight terminology nit since the patch embedding isn't really an encoder in the typical sense of how the term is used.

src/scope/core/pipelines/wan2_1/vace/utils/encoding.py
```diff
@@ -0,0 +1,344 @@
...
+            ref_img = white_canvas
+        else:
+            ref_img = ref_img.to(device)
+
+        prepared_refs.append(ref_img)
+
+    return prepared_refs
+
+
+def decode_vace_latent(vae, zs, ref_images=None):
```
yondonfu: I don't think is used anywhere and I think it's because we don't use frame stripping from the original VACE - will have a separate question about that elsewhere.

src/scope/core/pipelines/wan2_1/vace/utils/encoding.py
```diff
@@ -0,0 +1,344 @@
...
+    for z, refs in zip(zs, ref_images, strict=False):
+        if refs is not None:
+            # Remove reference frames from latent
+            z = z[:, len(refs) :, :, :]
+        trimmed_zs.append(z)
+
+    return vae.decode(trimmed_zs)
+
+
+def preprocess_depth_frames(depth_frames, target_height, target_width, device):
```
yondonfu: I think [preprocess_chunk](https://github.com/daydreamlive/scope/blob/783f3121a357c35697e9f37bff052fb6f2b81041/src/scope/core/pipelines/process.py#L9) already does everything here and is more general and if there is anything missing could be incorporated there.

The only thing that seems like it might be depth map specific is that L312 assumes that the video is grayscale so it converts to 3 channel RGB, but we could prob have a conditional in preprocess_chunk that handles that and checks if the input is grayscale and then ensures it is 3 channel RGB.

src/scope/core/pipelines/wan2_1/vace/utils/encoding.py
```diff
@@ -0,0 +1,344 @@
...
+    depth_frames = depth_frames.repeat(1, 3, 1, 1)
+
+    # Move to device and add batch dimension: [F, 3, H, W] -> [1, 3, F, H, W]
+    depth_frames = depth_frames.to(device)
+    depth_frames = depth_frames.unsqueeze(0).permute(0, 2, 1, 3, 4)
+
+    return depth_frames
+
+
+def extract_depth_chunk(depth_video, chunk_index, num_frames_per_chunk):
```
yondonfu: A utility for this doesn't seem necessary or specific to depth?

src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
ryanontheinside: This appears to be a vestige of my experimentation. I incorrectly attributed some strange behavior to these dummy frames, and at the time it made sense (concatenated on temporal dimension for bidirectionality). This is not the case, and i have apparently fixed the real problem, so I will just remove this path. There is no difference in the output with or without the frames, but better to match the original where possible.

ryanontheinside: https://github.com/user-attachments/assets/9ce840f7-dfd8-4870-97f6-9eb775241408

https://github.com/user-attachments/assets/9adfb49f-5932-4bdf-aff5-17f521be6e08



Update: the previous strange behavior i mentioned has been resolved, but there is still undesirable behavior when dummy frames are used. This behavior is not strange; its explainable with my previous rationale.

I am leaning towards never using dummy frames here - in that case i will document in a design document as you suggested instead of leaving this parameter

src/scope/core/pipelines/wan2_1/blocks/denoise.py
yondonfu: Actually noticed it can be None so this should be `type_hint=torch.Tensor | None`

src/scope/core/pipelines/wan2_1/vace/design.md
```diff
@@ -0,0 +1,243 @@
+# VACE Implementation Design
+
+This document outlines the key architectural differences between the Scope VACE implementation and the original VACE implementation from Alibaba's repository.
+
+## Overview
+
+Our VACE (Video-Augmented Conditional Encoding) implementation adapts the original design for causal/streaming generation in the Wan2.1 pipeline. While both implementations use the same hint injection mechanism at the transformer level, we diverge fundamentally in how reference images are integrated with video latents during the diffusion process.
```
yondonfu:
```suggestion
Our VACE (All-In-One Video Creation and Editing) implementation adapts the original design for causal/streaming generation in the Wan2.1 pipeline. While both implementations use the same hint injection mechanism at the transformer level, we diverge fundamentally in how reference images are integrated with video latents during the diffusion process.
```

src/scope/core/pipelines/wan2_1/vace/design.md
```diff
@@ -0,0 +1,243 @@
...
+    return x
+```
+
+The hint injection uses simple **residual addition** (not cross-attention). Zero-initialized projection layers (`self.after_proj`) in VACE blocks ensure hints start with no effect until trained.
+
+## Secondary Difference: No Dummy Frames for R2V Mode
+
+### Original VACE Approach
+
+For R2V (reference-to-video) generation, the original VACE preprocessing uses dummy (grayscale) frames as temporal placeholders when generating from reference images only:
```
yondonfu: I might be missing something here, but it doesn't look like the original VACE uses grayscale frames as placeholders?

AFAICT, when there is no source video, they use just use [zero tensors](https://github.com/ali-vilab/VACE/blob/48eb44f1c4be87cc65a98bff985a26976841e9f3/vace/models/wan/wan_vace.py#L214) which is  passed as the [input_frames arg](https://github.com/ali-vilab/VACE/blob/48eb44f1c4be87cc65a98bff985a26976841e9f3/vace/vace_wan_inference.py#L293) here which is then used in VACE context encoding [here](https://github.com/ali-vilab/VACE/blob/48eb44f1c4be87cc65a98bff985a26976841e9f3/vace/models/wan/wan_vace.py#L339).

src/scope/core/pipelines/wan2_1/vace/design.md
ryanontheinside: tired last night, too much faith in claude

src/scope/core/pipelines/wan2_1/vace/design.md
ryanontheinside: youre not missing anything, def zeros

ryanontheinside: fyi, i did use zeros in the implementation (since entirely removed), its only the design document with this mistake

PR #258, opened 2025-12-18
03 VACE feat: Longlive + Windows video support

ryanontheinside:
Depends upon #257  

Add VACE support for Longlive

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,6 +108,9 @@ dev = [
     "freezegun>=1.5.5",
 ]
 
+[tool.pytest.ini_options]
+testpaths = ["tests"]
+
 [tool.ruff]
 line-length = 88
 target-version = "py310"


frontend/src/lib/api.ts
```diff
@@ -382,6 +382,7 @@ export const getAssetUrl = (assetPath: string): string => {
   return `/api/v1/assets/${encodeURIComponent(filename)}`;
 };
 
+
```
yondonfu: Stray newline can be removed.

src/scope/core/pipelines/longlive/pipeline.py
```diff
@@ -168,6 +172,173 @@ def __init__(
...
+                    active_adapters[0] if len(active_adapters) == 1 else active_adapters
+                )
+                vace_wrapped_model.set_adapter(adapter_name)
+                logger.info(f"_enable_vace: Restored active adapter(s): {adapter_name}")
+
+        self.components.generator.model = vace_wrapped_model
+
+        # Step 6: Load separate VACE VAE for encoding
+        start = time.time()
+        vace_vae = WanVAEWrapper(
```
yondonfu: As inquired in other PRs I wonder if we can avoid doing this if we're using encode_with_latent(..,, use_cache=False) as it looks like we're still using the same WanVAEWrapper for the `vace_vae` here as well as the normal `vae` so it feels like we should be able to use the same one and just make sure not to use the streaming encode + cache.

src/scope/core/pipelines/longlive/pipeline.py
```diff
@@ -179,6 +350,10 @@ def __call__(self, **kwargs) -> torch.Tensor:
         return self._generate(**kwargs)
 
     def _generate(self, **kwargs) -> torch.Tensor:
+        # Lazy load VACE if needed
```
yondonfu, line 344: I don't love the lazy loading here because:

1. It makes managing the UX of the UI and UX more difficult. At the moment, the API can manage the delay of a pipeline load by indicating that the pipeline is not ready yet in its response and the UI can do the same by showing a "Loading..." indicator. If loading VACE weights happens outside of the pipeline loading process then there is yet another delay at that point.  In the UI, if a ref image is provided when the stream is started, the user sees "Loading...", "Connecting..." and then after the connection is established they just see a blank screen for a few seconds because VACE weights are now being loaded.
2. If a ref image is not provided when the stream starts and then the user sets a ref image mid-stream then in the UI the stream freezes while VACE weights are loaded.

If VACE weights loading was faster (eg something like < 500ms) then it might make more sense to allow it be dynamic/lazy. But, given that it takes a couple seconds, I think it would better for it to be incorporated into the pipeline load. This would mean that the UI/API would need to signal whether VACE should be enabled at pipeline load time (assuming we don't want to always have VACE enabled right now or to at least have the option of disabling).

The best solution that I can think of right now that doesn't require more drastic changes is to treat VACE as a toggle that works similarly to how LoRA adapters work with the permanent merge strategy - you decide up front before loading a pipeline whether you want VACE enabled. It could be enabled by default for pipelines that support it and could be toggled off.

Here is a sketch of what the VACE toggle could look like in the Settings panel.

Image

I think this design might also present a short-term solution to another problem: how does the server decide whether to pass incoming video chunks with the `video` param to a pipeline thereby triggering regular V2V (eg preparing noisy latents using the video) vs. the `input_frames` param to a pipeline thereby triggering VACE visual conditioning (and using pure noise for denoising). If VACE is enabled the incoming video chunks would be passed as `input_frames` to the pipeline and we would always use the video for VACE conditioning. If VACE is disabled the incoming video chunks would be passed as `video` to the pipeline for regular V2V. And this would allow us to avoid introducing another Input Mode in the UI right now because this can all happen under the current Video mode. We wouldn't have an immediate way to support passing both `video` and `input_frames` to a pipeline though eg use a video chunk to create noisy latents *and* use either the same video chunk or a different video chunk (for the latter, the UI would need to send two video streams) for VACE conditioning at the same time. I feel like that's ok for now given that most use cases should be covered by either regular V2V or VACE conditioning without a need for combining the two and would just be a limitation of the current structure of the UI. We could always re-visit if needed.

I imagine that in the short-term the video input when VACE is enabled could be pre-computed depth, flow, pose video *or* normal source video. If it is the former, then we get control map guided generation. Later on, if we chain a fast depth pipeline with a diffusion pipeline then then when using that composite pipeline a normal source video could be used since the depth maps would be computed on the fly and then used to guide generation.

src/scope/core/pipelines/video.py
```diff
@@ -1,8 +1,34 @@
...
+import numpy as np
 import torch
 from einops import rearrange
-from torchcodec.decoders import VideoDecoder
 from torchvision.transforms import v2
 
+logger = logging.getLogger(__name__)
+
+# Try to import torchcodec, fallback to OpenCV if unavailable
+_torchcodec_available = False
```
yondonfu: I'll see if there is a way to avoid having to do this...

src/scope/core/pipelines/longlive/pipeline.py
ryanontheinside: i addressed the vae in the #257

src/scope/core/pipelines/video.py
yondonfu: I opened https://github.com/daydreamlive/scope/pull/268 on top of this branch which should remove the need for torchcodec *and* cv2.

src/scope/core/pipelines/longlive/pipeline.py
yondonfu: This is addressed in a later PR.

PR #259, opened 2025-12-18
04 VACE feat: stream diffusionv2

ryanontheinside:
Depends upon #257 

Add VACE support for StreamDiffusionV2

### DIFF
--- a/src/scope/core/pipelines/video.py
+++ b/src/scope/core/pipelines/video.py
@@ -3,6 +3,7 @@ import torch
 from einops import rearrange
 from torchvision.transforms import v2
 
+
 def load_video(
     path: str,
     num_frames: int = None,


src/scope/core/pipelines/streamdiffusionv2/modules/causal_model.py
```diff
@@ -21,7 +21,9 @@
 # wan 1.3B model has a weird channel / head configurations and require max-autotune to work with flexattention
 # see https://github.com/pytorch/pytorch/issues/133254
 # change to default for other models
-flex_attention = torch.compile(flex_attention, dynamic=False, mode="max-autotune")
+
+#TODO: benchmark vs max-autotune. If perf dictates, we could make the mode conditional based on use vace.
+flex_attention = torch.compile(flex_attention, dynamic=False, mode="max-autotune-no-cudagraphs")
```
ryanontheinside, line 25: @yondonfu I want to draw your attention to this. This is required for VACE. We could make this conditional, but wanted your input before adding complexity.

src/scope/core/pipelines/streamdiffusionv2/modular_blocks.py
```diff
@@ -14,12 +14,14 @@
...
     TextConditioningBlock,
 )
+from ..wan2_1.vace.blocks import VaceEncodingBlock
 
 logger = diffusers_logging.get_logger(__name__)
 
 # Main pipeline blocks with multi-mode support (text-to-video and video-to-video)
 # AutoPreprocessVideoBlock: Routes to video preprocessing when 'video' input provided
 # AutoPrepareLatentsBlock: Routes to PrepareVideoLatentsBlock or PrepareLatentsBlock
+# VaceEncodingBlock: Encodes VACE context (R2V or depth) for conditioning
```
yondonfu:
```suggestion
# VaceEncodingBlock: Encodes VACE context for conditioning
```

since its not just depth figured to avoid confusion to just drop

src/scope/core/pipelines/streamdiffusionv2/pipeline.py
```diff
@@ -151,6 +154,92 @@ def __init__(
...
+        """
+        ref_images = kwargs.get("ref_images")
+        input_frames = kwargs.get("input_frames")
+
+        has_ref_images = ref_images is not None and len(ref_images) > 0
+        has_input_frames = input_frames is not None
+
+        return has_ref_images or has_input_frames
+
+    def _enable_vace(self):
```
yondonfu, line 188: I think this is the same impl as in LongLive in the previous PR. If VACE can be used with any Wan2.1 based model (eg RewardForcing), seems like this can be extracted somewhere to be re-used (already used in 2 places, soon to be 3 in RewardForcing). I think you mentioned the possibility of something like a VACEEnabledPipeline similar to the LoRAEnabledPipeline mixin. Now that I've gone through the first couple of PRs I think that might make sense!

src/scope/core/pipelines/streamdiffusionv2/test_r2v.py
yondonfu, line 1: Rename to `test_vace.py` to match LongLive?

src/scope/core/pipelines/streamdiffusionv2/modules/causal_model.py
yondonfu: I ran a very simple, not the most scientific benchmark off `main` using the streamdiffusionv2 test script. I ran the script 3x for each configuration and these are the results from the 3rd run for each configuration:

**max-autotune**
=== Performance Statistics ===
Latency - Avg: 0.30s, Max: 0.73s, Min: 0.23s
FPS - Avg: 14.10, Max: 17.15, Min: 6.83

**max-autotune-no-cudagraphs**
=== Performance Statistics ===
Latency - Avg: 0.27s, Max: 0.78s, Min: 0.24s
FPS - Avg: 15.80, Max: 16.54, Min: 6.40

I ran the test scripts a few more times after and from eyeballing things doesn't seem like a significant difference between the two configurations.

Given this I think it's ok to not introduce a conditional based on whether VACE is used. Let's just mention in the comment that this was changed from max-autotune to max-autotune-no-cudagraphs specifically because the max-autotune did not play well with the VACE code.

src/scope/core/pipelines/streamdiffusionv2/test_r2v.py
ryanontheinside: this test script only tests r2v as opposed to all vace concepts like the longlive script does. that was the rationale for the naming, but happy to change it

yondonfu: Ah ok let's just keep it then for now!

src/scope/core/pipelines/streamdiffusionv2/pipeline.py
yondonfu: This is addressed in a later PR.

PR #260, opened 2025-12-18
05 VACE feat: server + frontend

ryanontheinside:
depends upon #257, #258 , #259

### DIFF
--- a/src/scope/core/pipelines/video.py
+++ b/src/scope/core/pipelines/video.py
@@ -3,6 +3,7 @@ import torch
 from einops import rearrange
 from torchvision.transforms import v2
 
+
 def load_video(
     path: str,
     num_frames: int = None,


src/scope/server/frame_processor.py
yondonfu, line 1: As noted offline, I tried running this branch with pre-computed depth, flow, pose videos and the output for the pre-computed depth input looked different in the UI vs. when I ran the VACE test script for LongLive locally with the same input video. I believe the reason for this is that `input_frames`, which is used for VACE conditioning, is never passed to the pipeline right now. The `video` param is used to pass input video, but that is only used for regular V2V. I mentioned a potential way to setup the logic for deciding when to set `input_frames` in [this comment](https://github.com/daydreamlive/scope/pull/258#discussion_r2637440636) so the issue described here can be addressed once there is a decision made for that comment.

frontend/src/data/pipelines.ts
```diff
@@ -30,7 +30,7 @@ export const PIPELINES: Record<string, PipelineInfo> = {
     docsUrl:
       "https://github.com/daydreamlive/scope/blob/main/src/scope/core/pipelines/streamdiffusionv2/docs/usage.md",
     about:
-      "A streaming pipeline and autoregressive video diffusion model from the creators of the original StreamDiffusion project. The model is trained using Self-Forcing on Wan2.1 1.3b with modifications to support streaming.",
+      "A streaming pipeline and autoregressive video diffusion model from the creators of the original StreamDiffusion project. The model is trained using Self-Forcing on Wan2.1 1.3b with modifications to support streaming. Includes VACE (Video-Aware Condition Encoding) for reference image conditioning in text mode.",
```
yondonfu:
```suggestion
      "A streaming pipeline and autoregressive video diffusion model from the creators of the original StreamDiffusion project. The model is trained using Self-Forcing on Wan2.1 1.3b with modifications to support streaming. Includes VACE support for expanded creation and editing capabilities.",
```

frontend/src/data/pipelines.ts
```diff
@@ -46,7 +46,7 @@ export const PIPELINES: Record<string, PipelineInfo> = {
     docsUrl:
       "https://github.com/daydreamlive/scope/blob/main/src/scope/core/pipelines/longlive/docs/usage.md",
     about:
-      "A streaming pipeline and autoregressive video diffusion model from Nvidia, MIT, HKUST, HKU and THU. The model is trained using Self-Forcing on Wan2.1 1.3b with modifications to support smoother prompt switching and improved quality over longer time periods while maintaining fast generation.",
+      "A streaming pipeline and autoregressive video diffusion model from Nvidia, MIT, HKUST, HKU and THU. The model is trained using Self-Forcing on Wan2.1 1.3b with modifications to support smoother prompt switching and improved quality over longer time periods while maintaining fast generation. Includes VACE (Video-Aware Condition Encoding) for reference image conditioning.",
```
yondonfu:
```suggestion
      "A streaming pipeline and autoregressive video diffusion model from Nvidia, MIT, HKUST, HKU and THU. The model is trained using Self-Forcing on Wan2.1 1.3b with modifications to support smoother prompt switching and improved quality over longer time periods while maintaining fast generation. Includes VACE support for expanded creation and editing capabilities",
```

VACE seems to actually stand for Video All-In-One Creation and Editing, but that's a mouthful lol so think we can just use acronym since that's all that anyone in open source AI seems to use as a term.

frontend/src/pages/StreamPage.tsx
```diff
@@ -463,6 +478,29 @@ export function StreamPage() {
     // Note: Adding/removing LoRAs requires pipeline reload
   };
 
+  const handleRefImagesChange = (images: string[]) => {
+    updateSettings({ refImages: images });
+    // Note: Changing reference images in dropdown only updates local state.
+    // Use "Send Hint" button to send hints to backend during streaming.
```
yondonfu: Stale comment to be removed

frontend/src/data/pipelines.ts
yondonfu: Fixed in https://github.com/daydreamlive/scope/pull/269/changes/342979363157e5aef3d8520d9a6a4cd82e9ac526

frontend/src/data/pipelines.ts
yondonfu: Fixes in https://github.com/daydreamlive/scope/pull/269/changes/342979363157e5aef3d8520d9a6a4cd82e9ac526

src/scope/server/frame_processor.py
```diff
@@ -716,12 +716,42 @@ def process_chunk(self):
...
             if video_input is not None:
-                call_params["video"] = video_input
+                vace_enabled = getattr(pipeline, "vace_enabled", False)
+                if vace_enabled:
+                    # VACE mode: route to vace_input_frames for conditioning
+                    # video_input is list[torch.Tensor] from prepare_chunk where each is (1, H, W, C)
+                    # Need to convert to [B, C, F, H, W] format expected by VACE encoding block
+                    from scope.core.pipelines.process import preprocess_chunk
+
+                    vace_input_tensor = preprocess_chunk(
```
yondonfu, line 730: Moved this into a modular pipeline block in https://github.com/daydreamlive/scope/pull/269/commits/b2507dba2314156048c1be5643db71af03d993d4 to align with how preprocessing is handled with normal V2V.

PR #287, opened 2025-12-29
feat: Longlive VACE FFLF pipeline layer support

ryanontheinside:
This adds FFLF support for Longlive at the pipeline layer. API and UI to follow.

### DIFF
--- a/src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
+++ b/src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
@@ -302,16 +302,14 @@ class VaceEncodingBlock(ModularPipelineBlocks):
         first_frame_image = block_state.first_frame_image
         last_frame_image = block_state.last_frame_image
 
-        images_to_load = []
+        # Load reference images based on mode
         if extension_mode == "firstframe":
             images_to_load = [first_frame_image]
         elif extension_mode == "lastframe":
             images_to_load = [last_frame_image]
         elif extension_mode == "firstlastframe":
-            if current_start == 0:
-                images_to_load = [first_frame_image]
-            else:
-                images_to_load = [last_frame_image]
+            # Load BOTH images for firstlastframe mode
+            images_to_load = [first_frame_image, last_frame_image]
 
         prepared_refs = load_and_prepare_reference_images(
             images_to_load,
@@ -320,27 +318,24 @@ class VaceEncodingBlock(ModularPipelineBlocks):
             components.config.device,
         )
 
-        if hasattr(components, "vace_vae") and components.vace_vae is not None:
-            vace_vae = components.vace_vae
-        else:
-            vace_vae = components.vae
-
-        vae_dtype = next(vace_vae.parameters()).dtype
+        vae = components.vae
+        vae_dtype = next(vae.parameters()).dtype
 
         num_frames = (
             components.config.num_frame_per_block
             * components.config.vae_temporal_downsample_factor
         )
 
-        ref_at_start = extension_mode == "firstframe" or (
-            extension_mode == "firstlastframe" and current_start == 0
-        )
+        # Determine ref placement
+        ref_at_start = extension_mode in ("firstframe", "firstlastframe")
+        ref_at_end = extension_mode in ("lastframe", "firstlastframe")
 
         frames, masks = self._build_extension_frames_and_masks(
             prepared_refs=prepared_refs,
             num_frames=num_frames,
             temporal_group_size=components.config.vae_temporal_downsample_factor,
             ref_at_start=ref_at_start,
+            ref_at_end=ref_at_end,
             device=components.config.device,
             dtype=vae_dtype,
             height=block_state.height,
@@ -351,7 +346,7 @@ class VaceEncodingBlock(ModularPipelineBlocks):
         masks_to_encode = [masks]
 
         z0 = vace_encode_frames(
-            vae=vace_vae,
+            vae=vae,
             frames=frames_to_encode,
             ref_images=[None],
             masks=masks_to_encode,
@@ -385,6 +380,7 @@ class VaceEncodingBlock(ModularPipelineBlocks):
         num_frames: int,
         temporal_group_size: int,
         ref_at_start: bool,
+        ref_at_end: bool,
         device: torch.device,
         dtype: torch.dtype,
         height: int,
@@ -394,10 +390,13 @@ class VaceEncodingBlock(ModularPipelineBlocks):
         Build frames and masks for extension mode with reference frame replication.
 
         Args:
-            prepared_refs: List of prepared reference images [C, 1, H, W]
+            prepared_refs: List of prepared reference images [C, 1, H, W].
+                For firstframe/lastframe: single image.
+                For firstlastframe: [first_image, last_image].
             num_frames: Total number of frames to generate
             temporal_group_size: Number of frames in a temporal VAE group
-            ref_at_start: True for firstframe mode, False for lastframe mode
+            ref_at_start: True to place reference at start (firstframe, firstlastframe)
+            ref_at_end: True to place reference at end (lastframe, firstlastframe)
             device: Target device
             dtype: Target dtype for frames
             height: Frame height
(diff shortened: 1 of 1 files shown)


src/scope/core/pipelines/wan2_1/vace/utils/encoding.py
```diff
@@ -58,14 +58,16 @@ def vace_encode_frames(
...
+
         inactive_stacked = torch.stack(inactive, dim=0).to(dtype=vae_dtype)
         reactive_stacked = torch.stack(reactive, dim=0).to(dtype=vae_dtype)
-        # Use cache=True to ensure cache consistency for both inactive and reactive portions
-        inactive_out = vae.encode_to_latent(inactive_stacked, use_cache=True)
-        reactive_out = vae.encode_to_latent(reactive_stacked, use_cache=True)
-        # Transpose [B, F, C, H, W] -> [B, C, F, H, W] and concatenate along channel dim
+
+        inactive_out = vae.encode_to_latent(inactive_stacked, use_cache=use_cache)
+        reactive_out = vae.encode_to_latent(reactive_stacked, use_cache=False)
```
yondonfu: `use_cache` was set to True here in https://github.com/daydreamlive/scope/pull/283 to fix an issue with white flashing when using depth control videos. 

1. Is this change necessary for FFLF?
2. If yes, we need a different solution for the white flashing issue with depth control videos.

ryanontheinside: Thanks for catching this. It is required for the gradual last frame use case, which I think is an important one. The temporal blending weakens the diffusion-driven transformation such that it looks more like simple latent interpolation than anything else. 

https://github.com/user-attachments/assets/4661daf0-2637-43f0-aa78-06447f7b93f6


https://github.com/user-attachments/assets/1adefba5-30f9-4271-ab82-24d674a9774a

yondonfu: 1. Are you planning on addressing the white flashing issue with depth maps separately then?
2. What are these videos showing? Is it a comparison of using cache and not using cache during encoding?

src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
```diff
@@ -96,6 +101,22 @@ def inputs(self) -> list[InputParam]:
                 type_hint=torch.Tensor | None,
                 description="Spatial control masks [B, 1, F, H, W]: defines WHERE to apply conditioning (white=generate, black=preserve). Defaults to ones (all white) when None. Works with any vace_input_frames type.",
             ),
+            InputParam(
+                "extension_mode",
+                default=None,
+                type_hint=str,
+                description="Extension mode for temporal generation: 'firstframe' (ref at start, generate after), 'lastframe' (generate before, ref at end), or 'firstlastframe' (refs at both ends). Applies to specific chunks based on current_start_frame.",
```
yondonfu: Do we need an explicit concept of extension mode? Or can the mode be implicit based on whether `first_frame_image` and/or `last_frame_image` is provided?

Eg.

1. If `first_frame_image`, but no `last_frame_image` then we start with `first_frame_image` and generate rest.
2. If `last_frame_image`, but no `first_frame_image` then we generate everything with `last_frame_image` at end.
3. If both are present, then start with `first_frame_image` end with `last_frame_image` and generate everything else.

yondonfu: @ryanontheinside Following up on offline convo:

I have a pref for making the extension mode implicit based on whether `first_frame_image` and/or `last_frame_image` is provided because it would simplify the pipeline API usage by avoiding the need for an additional `extension_mode` param.

src/scope/core/pipelines/wan2_1/vace/utils/encoding.py
yondonfu: @ryanontheinside Following up on an offline convo:

Just want to make clear the problem I see with setting `use_cache=False` for the reactive portion right now in your change. This seems like it would be a regression if this PR is merged as-is.

I used this as the input control video:

https://github.com/user-attachments/assets/e0a9f294-914c-4435-95a6-f340d27b6258

I applied this diff:

```
diff --git a/src/scope/core/pipelines/longlive/test_vace.py b/src/scope/core/pipelines/longlive/test_vace.py
index 2d81c60..a8b9036 100644
--- a/src/scope/core/pipelines/longlive/test_vace.py
+++ b/src/scope/core/pipelines/longlive/test_vace.py
@@ -45,16 +45,16 @@ from .pipeline import LongLivePipeline
 CONFIG = {
     # ===== MODE SELECTION =====
     "use_r2v": False,  # Reference-to-Video: condition on reference images
-    "use_depth": False,  # Depth guidance: structural control via depth maps
+    "use_depth": True,  # Depth guidance: structural control via depth maps
     "use_inpainting": False,  # Inpainting: masked video-to-video generation
-    "use_extension": True,  # Extension mode: temporal generation (firstframe/lastframe/firstlastframe)
+    "use_extension": False,  # Extension mode: temporal generation (firstframe/lastframe/firstlastframe)
     # ===== INPUT PATHS =====
     # R2V: List of reference image paths (condition entire video, don't appear in output)
     "ref_images": [
         "frontend/public/assets/example.png",  # path/to/image.png
     ],
     # Depth: Path to depth map video (grayscale or RGB, will be converted)
-    "depth_video": "vace_tests/control_frames_depth.mp4",  # path/to/depth_video.mp4
+    "depth_video": "vace_tests/AnimateDiff_00003_scaled_5x.mp4",  # path/to/depth_video.mp4
     # Inpainting: Input video and mask video paths
     "input_video": "frontend/public/assets/test.mp4",  # path/to/input_video.mp4
     "mask_video": "vace_tests/circle_mask.mp4",  # path/to/mask_video.mp4
@@ -65,14 +65,14 @@ CONFIG = {
     # ===== GENERATION PARAMETERS =====
     "prompt": None,  # Set to override mode-specific prompts, or None to use defaults
     "prompt_r2v": "",  # Default prompt for R2V mode
-    "prompt_depth": "a cat walking towards the camera",  # Default prompt for depth mode
+    "prompt_depth": "a woman dancing",  # Default prompt for depth mode
     "prompt_inpainting": "a fireball",  # Default prompt for inpainting mode
     "prompt_extension": "",  # Default prompt for extension mode
-    "num_chunks": 2,  # Number of generation chunks
+    "num_chunks": 50,  # Number of generation chunks
     "frames_per_chunk": 12,  # Frames per chunk (12 = 3 latent * 4 temporal upsample)
-    "height": 512,
-    "width": 512,
-    "vace_context_scale": 1.5,  # VACE conditioning strength
+    "height": 480,
+    "width": 832,
+    "vace_context_scale": 1.0,  # VACE conditioning strength
     # ===== INPAINTING SPECIFIC =====
     "mask_threshold": 0.5,  # Threshold for binarizing mask (0-1)
     "mask_value": 127,  # Gray value for masked regions (0-255)
@@ -490,7 +490,7 @@ def main():
     print("Initializing pipeline...")
```

I ran:

```
uv run -m scope.core.pipelines.longlive.test_vace
```

I got:

https://github.com/user-attachments/assets/b37220fe-76da-4126-a85d-efc0d5a28e1f

Observe the white flashing effect throughout this output video.

src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
```diff
@@ -253,6 +286,162 @@ def _encode_reference_only(self, components, block_state, current_start):
...
+        first_frame_image = block_state.first_frame_image
+        last_frame_image = block_state.last_frame_image
+
+        images_to_load = []
+        if extension_mode == "firstframe":
+            images_to_load = [first_frame_image]
+        elif extension_mode == "lastframe":
+            images_to_load = [last_frame_image]
+        elif extension_mode == "firstlastframe":
+            if current_start == 0:
```
yondonfu: I'm not sure that I understand what's going on here? It doesn't seem like we'd ever load both `first_frame_image` and `last_frame_image`?

src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
```diff
@@ -253,6 +286,162 @@ def _encode_reference_only(self, components, block_state, current_start):
...
+                images_to_load = [last_frame_image]
+
+        prepared_refs = load_and_prepare_reference_images(
+            images_to_load,
+            block_state.height,
+            block_state.width,
+            components.config.device,
+        )
+
+        if hasattr(components, "vace_vae") and components.vace_vae is not None:
```
yondonfu: I don't think we have a concept of a `vace_vae` anymore?

PR #293, opened 2025-12-30
Run AutoPreprocessVideoBlock step after SetupCache

leszko:


yondonfu:
FYI @leszko pushed [df8ee17](https://github.com/daydreamlive/scope/pull/293/commits/df8ee1764a81263182266f7dc6d4cbcef659b508) which mves PreprocessVideo after SetupCaches for all pipelines because that is results in the correct behavior (eg. when caches are reset we need an additional pixel frame for the next generation during regular V2V) even if explicit errors were not triggered (I suspect this is because streamdiffusionv2 uses a smaller num_frame_per_block = 1 so 4 pixel frames for the first VAE encode is just too few frames overall - when num_frame_per_block = 3 for the other pipelines 12 pixel frames for the first VAE encode won't trigger an error, but really we should pass in 13 pixel frames due to how the VAE handles the first pixel frame on the first call after a cache reset).

PR #297, opened 2025-12-30
Krea VACE 14b Pipeline

BuffMcBigHuge:
This PR implements full VACE (Video All-In-One Creation and Editing) support for the KREA Realtime Video pipeline. It addresses significant architectural challenges in adapting the VACE system—originally designed for the 1.3B "Longlive" model—to work with the much larger Wan 2.1 14B model used by KREA.


.gitignore
```diff
@@ -20,3 +20,6 @@ notes/
 .cursorrules
 .cursorignore
 .cursorindexingignore
+
+.cursor/
+.specstory/
```
BuffMcBigHuge, line 25: @yondonfu I can remove this if you like. Spec story is a plugin for cursor chat history export that I use.

src/scope/core/pipelines/krea_realtime_video/blocks/recompute_kv_cache.py
```diff
@@ -138,6 +163,27 @@ def __call__(self, components, state: PipelineState) -> PipelineState:
...
+        #
+        # This is an algorithmic trade-off: skipping recompute may increase drift / error
+        # accumulation, but can save ~10–15% of pipeline time on B300 (cu130).
+        #
+        # Semantics:
+        # - Always run recompute for the first block (current_start_frame == 0).
+        # - Only start skipping once the cache is in steady-state:
+        #     current_start_frame >= kv_cache_num_frames
+        # - Recompute every N "blocks" (a block produces num_frame_per_block frames).
+        every_n = int(os.getenv("SCOPE_KV_CACHE_RECOMPUTE_EVERY", "1") or "1")
```
leszko: I'm not sure if we use env variables in other blocks, but I think we may want to try to use input params instead. This would help someone wanting to reuse the block or the pipeline to interact with it.

ryanontheinside:
> FWIW, I've played with this PR and I'm seeing this weird loop. @ryanontheinside mentioned it's probably that we're sending the reference image to each chunk.
> 
>  demo_krea.mp4

It could also be a problem with the vae after looking at another example. I had claude replicate the test_vace.py script from longlive and tested krea offline. The demarcation at chunk boundaries is familiar and makes me think vae issue. "She jumps":

https://github.com/user-attachments/assets/8730177b-3e75-40ef-854f-a84cea6a9a06

I also tested longlive; there is no regression there.

src/scope/core/pipelines/krea_realtime_video/pipeline.py
```diff
@@ -36,6 +36,15 @@
 
 
 class KreaRealtimeVideoPipeline(Pipeline, LoRAEnabledPipeline):
```
ryanontheinside, line 38: Can we use `VACEEnabledPipeline` here instead of the lazy loading? We switched to the mixin recently

src/scope/server/frame_processor.py
```diff
@@ -719,7 +719,18 @@ def process_chunk(self):
             # Route video input based on VACE status
             # We do not support combining normal V2V (denoising from noisy video latents) and VACE V2V editing
             if video_input is not None:
-                vace_enabled = getattr(pipeline, "vace_enabled", False)
+                # Route based on frontend's VACE intent (not pipeline.vace_enabled which is lazy-loaded)
+                # This fixes the chicken-and-egg problem where VACE isn't enabled until vace_input_frames arrives
+                vace_enabled = self.parameters.get("vace_enabled", False)
+
+                # Robustness fallback: If vace_enabled is missing (e.g. old frontend build),
```
ryanontheinside: we could expect the front end to match for the build here instead of adding fallbacks, wdyt?

src/scope/core/pipelines/krea_realtime_video/blocks/recompute_kv_cache.py
ryanontheinside: this is interesting - if you do move to input params would be good so see some documentation explaining what different values can be used

src/scope/core/pipelines/krea_realtime_video/blocks/recompute_kv_cache.py
```diff
@@ -11,6 +13,29 @@
 from scope.core.pipelines.wan2_1.utils import initialize_kv_cache
 
 
+def _get_block_mask_model(model: torch.nn.Module) -> torch.nn.Module:
```
ryanontheinside: could this be at the class level as opposed to the module level?

src/scope/core/pipelines/wan2_1/blocks/setup_caches.py
```diff
@@ -101,6 +104,12 @@ def inputs(self) -> list[InputParam]:
                 default=None,
                 description="Input frames for VACE conditioning (if present, indicates video input is enabled)",
             ),
+            InputParam(
+                "soft_transition_active",
```
ryanontheinside: Assuming this is vestigial from experimentation - AFAIK this is never set to true

src/scope/core/pipelines/krea_realtime_video/blocks/recompute_kv_cache.py
BuffMcBigHuge: Modified!

src/scope/core/pipelines/wan2_1/blocks/setup_caches.py
BuffMcBigHuge: Great find - removed.

src/scope/core/pipelines/krea_realtime_video/blocks/recompute_kv_cache.py
BuffMcBigHuge: More testing needed - I'll remove it for now.

src/scope/core/pipelines/krea_realtime_video/pipeline.py
BuffMcBigHuge: I spent some time [attempting this](https://github.com/daydreamlive/scope/commit/8b66b9f3cc84a7ca67f6a90636d855c82d242a41), but kept running into OOM on 5090. Any insight?

yondonfu:
Superseded by https://github.com/daydreamlive/scope/pull/311

PR #302, opened 2026-01-03
feat: decouple R2V reference images from V2V input video conditioning

ryanontheinside:
Add vace_use_input_video parameter to separate reference image conditioning from input video conditioning. These are orthogonal concerns that should not be inextricably linked. This will be equally important for FFLF.

### DIFF
--- a/frontend/src/components/SettingsPanel.tsx
+++ b/frontend/src/components/SettingsPanel.tsx
@@ -126,7 +126,7 @@ export function SettingsPanel({
   spoutAvailable = false,
   vaceEnabled = true,
   onVaceEnabledChange,
-  vaceUseInputVideo = false,
+  vaceUseInputVideo = true,
   onVaceUseInputVideoChange,
   vaceContextScale = 1.0,
   onVaceContextScaleChange,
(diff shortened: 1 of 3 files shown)


src/scope/server/frame_processor.py
```diff
@@ -720,7 +720,11 @@ def process_chunk(self):
             # We do not support combining normal V2V (denoising from noisy video latents) and VACE V2V editing
             if video_input is not None:
                 vace_enabled = getattr(pipeline, "vace_enabled", False)
-                if vace_enabled:
+                vace_use_input_video = self.parameters.get(
```
yondonfu: If the default is False this would be a breaking change for the few folks that are using VACE V2V in their projects via API right now. Unless there is a strong reason we need to make that breaking change now, let's make the default True instead.

frontend/src/components/SettingsPanel.tsx
```diff
@@ -124,6 +126,8 @@ export function SettingsPanel({
   spoutAvailable = false,
   vaceEnabled = true,
   onVaceEnabledChange,
+  vaceUseInputVideo = false,
```
yondonfu: Let's default to True here to match the backend?

ryanontheinside: done

yondonfu (approved):
LGTM

PR #310, opened 2026-01-06
fix: TAE memory pollution in VACE dual-encode with explicit cache obj…

ryanontheinside:
…ects

TAE's MemBlock architecture mixes processed memory state with input via channel concatenation. When VACE encodes inactive and reactive streams interleaved (for depth/flow/pose/inpainting modes), they pollute each other's memory, causing visual artifacts.

Root cause: Unlike WanVAE's CausalConv3d which prepends raw frames as temporal context, TAE's MemBlock accumulates processed state that gets corrupted when different content streams share the same memory.

Solution: Explicit encoder cache objects following WanVAE's feat_cache pattern. Each VACE stream gets its own isolated cache instance:

- Add TAEEncoderCache dataclass to hold MemBlock memory state
- Add create_encoder_cache() to TAEWrapper (returns cache) and WanVAEWrapper (returns None for compatibility)
- Add encoder_cache parameter to encode_to_latent()
- Update vace_encode_frames() with inactive_cache/reactive_cache params
- Update VaceEncodingBlock to manage caches across chunks

This provides:
- Explicit ownership: caller creates and manages cache lifecycle
- Temporal continuity: each stream maintains its own memory across chunks
- No cross-contamination: inactive and reactive encodings stay isolated
- Backwards compatibility: default internal cache when none provided
- Works for all VACE modes: inpainting, depth, r2v
- Works in absence of VACE: t2v, v2v

Tested with Longlive.

This has implications for #287 which i will address.

### DIFF
--- a/src/scope/core/pipelines/longlive/test_vace.py
+++ b/src/scope/core/pipelines/longlive/test_vace.py
@@ -35,9 +35,9 @@ from .pipeline import LongLivePipeline
 
 CONFIG = {
     # ===== MODE SELECTION =====
-    "use_r2v": True,  # Reference-to-Video: condition on reference images
+    "use_r2v": False,  # Reference-to-Video: condition on reference images
     "use_depth": False,  # Depth guidance: structural control via depth maps
-    "use_inpainting": False,  # Inpainting: masked video-to-video generation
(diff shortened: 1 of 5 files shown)


yondonfu (approved):
Nice! LGTM!

PR #311, opened 2026-01-06
Krea VACE 14b Pipeline - Mixin

BuffMcBigHuge:
1. **VACE architecture unification**
   - Migrated `KreaRealtimeVideoPipeline` from lazy loading to the unified `VACEEnabledPipeline` mixin
   - Enhanced mixin with `vace_layers` support, FP8 quantization, and text encoder CPU offloading
   - Unified VACE handling across all pipelines

2. **Krea V2V prompt reset bug fix**
   - Fixed prompt transition issue in V2V/VACE mode
   - Implemented mode-specific temporal interpolation defaults (0 for video mode, 4 for text mode)
   - Frontend now dynamically adjusts transition steps based on input mode

3. **Code quality improvements**
   - Refactored KV cache recomputation code (moved helper to class level)
   - Removed optional optimization code for simplicity
   - Resolved merge conflicts cleanly
   
This PR is a derivative of https://github.com/daydreamlive/scope/pull/297.

**Note:** _This branch doesn't allow for Krea + VACE on 32GB VRAM._

### DIFF
--- a/src/scope/core/pipelines/krea_realtime_video/pipeline.py
+++ b/src/scope/core/pipelines/krea_realtime_video/pipeline.py
@@ -142,11 +142,12 @@ class KreaRealtimeVideoPipeline(Pipeline, LoRAEnabledPipeline, VACEEnabledPipeli
                 block.compile(fullgraph=False)
 
         # Load VAE using create_vae factory (supports multiple VAE types)
+        # Note: VAE is shared across all Wan model sizes, stored in Wan2.1-T2V-1.3B
+        # By not sending model_name, create_vae will use the default "wan" VAE type.
         vae_type = getattr(config, "vae_type", "wan")
         start = time.time()
         vae = create_vae(
             model_dir=model_dir,
-            model_name=base_model_name,
             vae_type=vae_type,
             vae_path=vae_path,
         )


src/scope/core/pipelines/krea_realtime_video/modules/causal_model.py
```diff
@@ -380,12 +381,14 @@ def qkv_fn(x):
                 roped_query = rope_apply(q, grid_sizes, freqs).type_as(v)
                 roped_key = rope_apply(k, grid_sizes, freqs).type_as(v)
 
-                local_end_index = roped_key.shape[1]
-                kv_cache["k"][:, :local_end_index] = roped_key
-                kv_cache["v"][:, :local_end_index] = v
+                # Only update kv_cache if it exists (VACE forward passes kv_cache=None)
+                if kv_cache is not None:
```
ryanontheinside, line 386: We skip this with is_tf = False anyways, but this check is separately unreachable since kv_cache is always None here

src/scope/core/pipelines/krea_realtime_video/pipeline.py
```diff
@@ -83,7 +84,32 @@ def __init__(
         for block in generator.model.blocks:
             block.self_attn.fuse_projections()
 
-        # Initialize optional LoRA adapters on the underlying model BEFORE quantization.
+        # Load text encoder before VACE initialization (may be offloaded to CPU)
+        start = time.time()
+        text_encoder = WanTextEncoderWrapper(
```
ryanontheinside, line 89: I wonder if we should be more discriminate with the CPU offloading of the text encoder or exclude it entirely. I think you had mentioned that the Mixin approach already precludes 5090's with VACE+Krea, in which case we should probably not offload the text encoder at all. However if this does enable 5090+VACE+Krea, that will surely only be for development purposes, so maybe we do this with a flag and document it. As of now, the text encoder is always offloaded for Krea even when there is sufficient VRAM

src/scope/core/pipelines/wan2_1/vace/models/causal_vace_model.py
```diff
@@ -222,61 +227,83 @@ def _replace_blocks_with_hint_injection_support(self):
...
                 if key in new_state:
-                    new_state[key] = orig_state[key].clone()
+                    # Copy to CPU (where new_block is)
+                    new_state[key] = orig_state[key].to("cpu")
 
             new_block.load_state_dict(new_state, strict=False, assign=True)
             new_block.block_id = saved_block_id
 
+            # Move new block to target device/dtype
+            new_block = new_block.to(device=orig_device, dtype=orig_dtype)
```
ryanontheinside, line 263: IIUC, i think the memory optimization order is reversed. Currently, new_block is moved to GPU before orig_block is moved to CPU causing both blocks to be on GPU simultaneously

src/scope/core/pipelines/krea_realtime_video/pipeline.py
```diff
@@ -83,7 +84,32 @@ def __init__(
...
+            model_name=base_model_name,
+            model_dir=model_dir,
+            text_encoder_path=text_encoder_path,
+            tokenizer_path=tokenizer_path,
+        )
+        print(f"Loaded text encoder in {time.time() - start:3f}s")
+        # Move text encoder to target device but use dtype of weights
+        text_encoder = text_encoder.to(device=device)
```
ryanontheinside, line 97: if device is cuda device, this allocates to GPU, but _init_vace in mixin.py immediately moves to CPU

src/scope/core/pipelines/wan2_1/vace/utils/weight_loader.py
```diff
@@ -158,8 +158,9 @@ def load_vace_weights_only(model, vace_checkpoint_path: str) -> None:
             raise ValueError(error_msg)
 
     # Load into actual model (not PEFT wrapper)
+    # Use assign=True to preserve original tensor dtype (important for FP8 weights)
     missing_keys, unexpected_keys = actual_model.load_state_dict(
-        vace_state_dict, strict=False
+        vace_state_dict, strict=False, assign=True
```
ryanontheinside: Does assign=True interact correctly with the .to(device, dtype) calls in
  mixin.py (lines 157-158)? Those move VACE components to GPU before this
  load happens - wondering if assign=True might replace them with CPU tensors
  from the checkpoint, making that earlier allocation unnecessary.

src/scope/core/pipelines/krea_realtime_video/pipeline.py
BuffMcBigHuge: Great point. I added it in anyways but now I'm realizing that it adds quite a bit of latency when changing prompts. I think supporting 32gb is a stretch and wiser to skip offloading all together.

src/scope/core/pipelines/wan2_1/vace/utils/weight_loader.py
BuffMcBigHuge: I think you're right, i suppose this makes the gpu/cpu assignment duplicated. This is now fixed, great find.

src/scope/core/pipelines/wan2_1/vace/models/causal_vace_model.py
BuffMcBigHuge: Hmm I made the [change](https://github.com/daydreamlive/scope/pull/311/commits/f53eeecf1662208322feb91671003743f0a69cc5) but didn't see any VRAM saving.

src/scope/core/pipelines/wan2_1/vace/mixin.py
```diff
@@ -32,45 +36,54 @@ class VACEEnabledPipeline:
...
         Args:
             config: Pipeline configuration that may contain:
                 - 'vace_path': Path to VACE checkpoint (enables VACE)
                 - 'vace_in_dim': Input dimension for VACE (default 96)
+                - 'vace_layers': List of layer indices where VACE is applied (default every 2nd)
             model: Underlying diffusion model (CausalWanModel or wrapped version)
             device: Target device for VACE components
             dtype: Target dtype for VACE components
+            quantization: Optional quantization config (e.g., Quantization.FP8_E4M3FN)
+            text_encoder: Optional text encoder to offload to CPU before VACE loading
```
ryanontheinside: is this used?

BuffMcBigHuge: Cleaned this up!

PR #313, opened 2026-01-07
Add preprocesors

leszko:
This PR adds support for **chained pipelines**, allowing the frontend to configure and order preprocessing steps while keeping the backend pipeline model generic.

https://github.com/user-attachments/assets/64a35c30-b18a-4402-8812-c92e035de295

---

### How It Works

1. **Frontend**
   - The **VACE** panel includes a select box to add and reorder preprocessors.
   - Currently available: **Video Depth Anything** (more can be added via plugins).
   - The frontend sends an **ordered list of pipeline IDs** to the backend.

2. **Backend**
   - There is no special “preprocessor” concept - everything is a **Pipeline**.
   - UI visibility is controlled by the pipeline schema field `usage`, e.g.:
     ```python
     usage = ["preprocessor", "pipeline"]
     ```
     This displays a pipeline as both a preprocessor and a pipeline in the UI.

---

## Notes

- Each pipeline runs in its own thread (`pipeline_processor.py`); significant logic was moved from `frame_processor.py`.
- FPS calculation logic is unchanged; it can be optimized in a separate PR.
- FPS throttling per pipeline is not included; it mostly works without it but should be added later for resource-heavy preprocessors.
- Any pipeline update (e.g. prompt change) is currently propagated to each pipeline in the chain.

---

## Performance
I measured at 5090 Tensordock.

- **LongLive + VACE + Video Depth Anything:** ~10 FPS  
- **LongLive + VACE:** ~11.8 FPS  
- **First PoC:** ~11.3 FPS ([PR #298](https://github.com/daydreamlive/scope/pull/298))

The FPS drop is because Video Depth Anything now runs in **streaming mode** (1 frame at a time) instead of batch mode (used in the first PoC, 32 frames in/out), trading throughput for lower latency.

### DIFF
--- a/src/scope/core/pipelines/video_depth_anything/pipeline.py
+++ b/src/scope/core/pipelines/video_depth_anything/pipeline.py
@@ -96,6 +96,7 @@ class VideoDepthAnythingPipeline(Pipeline):
         # Convert input to numpy uint8 array
         # Frames from frame_processor are always (1, H, W, C), so we squeeze the T dimension
         frames = []
+        target_shape = None
         for frame in video:
             frame_np = (
                 frame.cpu().numpy()
@@ -103,7 +104,24 @@ class VideoDepthAnythingPipeline(Pipeline):
                 else np.array(frame)
             )
             # Squeeze T dimension: (1, H, W, C) -> (H, W, C)
-            frames.append(frame_np.squeeze(0))
+            frame_np = frame_np.squeeze(0)
+            # Use first frame's shape as target for consistency
+            if target_shape is None:
+                target_shape = frame_np.shape
+            elif frame_np.shape != target_shape:
+                # Resize frame to match target shape using torch interpolate
+                frame_tensor = torch.from_numpy(frame_np).permute(2, 0, 1).unsqueeze(0)
+                frame_tensor = torch.nn.functional.interpolate(
+                    frame_tensor.float(),
+                    size=(target_shape[0], target_shape[1]),
+                    mode="bilinear",
+                    align_corners=False,
+                )
+                frame_np = frame_tensor.squeeze(0).permute(1, 2, 0).numpy()
+                # Preserve original dtype
+                if frames[0].dtype == np.uint8:
+                    frame_np = frame_np.astype(np.uint8)
+            frames.append(frame_np)
         frames = np.stack(frames, axis=0)
 
         # Ensure uint8 format [0, 255]


frontend/src/components/SettingsPanel.tsx
```diff
@@ -414,6 +431,200 @@ export function SettingsPanel({
                     />
                   </div>
                 </div>
+
+                {/* Preprocessors Selector */}
```
yondonfu: Since we're just starting off here can we simplify and just support a single preprocessor in the UI for now given that we're still fleshing out our thinking around general chaining which may obviate the need for an explicit preprocessor concept? We can always update this later once we start using preprocessors in the wild.

Image

It could look like this and just default to None and have a dropdown.

frontend/src/components/SettingsPanel.tsx
```diff
@@ -414,6 +431,200 @@ export function SettingsPanel({
...
+                              onValueChange={newId => {
+                                const newIds = [...preprocessorIds];
+                                newIds[index] = newId;
+                                onPreprocessorIdsChange?.(newIds);
+                              }}
+                              disabled={isStreaming || isLoading}
+                            >
+                              <SelectTrigger className="h-8">
+                                <SelectValue>
+                                  {preprocessor?.name || id}
```
yondonfu: Let's use the pipeline ID to be consistent with the pipeline ID dropdown in the Settings panel?

src/scope/core/pipelines/base_schema.py
```diff
@@ -171,6 +171,10 @@ class BasePipelineConfig(BaseModel):
     # quantization=null is recommended, otherwise fp8_e4m3fn is recommended.
     # None means no specific recommendation (pipeline doesn't benefit from quantization).
     recommended_quantization_vram_threshold: ClassVar[float | None] = None
+    # Usage types: list of usage strings indicating how this pipeline can be used.
+    # Default is ["pipeline"] meaning it can be used as a main pipeline.
+    # Can also include "preprocessor" to allow use as a preprocessor.
+    usage: ClassVar[list[str]] = ["pipeline"]
```
yondonfu: Do we need a "pipeline" value? It seems that every pipeline should be usable as a standalone/main pipeline and the only extra information that needs to be captured is whether it should also be designated as a preprocessor.

If we remove this value then that would also avoid confusing scenarios like the one we have with video-depth-anything where its usage field is just set to preprocessor, but it should also be usable standalone. If we didn't have an explicit "pipeline" value we could just say that video-depth-anything by default is always usable standalone.

yondonfu: Could this be an enum instead since we know the exact values it can take on?

src/scope/server/pipeline_processor.py
```diff
@@ -0,0 +1,485 @@
...
+    ):
+        """Initialize a pipeline processor.
+
+        Args:
+            pipeline_manager: Pipeline manager for loading pipelines
+            pipeline_id: ID of the pipeline to run
+            initial_parameters: Initial parameters for the pipeline
+            track_input_fps: Whether to track input FPS (for first pipeline in chain)
+        """
+        self.pipeline_manager = pipeline_manager
```
yondonfu: Seems like it would be cleaner to pass in a pipeline instance instead of a PipelineManager because a) the PipelineManager is only used once in the constructor anyway and b) this would create a better separation of concerns eg. no need for a PipelineProcessor to be aware of a PipelineManager and it only needs to be aware of the pipeline that it is associated with.

src/scope/server/pipeline_processor.py
```diff
@@ -0,0 +1,485 @@
...
+            while not old_queue.empty():
+                try:
+                    frame = old_queue.get_nowait()
+                    self.output_queue.put_nowait(frame)
+                except queue.Empty:
+                    break
+
+            # Update next processor's input_queue to point to the new output_queue
+            if self.next_processor is not None:
+                self.next_processor.input_queue = self.output_queue
```
yondonfu: I think there might be a race condition here if the thread for `self.next_processor` happens to be blocking on `self.input_queue.get(timeout=0.1)` thought I'm not sure of the practical impact. If there is a way to ensure safe reference swapping could still be worthwhile just to make the code easier to manage and more worry free. If not then a comment would suffice.

src/scope/server/pipeline_processor.py
```diff
@@ -0,0 +1,485 @@
...
+        # Calculate output queue size based on next processor's requirements
+        next_pipeline = next_processor.pipeline
+        if hasattr(next_pipeline, "prepare"):
+            requirements = next_pipeline.prepare(video=True)
+            input_size = requirements.input_size
+            target_size = max(8, input_size * OUTPUT_QUEUE_MAX_SIZE_FACTOR)
+            self._resize_output_queue(target_size)
+
+        # Update next processor's input_queue to point to this output_queue
+        next_processor.input_queue = self.output_queue
```
yondonfu: Echoing other comment in `_resize_output_queue()`.

src/scope/server/pipeline_manager.py
```diff
@@ -123,7 +304,10 @@ async def load_pipeline(
         self, pipeline_id: str | None = None, load_params: dict | None = None
     ) -> bool:
         """
-        Load a pipeline asynchronously.
+        Load a pipeline asynchronously (legacy method for backward compatibility).
```
yondonfu: If this is not used anywhere anymore let's remove?

src/scope/server/frame_processor.py
```diff
@@ -195,19 +181,47 @@ def put(self, frame: VideoFrame) -> bool:
...
+            # Put frame into first processor's input queue
+            try:
+                first_processor.input_queue.put_nowait(frame_tensor)
+            except queue.Full:
+                # Queue full, drop frame (non-blocking)
+                logger.debug("First processor input queue full, dropping frame")
+                return False
+        else:
+            # No processors yet, buffer the frame
+            with self.frame_buffer_lock:
```
yondonfu: It doesn't look like this frame_buffer is drained anymore so if we buffer frames before processors are set we'd just lose them? Can the frame_buffer be removed altogether?

yondonfu: I also see that the prepare_chunk() function was removed which previously included a bunch of frame dropping/sampling logic that relied on frame_buffer and now if we're just relying on the input queue of processors then if the queue fills up we'd just drop the newest frames. Was this intentional? This seems like a non-trivial behavioral change. Seems like considering such a change whether it makes sense or not should be outside of the scope of this PR.

src/scope/core/pipelines/base_schema.py
leszko: Ahh, I thought that we want to use "VideoDepthAntyhing" only as a preprocessor, not as a standalone pipeline. Do you think it makes any sense to have it as a standalone pipieline?

To your question, we could remove it and have EVERY pipeline being listed in the pipelines, but I wonder if all pipelines make sense standalone. E.g. the next pipeline I plan to add is RIFE (postprocessor for frame interpolation)m does it make sense to run standalone? It will just be another passthrough pipeline. Good for testing, but then our UI will become noisy with pipelines that don't make any actual usage.

frontend/src/components/SettingsPanel.tsx
leszko: Updated.

One more thing that came to my mind is that maybe we should actually take "Preprocessor" out of VACE in the context that we plan to add Postprocessors (which won't be related to VACE). @yondonfu wdyt?

src/scope/server/pipeline_processor.py
leszko: Added a lock for the queue swap. Should not introduce any performance drop, because it's called from multiple threads only when change the queue size.

src/scope/server/frame_processor.py
leszko: > It doesn't look like this frame_buffer is drained anymore so if we buffer frames before processors are set we'd just lose them? Can the frame_buffer be removed altogether?

Actually, I think we don't need `self.frame_buffer` anymore, because we should always have pipeline processors and the for queue of the first processor should act as a buffer. And, yes if there are no preprocessors setup, we lose the frames, but it should not happen, because we first set up preprocessors and then start receiving frames. 

Removed `frame_buffer`.

leszko: > I also see that the prepare_chunk() function was removed which previously included a bunch of frame dropping/sampling logic that relied on frame_buffer and now if we're just relying on the input queue of processors then if the queue fills up we'd just drop the newest frames. Was this intentional? This seems like a non-trivial behavioral change. Seems like considering such a change whether it makes sense or not should be outside of the scope of this PR.

You're right. Restored `prepare_chunk()`.

frontend/src/components/SettingsPanel.tsx
yondonfu: Yeah that makes sense to me.

src/scope/core/pipelines/base_schema.py
yondonfu: I think it makes sense for it to be usable standalone because real-time depth map output can be useful on its own (used via the API), using it in the UI can help with testing. We can always update the UI to better filter, categorize so that it doesn't get noisy. My thought here is that at a config level it make sense to assume that every pipeline can also be used standalone.

yondonfu:
>  (Likely higher in practice; this seems limited by my remote setup (I also see ~15 FPS for passthrough)).

I think the input videos from the UI are still hardcoded with max FPS of 15 so that may also cause this.

frontend/src/components/SettingsPanel.tsx
leszko: Updated

src/scope/core/pipelines/base_schema.py
leszko: Ok, updated. So every pipeline is available as standalone. And if you add `usage = [UsageType.PREPROCESSOR]`, then it's _also_ available as a preprocessor.

yondonfu:
```
2026-01-14 08:05:34,283 - scope.server.pipeline_processor - ERROR - Error processing chunk for video-depth-anything: all input arrays must have the same shape
Traceback (most recent call last):
  File "C:\Users\yondo\Development\scope\src\scope\server\pipeline_processor.py", line 426, in process_chunk
    output = self.pipeline(**call_params)
  File "C:\Users\yondo\Development\scope\src\scope\core\pipelines\video_depth_anything\pipeline.py", line 107, in __call__
    frames = np.stack(frames, axis=0)
  File "C:\Users\yondo\Development\scope\.venv\lib\site-packages\numpy\_core\shape_base.py", line 460, in stack
    raise ValueError('all input arrays must have the same shape')
ValueError: all input arrays must have the same shape
```

I see this error in the logs when I use the attached input video.

- VACE + Use Input Video set to True
- H = 320, W = 576

This seems to happen on a some chunks and not on others. I suspect because there is no chunk preprocessing for the video-depth-anything pipeline like there is for the other diffusion pipelines. The chunk preprocessing (I think in `preprocess_chunk()`) ensures that all the input frames are the expected HxW before the pipeline starts handling them

https://github.com/user-attachments/assets/0e10ae66-5184-4835-b5cb-1ab9ee8dab32

src/scope/core/pipelines/video_depth_anything/pipeline.py
```diff
@@ -96,14 +96,32 @@ def __call__(self, **kwargs) -> torch.Tensor:
...
             # Squeeze T dimension: (1, H, W, C) -> (H, W, C)
-            frames.append(frame_np.squeeze(0))
+            frame_np = frame_np.squeeze(0)
+            # Use first frame's shape as target for consistency
+            if target_shape is None:
+                target_shape = frame_np.shape
+            elif frame_np.shape != target_shape:
+                # Resize frame to match target shape using torch interpolate
+                frame_tensor = torch.from_numpy(frame_np).permute(2, 0, 1).unsqueeze(0)
+                frame_tensor = torch.nn.functional.interpolate(
```
yondonfu, line 114: There is a bunch of additional overhead here:

- The resizing happens on CPU instead of GPU
- Conversion from numpy -> tensor -> numpy only to convert back to tensor within the conditional on L143

Looks like its tricky to avoid this since [infer_video_depth_one()](https://github.com/DepthAnything/Video-Depth-Anything/blob/4f5ae23172ba60fd7bc11ef671cca678842c7072/video_depth_anything/video_depth_stream.py#L77) accepts numpy arrays...

I think the ideal solution is to create a different wrapper function around `forward_features()` and `forward_depth()` that can either handle this resizing or accept tensor(s) so that the caller can move data to GPU once take care of resizing and then pass it to the function. 

Since this would require non-trivial changes, WDYT about tracking this in an issue and handling separately?

PR #317, opened 2026-01-08
feat: auto-generate spatial masks for FFLF extension mode

ryanontheinside:
When reference images require padding to fit target resolution, spatial masks now indicate padding regions (1=generate) vs image regions (0=preserve). This allows the model to freely generate in padded areas while preserving the first/last frame influence only where the actual image content exists.

User provided masks to follow.


yondonfu (approved):
LGTM

ryanontheinside:
> 2026-01-08.17-45-33.mp4 
> I still see what looks like slivers of white canvas on the sides here when I use the following ref image:

this is an edge case - the reference image is 826×481. I will update the handle this

yondonfu (approved):
LGTM

PR #318, opened 2026-01-08
Fix KV cache handling and cache VACE

ryanontheinside:
Two main fixes for Krea VACE 14b integration:

  1. Restore KV cache handling when block_mask is set - Restore condition: `if kv_cache is None or block_mask is not None:` - Restore KV cache population with guard for when kv_cache exists - Fixes RecomputeKVCacheBlock which sets block_mask for full attention but still needs to populate the cache for subsequent chunks - VACE continues to work as it explicitly passes kv_cache=None

  2. Cache VACE context tokens, saves a small amount of compute?


yondonfu (approved):
LGTM!

PR #319, opened 2026-01-08
feat: Plugin artifact declarations

yondonfu:
## Summary

- Plugins can now declare required artifacts via the `artifacts` ClassVar on their pipeline config
- Artifacts are automatically discovered when pipelines are registered
- `requires_models` is auto-inferred from artifacts (no need to set both)

## Changes

- Add `artifacts: ClassVar[list[Artifact]]` to `BasePipelineConfig`
- Create `artifact_registry.py` with unified `get_artifacts_for_pipeline()` that checks config first, falls back to legacy `PIPELINE_ARTIFACTS`
- Update `models_config.py` and `download_models.py` to use the new registry
- Auto-infer `requires_models` from non-empty artifacts in schema metadata

## Plugin Usage

```python
from scope.core.pipelines.base_schema import BasePipelineConfig
from scope.server.artifacts import HuggingfaceRepoArtifact

class MyPluginConfig(BasePipelineConfig):
    pipeline_id = "my-plugin"
    artifacts = [
        HuggingfaceRepoArtifact(
            repo_id="my-org/my-model",
            files=["model.pt"],
        ),
    ]
```

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/scope/core/pipelines/base_schema.py
+++ b/src/scope/core/pipelines/base_schema.py
@@ -286,7 +286,8 @@ class BasePipelineConfig(BaseModel):
         )
         metadata["docs_url"] = cls.docs_url
         metadata["estimated_vram_gb"] = cls.estimated_vram_gb
-        metadata["requires_models"] = cls.requires_models
+        # Infer requires_models from artifacts if not explicitly set
+        metadata["requires_models"] = cls.requires_models or bool(cls.artifacts)
         metadata["supports_lora"] = cls.supports_lora
         metadata["supports_vace"] = cls.supports_vace
         metadata["supports_cache_management"] = cls.supports_cache_management


src/scope/server/artifact_registry.py
```diff
@@ -0,0 +1,39 @@
...
+    if pipeline_class is not None:
+        config_class = pipeline_class.get_config_class()
+        config_artifacts = getattr(config_class, "artifacts", [])
+        if config_artifacts:
+            return config_artifacts
+
+    # Fall back to legacy PIPELINE_ARTIFACTS dict
+    from .pipeline_artifacts import PIPELINE_ARTIFACTS
+
+    return PIPELINE_ARTIFACTS.get(pipeline_id, [])
```
leszko, line 39: Could we just instead of having this fallback, remove the `PIPELINE_ARTIFACTS` constant at all and move the specific model weight files into the pipelines.

For example, instead of having this one:
```
    "streamdiffusionv2": [
        WAN_1_3B_ARTIFACT,
        UMT5_ENCODER_ARTIFACT,
        VACE_ARTIFACT,
        LIGHTVAE_ARTIFACT,
        TAE_ARTIFACT,
        LIGHTTAE_ARTIFACT,
        HuggingfaceRepoArtifact(
            repo_id="jerryfeng/StreamDiffusionV2",
            files=["wan_causal_dmd_v2v/model.pt"],
        ),
    ],
```

We just move it into `pipelines/streamdiffusionv2/registry.py`?

yondonfu: Yeah will make this refactor in a follow up!

yondonfu: https://github.com/daydreamlive/scope/pull/327

PR #323, opened 2026-01-09
feat: add API support for first-frame-last-frame (FFLF) extension mode

ryanontheinside:
Expose first_frame_image and last_frame_image parameters in the API schema to enable FFLF extension mode for all VACE-enabled pipelines. These parameters allow users to set reference frames at the start and/or end of chunks:
- firstframe mode: reference at start, generate continuation
- lastframe mode: generate lead-up, reference at end
- firstlastframe mode: references at both ends

The parameters are automatically cleared after use to prevent unintended persistence across generation chunks.

### DIFF
--- a/src/scope/server/pipeline_processor.py
+++ b/src/scope/server/pipeline_processor.py
@@ -449,16 +449,17 @@ class PipelineProcessor:
 
             output = self.pipeline(**call_params)
 
-            # Clear vace_ref_images from parameters after use to prevent sending them on subsequent chunks
-            # vace_ref_images should only be sent when explicitly provided in parameter updates
-            if (
-                "vace_ref_images" in call_params
-                and "vace_ref_images" in self.parameters
-            ):
-                self.parameters.pop("vace_ref_images", None)
-
-            if "images" in call_params and "images" in self.parameters:
-                self.parameters.pop("images", None)
+            # Clear one-shot parameters after use to prevent sending them on subsequent chunks
+            # These parameters should only be sent when explicitly provided in parameter updates
+            one_shot_params = [
+                "vace_ref_images",
+                "images",
+                "first_frame_image",
+                "last_frame_image",
+            ]
+            for param in one_shot_params:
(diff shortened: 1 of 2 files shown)


src/scope/server/frame_processor.py
```diff
@@ -731,13 +731,16 @@ def process_chunk(self):
...
-            ):
-                self.parameters.pop("vace_ref_images", None)
+            # Clear one-shot parameters after use to prevent sending them on subsequent chunks
+            # These parameters should only be sent when explicitly provided in parameter updates
+            one_shot_params = [
+                "vace_ref_images",
+                "first_frame_image",
+                "last_frame_image",
+            ]
+            for param in one_shot_params:
```
yondonfu: This logic is in [PipelineProcessor](https://github.com/daydreamlive/scope/blob/9de8f1ee4b77ad942f45b7e17e35ae0ee54980ce/src/scope/server/pipeline_processor.py#L452) now after the recent preprocess PR merge.

When you rebase off `main` just make sure that this change is made in that class/file. Also, the logic for the `images` param (a temp thing prior to implementing the dynamic param rendering system) [here](https://github.com/daydreamlive/scope/blob/9de8f1ee4b77ad942f45b7e17e35ae0ee54980ce/src/scope/server/pipeline_processor.py#L460) can be folded in here too.

src/scope/core/pipelines/wan2_1/vae/wan.py
```diff
@@ -204,4 +204,4 @@ def decode_to_pixel(
 
     def clear_cache(self):
         """Clear encoder/decoder cache for next sequence."""
-        self.model.first_batch = True
+        self.model.clear_cache()
```
yondonfu: Can you explain the rationale for this change? Wouldn't this mean that we wouldn't trigger any of the first_batch = True logic for the VAE module after this is called?

ryanontheinside: this is a bug, reverted

PR #324, opened 2026-01-09
feat: add UI support for first-frame-last-frame (FFLF) extension mode

ryanontheinside:
Add frontend controls for FFLF in the Settings panel under the VACE section:

Users can now select reference images for the start and/or end of chunks directly from the UI when VACE is enabled.

### DIFF
--- a/frontend/src/components/InputAndControlsPanel.tsx
+++ b/frontend/src/components/InputAndControlsPanel.tsx
@@ -14,7 +14,7 @@ import { Upload, ArrowUp } from "lucide-react";
 import { LabelWithTooltip } from "./ui/label-with-tooltip";
 import type { VideoSourceMode } from "../hooks/useVideoSource";
 import type { PromptItem, PromptTransition } from "../lib/api";
-import type { InputMode, PipelineInfo } from "../types";
+import type { ExtensionMode, InputMode, PipelineInfo } from "../types";
 import { PromptInput } from "./PromptInput";
 import { TimelinePromptEditor } from "./TimelinePromptEditor";
 import type { TimelinePrompt } from "./PromptTimeline";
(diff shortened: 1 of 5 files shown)


frontend/src/components/InputAndControlsPanel.tsx
```diff
@@ -326,6 +341,92 @@ export function InputAndControlsPanel({
...
+                  maxImages={1}
+                  hideLabel
+                />
+              </div>
+            </div>
+            {(firstFrameImage || lastFrameImage) && (
+              <div className="space-y-2 mt-2">
+                <div className="flex items-center justify-between gap-2">
+                  <span className="text-xs text-muted-foreground">Mode:</span>
+                  <Select
```
yondonfu, line 401: Is there a benefit in being able to set the mode explicitly? Can we just allow the first and/or last frame to be set and the mode is just implicit, which removes the need for this extra Select, based on whether one or both are set?

frontend/src/components/InputAndControlsPanel.tsx
```diff
@@ -326,6 +341,92 @@ export function InputAndControlsPanel({
...
+                    onFirstFrameImageChange?.(images[0] || undefined);
+                  }}
+                  disabled={isDownloading}
+                  maxImages={1}
+                  hideLabel
+                />
+              </div>
+              <div className="space-y-1">
+                <span className="text-xs text-muted-foreground">Last Frame</span>
+                <ImageManager
```
yondonfu, line 385: Should pass label as something like "Extension Frames" or else the alt text (when image cannot be shown) for ImageManager will default to "Reference Images'.

frontend/src/components/InputAndControlsPanel.tsx
ryanontheinside: this enables sending both first and last frame, or just one or the other. it is implicit on the back end as discussed.

yondonfu: Upon further thought I see why the Select is here. Without it if you set both first and last frame but then you want to only send first frame you'd have to remove last frame (applies vice versa too) which would be slightly annoying. This Select lets you set both and leave them and only change first/last frame if you want the contents of either to change.

This feels reasonable for now. Leaving further UX thought on best way to configure FFLF to be considered separately.

PR #330, opened 2026-01-12
Add RIFE pipeline

leszko:
The RIFE pipeline is vendored from https://github.com/hzwer/Practical-RIFE.

**Important Notes**
1. The RIFE model weights in Practical-RIFE are distributed via Google Drive (packaged as a ZIP file). To support this, I added `GoogleDriveArtifact`.
2. While we could avoid Google Drive by sourcing weights from Hugging Face, there are some drawbacks:
   - I encountered noticeable shaking/blurring artifacts with all Hugging Face-hosted weights I tested. This is likely fixable, but investigating it further would take additional time.
   - Supporting Google Drive is probably valuable long-term anyway. Many researchers still publish model weights exclusively on Google Drive (e.g., our initial StreamDiffusion v2 setup also relied on Google Drive).

### DIFF
--- a/src/scope/core/pipelines/process.py
+++ b/src/scope/core/pipelines/process.py
@@ -65,12 +65,10 @@ def normalize_frame_sizes(
         # Move to target device if specified (before resizing for efficiency)
         if device is not None:
             frame = frame.to(device=device)
+        frame = frame.to(dtype=output_dtype)
 
         h, w = frame.shape[1], frame.shape[2]
         if not _needs_resize(h, w, target_h, target_w):
-            # No resize needed, just convert dtype if specified
-            if dtype is not None:
-                frame = frame.to(dtype=dtype)
             normalized.append(frame)
         else:
             logger.debug(f"Resized frame {i} from {w}x{h} to {target_w}x{target_h}")


src/scope/core/pipelines/process.py
```diff
@@ -6,6 +6,47 @@
 logger = logging.getLogger(__name__)
 
 
+def normalize_frame_sizes(frames: list[torch.Tensor]) -> list[torch.Tensor]:
```
yondonfu: There seems to be duplication of this logic in preprocess_chunk. If there is a need to avoid doing the last normalization step in preprocess_chunk can we at least use this helper in there too so we use the same resizing logic everywhere.

src/scope/core/pipelines/rife/modules/IFNet_HDv3.py
```diff
@@ -0,0 +1,193 @@
+import torch
```
yondonfu, line 4: Let's add a Modified from... comment with link to file where code was ported from for proper attribution consistent to what is done in other built-in pipelines. Let's do this for all the ported files.

leszko: Updated

src/scope/core/pipelines/process.py
leszko: Updated

leszko:
> Will give it a test tm!

Ok, I've rebased `rafal/postprocessors` https://github.com/daydreamlive/scope/pull/331 if you prefer to also test it as a postprocessor.

src/scope/core/pipelines/process.py
```diff
@@ -61,28 +74,13 @@ def preprocess_chunk(
...
-
-        frame_resized = torch.nn.functional.interpolate(
-            frame,
-            size=(height, width),
-            mode="bilinear",
-            align_corners=False,
-        )
-
-        logger.debug(f"Resized frame from {W}x{H} to {width}x{height}")
+        if height is not None and width is not None:
```
yondonfu: IIUC the resizing logic in `normalize_frame_sizes()` and here are essentially the same except the former uses the height & width of the first frame in the list as the target height & width and the latter uses an explicitly passed in target height & width?

If this is the case, could we simplify to this?

- `normalize_frame_sizes()` accepts an optional target_height and target_width
- If target_height and target_width are used used those values for resizing
- If target_height is not provided, use height of first frame as target
- If target_width is not provided, use width of first frame as target

And then do the resizing.

And then preprocess_chunk() just calls normalize_frame_sizes() internally?

leszko: ok, I think you meant something like this: https://github.com/daydreamlive/scope/pull/330/commits/97eaa77b20edbddec0867ce0feba16ca2ed42a51

I think it's fine. The only thing is that we need to use it before rearranging the tensors. But I think it's fine.

src/scope/core/pipelines/process.py
```diff
@@ -67,19 +74,14 @@ def preprocess_chunk(
     height: int | None = None,
     width: int | None = None,
 ) -> torch.Tensor:
-    frames = []
+    # Normalize frame sizes while still in THWC format
+    chunk = normalize_frame_sizes(chunk, target_height=height, target_width=width)
```
yondonfu: If this runs before L83 I think this means that if the chunk is on CPU that we'll do the resizing on CPU instead of on the target device (eg GPU if device = "cuda").

Maybe we can move the `.to()` call into the normalize call?

If the normalize call does everything then perhaps we don't need the loop starting on L80 and the only thing this function does is L88-90.

leszko: https://github.com/daydreamlive/scope/pull/330/commits/6b30eacc5f8a244916c0fa8e0a0511530e64b5c2

I think it looks better now. Shame on me, this took so many iterations 🤦

src/scope/core/pipelines/process.py
```diff
@@ -6,45 +6,93 @@
...
+    for i, frame in enumerate(frames):
+        # Move to target device if specified (before resizing for efficiency)
+        if device is not None:
+            frame = frame.to(device=device)
 
-        frames.append(frame_resized)
+        h, w = frame.shape[1], frame.shape[2]
+        if not _needs_resize(h, w, target_h, target_w):
+            # No resize needed, just convert dtype if specified
+            if dtype is not None:
```
yondonfu: 1. This conditional isn't needed if we use `output_dtype` instead which is never None
2. I think we want to always run `frame = frame.to(dtype=output_dtype)` (after moving the frame to `device`) because we always want frames to be on the same device and use the same dtype regardless of whether resizing is required.

leszko: https://github.com/daydreamlive/scope/pull/330/commits/275f8b1777eaccf7ec559b0e297ec03572057614

PR #332, opened 2026-01-12
Fix Krea

leszko:


BuffMcBigHuge:
Fixed in https://github.com/daydreamlive/scope/pull/311.

Issue #334, opened 2026-01-13
Krea does not work with default params on 5090


leszko:
### Summary

On 5090, the default UI params for Krea are:
- VACE => ON
- FP8 Quantization => ON

This results in the following error:
```
2026-01-13 13:38:48,530 - scope.server.pipeline_manager - ERROR - Failed to load pipeline krea-realtime-video: tuple index out of range. If this error persists, consider removing the models directory '/home/user/.daydream-scope/models' and re-downloading models.
```

My suggestion is to turn the VACE OFF by default on 5090. We already have a detection of the HW (to decide if FP8 is on or off depending on the VRAM). Then we can do the same for VACE, so that it's OFF by default on 5090.

### Platform

Ubuntu 20.04

### Nvidia GPU

5090

### Scope Version

main

### uv Version

0.1.0b2 (a6f393f)

PR #338, opened 2026-01-13
KREA Default Params 5090 Fix

BuffMcBigHuge:
This alleviates user frustration by defaulting VACE to "OFF" when on 5090 so KREA works with default params. It also handles FP8 when enabled while VACE is enabled.

Fix https://github.com/daydreamlive/scope/issues/334


leszko:
@BuffMcBigHuge Looks good. I updated it a little bit (to work for every pipeline, not only for krea). Plus, rebase to main and merged. Thanks for this work!

PR #339, opened 2026-01-14
Improve FPS calculation

leszko:
Simplify the FPS calculation. Now it will take into consideration only how fast the last pipeline processor produces frames. I think it's good this way and simpler. It will also work well with pre/post-processors.


src/scope/server/tracks.py
```diff
@@ -116,7 +116,7 @@ async def recv(self) -> VideoFrame:
             try:
                 # Update FPS: use minimum of input FPS and pipeline FPS
```
yondonfu: Not blocking for this PR, but can we update this comment? @leszko

leszko: Updated

PR #352, opened 2026-01-16
Add Dynamic Parameter Rendering

leszko:
Add support for defining parts of the frontend in plugins.

# Demo

https://github.com/user-attachments/assets/6ae9ea70-69d2-4205-8db7-70deff25b69e

# Sample plugin
It's simplest to understand how it works by looking at a sample plugin: https://github.com/leszko/scope-plugin-ui

1. Look at [schema.py](https://github.com/leszko/scope-plugin-ui/blob/main/scope_plugin_ui/schema.py), it introduces custom UI fields, e.g.
```python
    config_slider: int = Field(
        default=50,
        ge=0,
        le=100,
        description="Integer 0–100 in Settings",
        json_schema_extra=ui_field_config(
            order=3, is_load_param=True, label="Slider"
        ),
    )
```
2. Install the plugin with `DAYDREAM_SCOPE_PREVIEW=1 uv run daydream-scope install git+https://github.com/leszko/scope-plugin-ui`
3. Look the fields rendered in UI
image

# What and how to define
Any field in `schema.py` that includes `json_schema_extra` will be rendered in the UI.  
You can define the following keys inside `json_schema_extra` to control how the field is displayed:

- **`order`** *(int)* — Controls the rendering order in the UI (lower numbers appear first).
- **`category`** *(str)* — Determines where the field is displayed. Possible values:
  - `input`
  - `configuration` (default)
- **`component`** *(str)* — The name of a custom or complex UI component to use for this field.
- **`label`** *(str)* — The label shown in the UI. If not provided, the field’s `description` is used.
- **`modes`** *(list[str])* — Specifies which modes the field is visible in.  
  Defaults to `["text", "video"]`.

# Resources:
- Original RFC: [Notion](https://www.notion.so/livepeer/RFC-Dynamic-Parameter-Rendering-System-2e70a348568780ad9e8bfb09843973d8)
- Discord discussions: [thread](https://discord.com/channels/423160867534929930/1460779368539492433/1460779371768971345)

# Non-Goals
- Complex components can be improved, but I'll leave it for now, better to send a separate PR for that; the intention of this PR is to have support for primitive fields for plugins
- Complex components and their usage in SettingsPanel, I also leave it as a separate PR
- UI Schema JSON returned from the Server API endpoint (for now let's just use the same schema endpoint)

### DIFF
--- a/frontend/src/components/ComplexFields.tsx
+++ b/frontend/src/components/ComplexFields.tsx
@@ -185,7 +185,7 @@ export function SchemaComplexField({
             </div>
             <div className="flex items-center gap-2">
               <LabelWithTooltip
-                label="Scale:"
+                label="Scale"
                 tooltip="Scaling factor for VACE hint injection. Higher values make reference images more influential."
                 className="text-xs text-muted-foreground w-16"
               />
@@ -240,7 +240,7 @@ export function SchemaComplexField({
                 <LabelWithTooltip
                   label={PARAMETER_METADATA.height.label}
                   tooltip={PARAMETER_METADATA.height.tooltip}
-                  className="text-sm text-foreground w-14"
+                  className="text-sm font-medium w-14"
                 />
                 <div
                   className={`flex-1 flex items-center border rounded-full overflow-hidden h-8 ${ctx.heightError ? "border-red-500" : ""}`}
@@ -286,7 +286,7 @@ export function SchemaComplexField({
                 <LabelWithTooltip
                   label={PARAMETER_METADATA.width.label}
                   tooltip={PARAMETER_METADATA.width.tooltip}
-                  className="text-sm text-foreground w-14"
+                  className="text-sm font-medium w-14"
                 />
                 <div
                   className={`flex-1 flex items-center border rounded-full overflow-hidden h-8 ${ctx.widthError ? "border-red-500" : ""}`}
@@ -359,7 +359,7 @@ export function SchemaComplexField({
                 max={1.0}
                 step={0.01}
                 incrementAmount={0.01}
-                labelClassName="text-sm text-foreground w-20"
+                labelClassName="text-sm font-medium w-20"
                 valueFormatter={ctx.kvCacheAttentionBiasSlider.formatValue}
                 inputParser={v => parseFloat(v) || 1.0}
               />
@@ -368,7 +368,7 @@ export function SchemaComplexField({
               <LabelWithTooltip
                 label={PARAMETER_METADATA.manageCache.label}
                 tooltip={PARAMETER_METADATA.manageCache.tooltip}
-                className="text-sm text-foreground"
+                className="text-sm font-medium"
               />
               <Toggle
                 pressed={ctx.manageCache ?? true}
@@ -384,7 +384,7 @@ export function SchemaComplexField({
               <LabelWithTooltip
                 label={PARAMETER_METADATA.resetCache.label}
                 tooltip={PARAMETER_METADATA.resetCache.tooltip}
-                className="text-sm text-foreground"
+                className="text-sm font-medium"
               />
               <Button
                 type="button"
@@ -428,7 +428,7 @@ export function SchemaComplexField({
               <LabelWithTooltip
                 label={PARAMETER_METADATA.noiseController.label}
                 tooltip={PARAMETER_METADATA.noiseController.tooltip}
-                className="text-sm text-foreground"
+                className="text-sm font-medium"
               />
               <Toggle
                 pressed={ctx.noiseController ?? true}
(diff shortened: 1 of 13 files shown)


src/scope/core/pipelines/base_schema.py
```diff
@@ -96,6 +98,55 @@ def vace_context_scale_field(default: float = 1.0) -> FieldInfo:
...
+    If category is omitted, the frontend treats it as "configuration".
+
+    Args:
+        order: Display order (lower first). If omitted, Pydantic field order is used.
+        component: Complex component name ("vace", "lora", "denoising_steps",
+            "quantization", "cache", "image"). Use "image" for image-path fields
+            (str | None); the UI renders an image picker like first_frame_image.
+            Omit for primitive widgets.
+        modes: Restrict to input modes, e.g. ["video"]. Omit to show in all modes.
+        is_load_param: If True, this field is a load param (passed when loading the
```
yondonfu, line 124: Does this belong here? The distinction between runtime vs. load params seems like a data level distinction as opposed to a UI level distinction?

leszko: We need to decide if the UI elements are disabled when the stream is active. E.g. you cannot modify height/width when the stream is ongoing.

Some other options:
- Rename to `enabled_when_stream_active` or something similar to clarify the intention
- Do not allow load params at all to define by plugins. So, all plugin-defined UI elements are always active

yondonfu: Right that's why the UI needs to know if it is a load param. I'm wondering if it makes sense to just make that distinction outside of json_extra_schema since this is just being used for UI logic and whether something is a load vs runtime param affects whether it should be passed in the pipeline __init__ or the __call__ which is not just a UI concern. The immediate thought is having some way to mark a field as runtime vs load eg separate load/runtime config. I don't have a strong pref here so just making a note. If you think this can be considered separately no problem.

frontend/src/components/ComplexFields.tsx
```diff
@@ -0,0 +1,517 @@
...
+          loraMergeStrategy={ctx.loraMergeStrategy ?? "permanent_merge"}
+        />
+      </div>
+    );
+  }
+
+  if (component === "resolution") {
+    if (rendered.has("resolution")) return null;
+    rendered.add("resolution");
+    if (!ctx.supportsQuantization) return null;
```
yondonfu: Why does quantization matter here?

frontend/src/components/ComplexFields.tsx
```diff
@@ -0,0 +1,517 @@
...
+            </div>
+          </div>
+        </div>
+      </div>
+    );
+  }
+
+  if (component === "denoising_steps" && !rendered.has("denoising_steps")) {
+    rendered.add("denoising_steps");
+    if (!ctx.supportsQuantization) return null;
```
yondonfu: Why is this here?

frontend/src/components/InputAndControlsPanel.tsx
```diff
@@ -529,6 +547,91 @@ export function InputAndControlsPanel({
             );
           })()}
         </div>
+
+        {/* Schema-driven input fields (category "input"), below app-defined sections */}
+        {configSchema &&
```
yondonfu, line 552: Will we be able to in a separate PR remove the app-defined sections if this is here?

leszko:
> Doesn't need to happen immediately, but for my own understanding will we be able to slim down BasePipelineConfig too? It seems to have _a lot_ of stuff.

Yes, agree. The same actually with some parts of the frontend, like `StreamPage.tsx`, `schemaSettings.tsx`, and `SettingsPanel.tsx`. I have a _feeling_  there is too much happening there. But let's keep it as a separate PR. This is only have a lot of changes, I don't want to introduce additional refactor here.

frontend/src/components/ComplexFields.tsx
leszko: Removed

frontend/src/components/ComplexFields.tsx
leszko: Removed. Stupid claude.

frontend/src/components/InputAndControlsPanel.tsx
leszko: Yeah, let's maybe do it in separate refactoring PR, try to convert every part of the Input & Settings section into Complex Components. For now, it's ok, these code should be applied to each pipeline (though it may be hidden for some), so it's mainly about refactoring it into Complex Components.

src/scope/core/pipelines/base_schema.py
leszko: Let’s keep things as they are for now.

I agree that the backend could eventually use `is_load_param` to decide whether params should be passed to `__init__()` or `__call__()`. However, at this stage, I think the simplest and safest approach is to pass all params to both.

The main reason I don’t want to introduce that separation yet is that some params will likely apply to both cases, which requires more careful design. For now, filtering this in the UI feels like the best and least complex solution.

PR #359, opened 2026-01-20
feat: optical flow tensorrt

ryanontheinside:
This PR adds an optical flow preprocessor for use with VACE. Apart from usage of raft small, it follows the specifications of VACE.

Currently, torch.compile is used in favor of tensorrt. As things stand, optical flow will not be the bottleneck, tensorrt proved to be too problematic for something we do not currently need. This can be revisited down the line if needed.

### DIFF
--- a/src/scope/core/pipelines/optical_flow/pipeline.py
+++ b/src/scope/core/pipelines/optical_flow/pipeline.py
@@ -65,8 +65,10 @@ class OpticalFlowPipeline(Pipeline):
 
         self._pytorch_model = None
 
-        # Warmup: load and compile model to avoid delay on first frame
-        self._ensure_pytorch_model()
+        # Warmup: load model and run forward passes to trigger JIT compilation.
+        # This avoids delay on first real frame since torch.compile only
+        # triggers compilation on the first forward pass.
+        self._warmup()
 
     def _ensure_pytorch_model(self):
         """Lazily initialize the PyTorch RAFT model with torch.compile.
@@ -101,6 +103,28 @@ class OpticalFlowPipeline(Pipeline):
             f"Loaded and compiled RAFT {model_size} model in {time.time() - start:.3f}s"
         )
         return self._pytorch_model
+
+    def _warmup(self):
+        """Warmup the model by running forward passes to trigger JIT compilation."""
+        model = self._ensure_pytorch_model()
+
+        logger.info("Running warmup forward passes to trigger JIT compilation...")
+        start = time.time()
+
+        # Create dummy frames at the expected size (512x512 padded to multiple of 8)
+        dummy_frame = (
+            torch.randn(
+                1, 3, self._height, self._width, device=self.device, dtype=torch.float32
+            )
+            * 255.0
+        )
+
+        # Run a few forward passes to trigger and stabilize JIT compilation
+        with torch.no_grad():
+            for _ in range(3):
+                _ = model(dummy_frame, dummy_frame)
+
+        logger.info(f"Warmup completed in {time.time() - start:.3f}s")
 
     def _preprocess_frame_batch(
         self, frames: torch.Tensor


src/scope/core/pipelines/optical_flow/pipeline.py
```diff
@@ -0,0 +1,260 @@
...
+            device: Target device (defaults to CUDA if available)
+        """
+        from .engine import DEFAULT_HEIGHT, DEFAULT_WIDTH
+
+        self.device = (
+            device
+            if device is not None
+            else torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        )
+        self._height = DEFAULT_HEIGHT
```
yondonfu, line 57: Since we know the height & width WDYT about adding a few warmup calls at the end of  `__init__`? An issue right now is that the UI will indicate that loading & connection is complete, but then there is a blank screen for a few seconds because the first call is slower due to torch.compile and there aren't frames returned until the first call complete.

src/scope/core/pipelines/optical_flow/pipeline.py
```diff
@@ -0,0 +1,260 @@
...
+        # Remove padding
+        return flow[:, :orig_h, :orig_w]
+
+    def prepare(self, **kwargs) -> Requirements:
+        """Return pipeline requirements.
+
+        Returns:
+            Requirements specifying input_size needed for temporal consistency
+        """
+        return Requirements(input_size=4)
```
yondonfu: Why 4?

src/scope/core/pipelines/optical_flow/pipeline.py
```diff
@@ -0,0 +1,260 @@
...
+        self.device = (
+            device
+            if device is not None
+            else torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        )
+        self._height = DEFAULT_HEIGHT
+        self._width = DEFAULT_WIDTH
+
+        # Read settings from config
+        model_size = getattr(config, "model_size", "large")
```
yondonfu: The default is large here but small in the schema. Which should it be? And can we be consistent?

src/scope/core/pipelines/optical_flow/pipeline.py
```diff
@@ -0,0 +1,260 @@
...
+
+        # Pad to multiple of 8 if needed
+        pad_h = (8 - h % 8) % 8
+        pad_w = (8 - w % 8) % 8
+        if pad_h > 0 or pad_w > 0:
+            frames_scaled = F.pad(frames_scaled, [0, pad_w, 0, pad_h])
+
+        return frames_scaled, h, w
+
+    def _compute_optical_flow(
```
yondonfu, line 155: Do we need a `torch.no_grad()` decorator?

src/scope/core/pipelines/optical_flow/pipeline.py
```diff
@@ -0,0 +1,260 @@
...
+        )
+        self._height = DEFAULT_HEIGHT
+        self._width = DEFAULT_WIDTH
+
+        # Read settings from config
+        model_size = getattr(config, "model_size", "large")
+
+        # Model configuration
+        self._use_large_model = model_size == "large"
+        self._model_name = "raft_large" if self._use_large_model else "raft_small"
```
yondonfu: This is unused?

src/scope/core/pipelines/optical_flow/pipeline.py
```diff
@@ -0,0 +1,260 @@
...
+import torch
+import torch._inductor.config as inductor_config
+import torch.nn.functional as F
+from torchvision.utils import flow_to_image
+
+from ..interface import Pipeline, Requirements
+from .schema import OpticalFlowConfig
+
+# Disable CPU codegen to avoid needing MSVC on Windows
+inductor_config.disable_cpp_codegen = True
```
yondonfu: I think this has a global effect eg after this file is imported this always gets applied regardless of whether the pipeline is used? If so, any implications of having CPP codegen always be disabled?

src/scope/core/pipelines/optical_flow/pipeline.py
```diff
@@ -0,0 +1,260 @@
...
+        """
+        from .engine import DEFAULT_HEIGHT, DEFAULT_WIDTH
+
+        self.device = (
+            device
+            if device is not None
+            else torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        )
+        self._height = DEFAULT_HEIGHT
+        self._width = DEFAULT_WIDTH
```
yondonfu, line 58: IIUC we compute flow at 512x512 (resize input frames to this by default) and then resize to their original res. Is the reason for this that RAFT performs better and consumes less memory at 512x512 and the quality is good enough w.r.t to capturing motion and we don't care about being pixel perfect to the input frames?

src/scope/core/pipelines/optical_flow/compare_flow.py
```diff
@@ -0,0 +1,351 @@
...
+Loads a video, runs OpticalFlowPipeline with both backends, and saves
+side-by-side comparison outputs so quality differences can be visually inspected.
+
+Usage:
+    uv run python -m scope.core.pipelines.optical_flow.compare_flow
+"""
+
+from pathlib import Path
+
+import cv2
```
yondonfu: This won't work with as-is for a fresh install since `cv` is not a dep.

Let's either cleanup this file if we plan on maintaining it or remove if it was just for debugging?

ryanontheinside: will remove file, not needed

src/scope/core/pipelines/optical_flow/pipeline.py
ryanontheinside: Yea this is pretty typical, produces nearly identical results.

src/scope/core/pipelines/optical_flow/pipeline.py
ryanontheinside: Real VACE uses large, i think using small for real-time is reasonable. Will make consistent.

src/scope/core/pipelines/optical_flow/pipeline.py
ryanontheinside: Mistake, actually should be 2

src/scope/core/pipelines/optical_flow/pipeline.py
ryanontheinside: thanks for catching this, we can just set this flag when needed as opposed to on import of the file.

all of the other torch compiling we do is for gpu ops, so they should be unaffected.

his is only an issue for windows, so I added a platform condition to this flag. That being said, we could require MSVC to be installed to avoid usage of this flag entirely. 

WDYT?

src/scope/core/pipelines/optical_flow/pipeline.py
ryanontheinside: Done. I hadn't considered warmup calls could be used in this way/for this purpose, ty

src/scope/core/pipelines/optical_flow/pipeline.py
yondonfu: > we can just set this flag when needed as opposed to on import of the file

I think that is better. I suspect this change would still get picked up globally it just won't happen until the pipeline is actually loaded which is better than as soon as the file is imported.

I don't have a solution in mind to avoid this and I think it'd be better not to require MSVC right now as yet another thing you have to install so I'm comfortable leaving as is for now.

src/scope/core/pipelines/optical_flow/pipeline.py
```diff
@@ -0,0 +1,270 @@
...
+        # Read settings from config
+        model_size = getattr(config, "model_size", "small")
+
+        # Model configuration
+        self._use_large_model = model_size == "large"
+
+        self._pytorch_model = None
+
+        # Warmup: load and compile model to avoid delay on first frame
+        self._ensure_pytorch_model()
```
yondonfu: This just loads and runs torch.compile on the model but I don't think the JIT compilation will kick in until the first forward pass/call of the model? I'm assuming that to actually avoid triggering slow compilation when __call__ is invoked for the first time (eg by the PipelineProcessor) we'd need to trigger a few warmup calls here so the compilation happens here instead of later.

src/scope/core/pipelines/optical_flow/pipeline.py
ryanontheinside: ok cool, same page.

PR #362, opened 2026-01-20
feat: gray processor

ryanontheinside:
grayscale conditioning frames in accordance with VACE

### DIFF
--- a/src/scope/core/pipelines/gray/pipeline.py
+++ b/src/scope/core/pipelines/gray/pipeline.py
@@ -30,16 +30,12 @@ class GrayPipeline(Pipeline):
 
     def __init__(
         self,
-        config=None,
         device: torch.device | None = None,
-        dtype: torch.dtype = torch.float16,
     ):
         """Initialize the Gray pipeline.
 
         Args:
-            config: Pipeline configuration (unused — no model to configure)
             device: Target device (defaults to CUDA if available)
-            dtype: Data type (unused — this pipeline uses float32 tensor ops)
         """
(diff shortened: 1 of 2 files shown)


src/scope/core/pipelines/gray/pipeline.py
```diff
@@ -0,0 +1,83 @@
...
+            device: Target device (defaults to CUDA if available)
+            dtype: Data type (unused — this pipeline uses float32 tensor ops)
+        """
+        self.device = (
+            device
+            if device is not None
+            else torch.device("cuda" if torch.cuda.is_available() else "cpu")
+        )
+        self.rgb_weights = _RGB_WEIGHTS.to(device=self.device)
+        logger.info("Gray preprocessor initialized (no model required)")
```
yondonfu: Can we remove this to make logs less noisy since PipelineManager already logs something when this pipeline is initialized?

src/scope/core/pipelines/gray/pipeline.py
```diff
@@ -0,0 +1,83 @@
...
+class GrayPipeline(Pipeline):
+    """Grayscale conversion preprocessor optimized for realtime use."""
+
+    @classmethod
+    def get_config_class(cls) -> type["BasePipelineConfig"]:
+        return GrayConfig
+
+    def __init__(
+        self,
+        config=None,
```
yondonfu: This can prob be omitted if unused? I don't think it is a required param anywhere, but lmk if I am mistaken.

PR #368, opened 2026-01-21
feat: multi output processors

ryanontheinside:
## NOTE: rebase on main after #378 is merged

This is to support masking preprocessor plugins for VACE. 

An example plugin implementation: 
https://github.com/daydreamlive/scope_yolo_mask

Other more advanced masking techniques can be revisited later (SAM2, SAM3) using this same infrastructure. 

This change will enable inpainting with VACE, regional LoRA, background removal/preservation, and more.

### DIFF
--- a/src/scope/server/pipeline_processor.py
+++ b/src/scope/server/pipeline_processor.py
@@ -10,6 +10,7 @@ from typing import Any
 import torch
 
 from scope.core.pipelines.controller import parse_ctrl_input
+from scope.core.pipelines.wan2_1.vace import VACEEnabledPipeline
 
 from .pipeline_manager import PipelineNotAvailableException
 
@@ -84,6 +85,9 @@ class PipelineProcessor:
             "vace_use_input_video", True
         )
 
+        # Cache VACE support check to avoid isinstance on every chunk
+        self._pipeline_supports_vace = isinstance(pipeline, VACEEnabledPipeline)
+
         self.output_size = None
 
     def _resize_output_queue(self, target_size: int):
@@ -125,6 +129,7 @@ class PipelineProcessor:
         self.next_processor = next_processor
 
         # Calculate output queue size based on next processor's requirements
+        # Also pass chunk size to this processor so preprocessors can use it
         next_pipeline = next_processor.pipeline
         if hasattr(next_pipeline, "prepare"):
             requirements = next_pipeline.prepare(video=True)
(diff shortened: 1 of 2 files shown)


src/scope/server/pipeline_processor.py
```diff
@@ -339,6 +346,8 @@ def process_chunk(self):
         video_input = None
         if requirements is not None:
             current_chunk_size = requirements.input_size
+            if current_chunk_size is None:
+                return
```
leszko: What is the scenario in which you would like to return here? In general we had 2 scenarios:
1. Pipeline needs input video frames, then it specifies how many of them it needs with `input_size` and waits until enough frames are available
2. Pipeline does not need input video frames, then it returns `None` from Requirements and just process without any video frames

Now you add a scenario in which Pipeline needs `None` input frames, which sounds a little weird, but I think I just don't get what's the scenario here.

src/scope/server/pipeline_processor.py
```diff
@@ -375,26 +384,29 @@ def process_chunk(self):
...
-                    pipeline_supports_vace
+                    self._pipeline_supports_vace
                     and self.vace_enabled
                     and self.vace_use_input_video
                 ):
-                    # VACE conditioning: route to vace_input_frames
                     call_params["vace_input_frames"] = video_input
                 else:
-                    # Latent initialization: route to video
                     call_params["video"] = video_input
```
leszko, line 396: Shouldn't it be like this:
```python
if (
    "vace_input_frames" not in call_params
    and self._pipeline_supports_vace
    and self.vace_enabled
    and _vace_use_input_video
):
    # VACE conditioning: route to vace_input_frames
    call_params["vace_input_frames"] = video_input
else:
    # Latent initialization: route to video
```

src/scope/server/pipeline_processor.py
```diff
@@ -375,26 +384,29 @@ def process_chunk(self):
...
 
+            # Handle dict returns from preprocessors (dual-output pattern)
+            # Preprocessors return {"video": frames, ...extra_params}
+            if isinstance(output, dict):
+                extra_params = {k: v for k, v in output.items() if k != "video"}
+                if extra_params and self.next_processor is not None:
+                    self.next_processor.parameters.update(extra_params)
+                output = output.get("video")
+                if output is None:
+                    return
```
leszko: This makes me think that our pipeline interface is not good anymore. I've fired https://github.com/daydreamlive/scope/pull/370 and I'll open a discussion in the Discord thread.

ryanontheinside: perfect, same page

src/scope/server/pipeline_processor.py
ryanontheinside: i think this has a subtle bug - it would set call_params["video"] = video_input even when video_input is None or when vace_input_frames is already provided by a preprocessor. The current   
  logic is we only route video_input if it exists AND preprocessor hasn't already provided vace_input_frames.

src/scope/server/pipeline_processor.py
ryanontheinside: i think #370  will render this moot, but for the record i dont like it either

src/scope/server/pipeline_processor.py
```diff
@@ -123,12 +127,15 @@ def set_next_processor(self, next_processor: "PipelineProcessor"):
...
         # Calculate output queue size based on next processor's requirements
+        # Also pass chunk size to this processor so preprocessors can use it
         next_pipeline = next_processor.pipeline
         if hasattr(next_pipeline, "prepare"):
             requirements = next_pipeline.prepare(video=True)
             input_size = requirements.input_size
             target_size = max(8, input_size * OUTPUT_QUEUE_MAX_SIZE_FACTOR)
             self._resize_output_queue(target_size)
+            # Store downstream chunk size for preprocessors to use via prepare()
+            self.parameters["input_size"] = input_size
```
leszko: Just a small update, look at this: https://github.com/daydreamlive/scope/pull/378

It's the same logic, but works also for previous pipeline output size. Plus renamed it to "next_input_size"

src/scope/server/pipeline_processor.py
leszko: Ok #370 is merged.

But...one thing I realized is that we cannot just call `self.next_processor.parameters.update(extra_params)` here, because multiple threads should not modify the same map at the same time.

How about changing it to `self.next_processor.update_parameters()`?

ryanontheinside:
closed in favor of #391

PR #369, opened 2026-01-22
Add pipeline throttling

leszko:
## Summary
- Adds FPS-based throttling to prevent upstream pipelines from overwhelming downstream pipelines in chained setups (A → B → C)
- When an upstream pipeline produces frames faster than the downstream can consume, the throttler adds delays to match the consumption rate

## How it works
The `PipelineThrottler` tracks input/output FPS for each pipeline processor:
- Measures how fast each pipeline consumes frames (input FPS) and produces frames (output FPS)
- When output FPS exceeds 1.5× the next pipeline's input FPS, introduces a delay to slow down production
- Throttling only applies to pipelines with video input that are chained to a next processor

## Results
Tested with video-depth-anything + longlive + rife on 5090:
- **base (main):** 13.6-20.0 FPS
- **with throttling (this PR):** 17.0-20.4 FPS

### DIFF
--- a/src/scope/server/pipeline_throttler.py
+++ b/src/scope/server/pipeline_throttler.py
@@ -1,0 +1,188 @@
+"""Pipeline throttler for controlling frame processing rate in chained pipelines."""
+
+import logging
+import threading
+import time
+from collections import deque
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .pipeline_processor import PipelineProcessor
+
+logger = logging.getLogger(__name__)
+
+# Throttling constants
+FPS_SAMPLE_SIZE = 30
+FPS_MIN_SAMPLES = 5
+MIN_FPS = 1.0
+MAX_FPS = 120.0
+
+# Multiplier for target FPS when throttling
+# e.g., if next pipeline processes at 6 FPS, target ~9 FPS (1.5x)
+THROTTLE_TARGET_MULTIPLIER = 1.5
+
+
+class PipelineThrottler:
+    """Controls processing rate of a pipeline based on downstream pipeline performance.
+
+    When pipelines are chained (A -> B -> C), a faster upstream pipeline should not
+    produce frames much faster than the downstream pipeline can consume them.
+    This throttler measures the downstream pipeline's input processing rate and
+    adds appropriate delays to match it.
+
+    To disable throttling, simply don't create a throttler instance (set to None).
+    """
+
+    def __init__(self):
+        """Initialize the throttler."""
+        self._lock = threading.Lock()
+        self._stop_event = threading.Event()
+
+        # Track this pipeline's output FPS (how fast it produces frames)
+        self._output_times: deque[float] = deque(maxlen=FPS_SAMPLE_SIZE)
+        self._output_fps: float = MAX_FPS
+
+        # Track this pipeline's input FPS (how fast it consumes frames)
+        self._input_times: deque[float] = deque(maxlen=FPS_SAMPLE_SIZE)
+        self._input_fps: float = MAX_FPS
+
+        # Reference to next processor (set externally)
+        self._next_processor: PipelineProcessor | None = None
+
+    def set_next_processor(self, processor: "PipelineProcessor | None"):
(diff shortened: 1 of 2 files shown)


src/scope/server/pipeline_throttler.py
```diff
@@ -0,0 +1,213 @@
...
+        """Apply throttling by sleeping if necessary.
+
+        This should be called after processing a batch and before starting the next.
+        """
+        delay = self.calculate_delay()
+        if delay > 0:
+            logger.debug(
+                f"Throttling: sleeping {delay:.3f}s "
+                f"(output={self._output_fps:.1f}fps, "
+                f"next_input={self._get_next_input_fps():.1f}fps)"
```
yondonfu: `self._get_next_input_fps()` is called without holding a lock here but the comment on L202 says that it needs to be called with the lock held.

src/scope/server/pipeline_throttler.py
```diff
@@ -0,0 +1,213 @@
...
+    def throttle(self):
+        """Apply throttling by sleeping if necessary.
+
+        This should be called after processing a batch and before starting the next.
+        """
+        delay = self.calculate_delay()
+        if delay > 0:
+            logger.debug(
+                f"Throttling: sleeping {delay:.3f}s "
+                f"(output={self._output_fps:.1f}fps, "
```
yondonfu: Minor: `self._output_fps` is read without a lock which is inconsistent with the locking logic used elsewhere in this file.

src/scope/server/pipeline_throttler.py
```diff
@@ -0,0 +1,213 @@
...
+                f"next_input={self._get_next_input_fps():.1f}fps)"
+            )
+            time.sleep(delay)
```
yondonfu: Consider using a `threading.Event()` which can be waited upon and is interruptible unlike `time.sleep`.

eg

```
self._stop_event = threading.Event()

....
# throttle()
self._stop_event.wait(delay)
```

src/scope/server/pipeline_throttler.py
```diff
@@ -0,0 +1,213 @@
...
+        """Get the current input FPS (how fast this pipeline consumes frames)."""
+        with self._lock:
+            return self._input_fps
+
+    def get_output_fps(self) -> float:
+        """Get the current output FPS (how fast this pipeline produces frames)."""
+        with self._lock:
+            return self._output_fps
+
+    def should_throttle(self) -> bool:
```
yondonfu: Looks like dead code never used.

src/scope/server/pipeline_throttler.py
```diff
@@ -0,0 +1,213 @@
...
+            )
+            time.sleep(delay)
+
+    def _get_next_input_fps(self) -> float:
+        """Get next processor's input FPS. Must be called with lock held."""
+        if self._next_processor is None:
+            return MAX_FPS
+        return self._next_processor.throttler.get_input_fps()
+
+    def reset(self):
```
yondonfu: Doesn't look like this is used? I guess this might be relevant if pipelines in a PipelineProcessor could be swapped, but if we assume that a pipeline + PipelineProcessor are swapped out together then shouldn't need this right?

yondonfu (changes requested):
Looks pretty good! The main comment is about the locking in `throttle` and everything else is minor for your consideration.

src/scope/server/pipeline_throttler.py
leszko: Updated

src/scope/server/pipeline_throttler.py
leszko: Removed

src/scope/server/pipeline_throttler.py
leszko: Good idea, updated

src/scope/server/pipeline_throttler.py
leszko: Updated

src/scope/server/pipeline_throttler.py
leszko: Updated

PR #370, opened 2026-01-22
Update pipeline interface

leszko:
## Summary
After reviewing recent PRs ([#368](https://github.com/daydreamlive/scope/pull/368) and [#312](https://github.com/daydreamlive/scope/pull/312)), it’s clear the current `Pipeline` interface no longer matches real usage. This PR updates the interface to support existing and emerging pipeline patterns.

## Issues
1. `Pipeline.__call__()` accepts an `input` parameter that is unused by all pipelines.
2. `Pipeline.__call__()` returns a single `torch.Tensor`, which does not support:
   - Multi-output pipelines (e.g. separate mask outputs)
   - Pipelines that also return audio (e.g. LTX-2)
3. Pipeline requirements cannot adapt to frame counts produced by upstream pipelines.

## Changes
- Remove the unused `input` argument from `Pipeline.__call__()`
- Change pipeline outputs to return a `dict`
  - The main video output is available under the `"video"` key
- Add `auto_input_size` to pipeline requirements to support dynamic frame counts

## Compatibility
Plugins are not yet published, so this change is intentionally not backward-compatible. We could make it backward-compatible, but I think it's not worth it. Better to update all existing pipelines.

### DIFF
--- a/src/scope/core/pipelines/interface.py
+++ b/src/scope/core/pipelines/interface.py
@@ -3,7 +3,6 @@
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING
 
-import torch
 from pydantic import BaseModel
 
 if TYPE_CHECKING:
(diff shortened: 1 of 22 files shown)


src/scope/server/pipeline_processor.py
```diff
@@ -393,7 +393,10 @@ def process_chunk(self):
                     # Latent initialization: route to video
                     call_params["video"] = video_input
 
-            output = self.pipeline(**call_params)
+            output_dict = self.pipeline(**call_params)
```
ryanontheinside, line 396: should we extract other keys and forward them to handle processors that produce auxiliary outputs?

leszko: Yes, I think we should actually pass the whole dict into the next processor.

But... I'd prefer to do it in a separate PR to keep this PR dedicated to just the interface change.

One thing to think about is how to pass data between pipeline processors.
- For video, we currently extract individual frames and add then to the output_queue
- If we want to pass all outputs, that poses a question how to pass them:
  - One option is what you did [here](https://github.com/daydreamlive/scope/pull/368/files#r2716236477), so treat "video" the same as before and all other params just set as parameter
  - More correct and long-term option would be to queue the whole "dict", but then we would need to rethink how the queuing and waiting works, because the each pipeline can produce / consume different n of frames

src/scope/core/pipelines/interface.py
```diff
@@ -14,6 +13,8 @@ class Requirements(BaseModel):
     """Requirements for pipeline configuration."""
 
     input_size: int
+    # If True, the pipeline will automatically determine the input size based on the previous pipeline's output size.
+    auto_input_size: bool = False
```
yondonfu: How is this supposed to be used? Why can't a pipeline just always set its input_size independently?

leszko: Let's remove it from this PR, I'll discuss it separately.

PR #378, opened 2026-01-26
Propagate prev_output_size and next_input_size parameters

leszko:
A pipeline may want to decide on its input depending on the previous pipeline output size or the next pipeline input size.

I initially thought this would be the `auto_adjust` param in the Requirements, but not I think we should just propagate this param.

### DIFF
--- a/src/scope/core/pipelines/rife/pipeline.py
+++ b/src/scope/core/pipelines/rife/pipeline.py
@@ -55,7 +55,7 @@ class RIFEPipeline(Pipeline):
         logger.info("RIFE HDv3 model loaded successfully")
 
     def prepare(self, **kwargs) -> Requirements:
-        return Requirements(input_size=12)
+        return Requirements(input_size=kwargs.get("prev_output_size", 12))
 
     def __call__(
         self,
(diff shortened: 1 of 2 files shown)


src/scope/server/pipeline_processor.py
```diff
@@ -129,6 +131,8 @@ def set_next_processor(self, next_processor: "PipelineProcessor"):
             input_size = requirements.input_size
             target_size = max(8, input_size * OUTPUT_QUEUE_MAX_SIZE_FACTOR)
             self._resize_output_queue(target_size)
+            # Store downstream chunk size for preprocessors to use via prepare()
+            self.parameters["next_input_size"] = input_size
```
yondonfu, line 135: I'm a bit confused as to why downstream pipelines need to be aware of the input size required by upstream pipelines. What is the downside of the downstream pipeline independently setting its own required input size and then the PipelienProcessor buffering as is needed to make sure that it receives an input with the correct size?

src/scope/core/pipelines/rife/pipeline.py
```diff
@@ -55,7 +55,7 @@ def __init__(
         logger.info("RIFE HDv3 model loaded successfully")
 
     def prepare(self, **kwargs) -> Requirements:
-        return Requirements(input_size=12)
+        return Requirements(input_size=kwargs.get("prev_output_size", 12))
```
yondonfu, line 58: What is the downside of the pipeline just setting this independently?

PR #385, opened 2026-01-28
fix: frame processing

varshith15:
- adds memory buffers which speed up CPU -> GPU transfer and also making it non-blocking
- The streamdiffusion UI FPS goes up from 9.8 to 13.8 while the average test FPS being 16.29

### DIFF
--- a/src/scope/server/frame_processor.py
+++ b/src/scope/server/frame_processor.py
@@ -191,6 +191,9 @@ class FrameProcessor:
             if torch.cuda.is_available():
                 shape = frame_array.shape
                 pinned_buffer = self._get_or_create_pinned_buffer(shape)
+                # Note: We reuse pinned buffers for performance. This assumes the copy_()
+                # operation complete before the next frame arrives.
+                # In practice, copy_() is very fast (~microseconds) and frames arrive at 60 FPS at max
                 pinned_buffer.copy_(torch.as_tensor(frame_array, dtype=torch.uint8))
                 frame_tensor = pinned_buffer.cuda(non_blocking=True)
             else:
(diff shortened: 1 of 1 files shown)


src/scope/server/frame_processor.py
```diff
@@ -170,9 +186,16 @@ def put(self, frame: VideoFrame) -> bool:
...
 
-            # Convert VideoFrame to (H, W, C) uint8 tensor, then to [1, H, W, C]
             frame_array = frame.to_ndarray(format="rgb24")
-            frame_tensor = torch.from_numpy(frame_array)
+
+            if torch.cuda.is_available():
+                shape = frame_array.shape
+                pinned_buffer = self._get_or_create_pinned_buffer(shape)
+                pinned_buffer.copy_(torch.as_tensor(frame_array, dtype=torch.uint8))
+                frame_tensor = pinned_buffer.cuda(non_blocking=True)
```
yondonfu, line 198: Since this is async in theory I think it is possible for a frame to come in (ignoring who the caller of put() is) and causing an overwrite of what is in the buffer though in practice it seems unlikely given speed of copy vs. frame arrival. Lmk if that sounds right to you and if so can we just add a comment noting the assumption?

src/scope/server/frame_processor.py
```diff
@@ -494,7 +519,17 @@ def _spout_receiver_loop(self):
                     # Convert to tensor and put into first processor's input queue
                     if self.pipeline_processors:
                         first_processor = self.pipeline_processors[0]
-                        frame_tensor = torch.from_numpy(rgb_frame)
+
+                        if torch.cuda.is_available():
+                            shape = rgb_frame.shape
```
yondonfu, line 527: Since this is async in theory I think it is possible for a frame to come in (ignoring who the caller of put() is) and causing an overwrite of what is in the buffer though in practice it seems unlikely given speed of copy vs. frame arrival. Lmk if that sounds right to you and if so can we just add a comment noting the assumption?

yondonfu (changes requested):
Nice! Just a minor comment about a comment. And need to signoff the commits - can just do a `git rebase -i main --signoff`.

yondonfu (approved):
LGTM!

PR #388, opened 2026-01-28
feat: support all vace combinations

ryanontheinside:
This PR adds support for combinations of VACE features. This is backwards compatible, and assuming we add masking plugins and chaining preprocessors, it should be automagically supported by the front end.


src/scope/core/pipelines/longlive/test_vace.py
leszko:
```suggestion
        get_model_file_path("WanVideo_comfy/Wan2_1-VACE_module_1_3B_bf16.safetensors")
```

Not related to this PR, but maybe update here (and also in other VACE tests) the model weights to what we use in the Scope App.

src/scope/core/pipelines/longlive/test_vace.py
```diff
@@ -44,31 +48,35 @@
...
     "use_r2v": False,  # Reference-to-Video: condition on reference images
-    "use_depth": False,  # Depth guidance: structural control via depth maps
-    "use_inpainting": False,  # Inpainting: masked video-to-video generation
+    "use_depth": True,  # Depth guidance: structural control via depth maps
+    "use_inpainting": True,  # Inpainting: masked video-to-video generation
     "use_extension": True,  # Extension mode: temporal generation (firstframe/lastframe/firstlastframe)
     # ===== INPUT PATHS =====
     # R2V: List of reference image paths (condition entire video, don't appear in output)
     "ref_images": [
         "frontend/public/assets/example.png",  # path/to/image.png
```
leszko: Not related to this PR, but I think we don't have this file in the repo anymore. Could you add it? Otherwise the test fails.

src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
```diff
@@ -337,6 +341,25 @@ def _encode_extension_mode(
...
+                input_frames = input_frames[0]
+            conditioning_frames = input_frames.to(dtype=vae_dtype)
+
+            # Also get conditioning masks if provided
+            if block_state.vace_input_masks is not None:
+                input_masks = block_state.vace_input_masks
+                if input_masks.dim() == 5:
+                    # [B, 1, F, H, W] -> [1, F, H, W] (take first batch)
+                    input_masks = input_masks[0]
+                conditioning_masks = input_masks.to(dtype=vae_dtype)
```
leszko: I think you don't need to covert here to `vae_dtype`. The masks are anyway converted to `float32` in `_build_base_sequence`.

src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
```diff
@@ -337,6 +341,25 @@ def _encode_extension_mode(
...
+        # Check for conditioning inputs (inpainting/depth) to combine with extension
+        conditioning_frames = None
+        conditioning_masks = None
+        if block_state.vace_input_frames is not None:
+            # Extract conditioning frames for dummy positions
+            input_frames = block_state.vace_input_frames
+            if input_frames.dim() == 5:
+                # [B, C, F, H, W] -> [C, F, H, W] (take first batch)
+                input_frames = input_frames[0]
+            conditioning_frames = input_frames.to(dtype=vae_dtype)
```
leszko:
```suggestion
            conditioning_frames = input_frames.to(device=components.config.device, dtype=vae_dtype)
```
Just in case the input_frames are on CPU for some reason.

src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
leszko:
```suggestion
                conditioning_masks = input_masks.to(device=components.config.device)
```

Also you can move them to GPU, just in case they are in CPU.

src/scope/core/pipelines/longlive/test_vace.py
```diff
@@ -44,31 +48,35 @@
...
     ],
     # Depth: Path to depth map video (grayscale or RGB, will be converted)
-    "depth_video": "vace_tests/control_frames_depth.mp4",  # path/to/depth_video.mp4
+    "depth_video": "vace_tests/white_square_moving.mp4",  # path/to/depth_video.mp4
```
leszko: I think you need to add this mp4 file into tests.

src/scope/core/pipelines/longlive/test_vace.py
```diff
@@ -44,31 +48,35 @@
...
     # Inpainting: Input video and mask video paths
     "input_video": "frontend/public/assets/test.mp4",  # path/to/input_video.mp4
-    "mask_video": "vace_tests/circle_mask.mp4",  # path/to/mask_video.mp4
+    "mask_video": "vace_tests/static_mask_half_white_half_black.mp4",  # path/to/mask_video.mp4
```
leszko: I think you need to add this mp4 file into tests.

src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
```diff
@@ -397,127 +453,222 @@ def _build_extension_frames_and_masks(
...
+        Without inpainting:
+            - Replace frames with replicated reference
+            - Output mask: spatial_mask (for padding regions) or zeros (preserve all)
+        """
+        end_idx = start_idx + temporal_group_size
+        ref_replicated = ref_tensor.repeat(1, temporal_group_size, 1, 1)
+
+        # Clone to avoid modifying input tensors
+        frames = frames.clone()
+        masks = masks.clone()
```
leszko: I see this `_overlay_reference_at()` is called many times, so each time the tensors for `frames` and `masks` are cloned. Not sure how much overhead it adds, but maybe something worth to consider is if we really need to always clone it.

src/scope/core/pipelines/longlive/test_vace.py
ryanontheinside: I switched it to use the scope icon. For first frame last frame images, i changed to generic paths. We might want to ship with a few more assets in tests/fixtures, which i added to house the two mask videos.

src/scope/core/pipelines/longlive/test_vace.py
ryanontheinside: added tests/fixtures for things like this. we should add more.

src/scope/core/pipelines/longlive/test_vace.py
```diff
@@ -44,31 +48,35 @@
...
-    "input_video": "frontend/public/assets/test.mp4",  # path/to/input_video.mp4
-    "mask_video": "vace_tests/circle_mask.mp4",  # path/to/mask_video.mp4
+    "input_video": "frontend/public/assets/test.mp4",
+    "mask_video": "tests/fixtures/static_mask_half_white_half_black.mp4",
     # Extension: Frame images (appear in output video as actual frames)
-    "first_frame_image": "frontend/public/assets/woman1.jpg",  # For firstframe or firstlastframe modes
-    "last_frame_image": "frontend/public/assets/woman2.jpg",  # For lastframe or firstlastframe modes
-    "extension_mode": "firstlastframe",  # "firstframe", "lastframe", or "firstlastframe"
+    "first_frame_image": "path/to/first_frame.png",
+    "last_frame_image": "path/to/last_frame.png",
```
leszko: Could you also add the following to `tests/fixtures/`?
- first frame image
- last frame image
- reference image

Plus, update the paths here. Then someone (like me 🙃 ) could just clone the repo and run `vace_test.py`. Otherwise, poor claude needs to generate these images for me...

PR #397, opened 2026-02-04
Daydream Cloud: Remote GPU Processing in Scope

emranemran:
# Daydream Cloud: Remote GPU Processing in Scope

## Summary

This PR adds a **Cloud Mode** toggle to Scope, enabling users to run video generation pipelines on remote cloud GPUs (H100s for now) instead of requiring a local GPU.  Aside from Plugins and LoRAs, Local and Cloud mode are at feature parity.

### Local Mode (Current)
```
┌─────────────────────────────────────────────────────────────┐
│                      User's Machine                          │
│  ┌─────────────┐     WebRTC      ┌────────────────────────┐ │
│  │   Browser   │ ◄─────────────► │    Scope Backend       │ │
│  │  (Frontend) │                 │  (Local GPU - RTX/etc) │ │
│  └─────────────┘                 └────────────────────────┘ │
│                                            │                 │
│                                     ┌──────┴──────┐          │
│                                     │    Spout    │          │
│                                     │  (Windows)  │          │
│                                     └─────────────┘          │
└─────────────────────────────────────────────────────────────┘
```

### Cloud Mode

**When to use**: Desktop app where Spout I/O or local frame processing is needed.

**How to enable**: Start Scope, sign in, click "Connect to Cloud" in UI.

```
┌───────────────────────────────────────────────────────────────────────┐
│                           User's Machine                               │
│                                                                        │
│  ┌───────────┐        ┌─────────────────────────┐        ┌─────────┐  │
│  │  Browser  │ WebRTC │    Scope Backend        │ Spout  │ TouchD  │  │
│  │ (Frontend)│◄──────►│    (Client Mode)        │◄──────►│ Resolume│  │
│  └───────────┘        │                         │        │ etc.    │  │
│                       │  ┌───────────────────┐  │        └─────────┘  │
│                       │  │ CloudConnection   │  │                     │
│                       │  │ Manager           │  │                     │
│                       │  └────────┬──────────┘  │                     │
│                       └───────────┼─────────────┘                     │
└───────────────────────────────────┼───────────────────────────────────┘
                                    │
                          WebSocket │ + WebRTC
                          (signaling│ + video)
                                    │
                                    ▼
                       ┌─────────────────────────────────┐
                       │         Daydream Cloud            │
                       │  ┌───────────────────────────┐  │
                       │  │       fal_app.py          │  │
                       │  │  ┌─────────────────────┐  │  │
                       │  │  │   Scope Backend     │  │  │
                       │  │  │   (H100 GPU)        │  │  │
                       │  │  └─────────────────────┘  │  │
                       │  └───────────────────────────┘  │
                       └─────────────────────────────────┘

Flow:
1. User clicks "Connect to Cloud" in Settings panel
2. Local backend connects to Cloud backend via WebSocket
3. Local backend acts as WebRTC CLIENT to Cloud backend (role reversal!)
4. Frames from browser/Spout → local backend → Cloud → processed → back
5. Spout I/O works transparently (local backend handles it)
```

### Backend (scope-server)

| File | Purpose |
|------|---------|
| `cloud_connection.py` | WebSocket connection manager with request/response correlation |
| `cloud_webrtc_client.py` | WebRTC peer connection (client role) to fal.ai |
| `cloud_track.py` | MediaStreamTrack for bidirectional video relay |
| `frame_processor.py` | Modified to route frames to cloud when connected |
| `webrtc.py` | Extended to handle cloud mode offers |

### Cloud (fal.ai)

| File | Purpose |
|------|---------|
| `fal_app.py` | Serverless wrapper running Scope on fal.ai H100 |
| `FAL_DEPLOYMENT.md` | Deployment guide for fal.ai |

### Frontend

| File | Purpose |
|------|---------|
| `cloudAdapter.ts` | WebSocket-based API client for fal.ai |
| `cloudContext.tsx` | React context for cloud mode state |
| `useUnifiedWebRTC.ts` | Unified hook supporting both local and cloud modes |
| `useApi.ts` | API hook that routes to local or cloud |
| `CloudModeToggle.tsx` | UI component for enabling cloud mode |

## API Endpoints

New endpoints for cloud connection management:

```
POST /api/v1/cloud/connect    - Connect to fal.ai cloud
POST /api/v1/cloud/disconnect - Disconnect from cloud
GET  /api/v1/cloud/status     - Get connection status and stats
```

When connected to cloud, existing endpoints are automatically proxied:
- `POST /api/v1/pipeline/load` → proxied to cloud
- `GET /api/v1/pipeline/status` → proxied to cloud
- `GET /api/v1/pipelines/schemas` → proxied to cloud
- `POST /api/v1/webrtc/offer` → handled locally (relay mode)

### Core Functionality
- [x] Connect/disconnect to fal.ai cloud
- [x] WebRTC video relay through local backend
- [x] API call proxying via WebSocket
- [x] Spout input/output in cloud mode
- [x] Reference image upload to cloud
- [x] Recording download from cloud

## Authentication

Cloud mode requires Daydream authentication:
1. User signs in via OAuth flow
2. Frontend passes user ID to backend
3. Backend includes user ID in cloud connection
4. Events are tracked with user context in Kafka (only on Cloud backend - Local backend does not track anything)

### DIFF
--- a/src/scope/cloud/fal_app.py
+++ b/src/scope/cloud/fal_app.py
@@ -13,6 +13,7 @@ import asyncio
 import json
 import os
 import shutil
+import subprocess as _subprocess
 import time
 import uuid
 from typing import Any
@@ -32,8 +33,8 @@ async def validate_user_access(user_id: str) -> tuple[bool, str]:
     Returns (is_valid, reason) tuple.
     Access is granted if user is a cohort participant OR has @livepeer.org email.
     """
-    import urllib.request
     import urllib.error
+    import urllib.request
 
     if not user_id:
         return False, "No user ID provided"
@@ -210,7 +211,6 @@ def cleanup_session_data():
 # 5. fal deploy fal_app.py --auth public
 
 # Get git SHA at deploy time (this runs when the file is loaded during fal deploy)
-import subprocess as _subprocess
 
 
 def _get_git_sha() -> str:
@@ -740,7 +740,7 @@ class ScopeApp(fal.App, keep_alive=300):
                     message = await asyncio.wait_for(
                         ws.receive_text(), timeout=TIMEOUT_CHECK_INTERVAL_SECONDS
                     )
-                except (asyncio.TimeoutError, TimeoutError):
+                except (asyncio.TimeoutError, TimeoutError):  # noqa: UP041
                     if await check_max_duration_exceeded():
                         break
                     continue
--- a/src/scope/server/app.py
+++ b/src/scope/server/app.py
@@ -33,16 +33,16 @@ if TYPE_CHECKING:
 
 from .download_models import download_models
 from .download_progress_manager import download_progress_manager
-from .kafka_publisher import (
-    KafkaPublisher,
-    is_kafka_enabled,
-    set_kafka_publisher,
-)
 from .file_utils import (
     IMAGE_EXTENSIONS,
     LORA_EXTENSIONS,
     VIDEO_EXTENSIONS,
     iter_files,
+)
+from .kafka_publisher import (
+    KafkaPublisher,
+    is_kafka_enabled,
+    set_kafka_publisher,
 )
 from .logs_config import (
     cleanup_old_logs,


src/scope/core/pipelines/wan2_1/blocks/setup_caches.py
```diff
@@ -164,6 +164,7 @@ def __call__(self, components, state: PipelineState) -> tuple[Any, PipelineState
             components.config.vae_spatial_downsample_factor
             * components.config.patch_embedding_spatial_downsample_factor
         )
+
```
yondonfu: Stray can be removed.

src/scope/server/build.py
```diff
@@ -65,10 +65,10 @@ def main() -> None:
     # Check if build was successful
     dist_dir = frontend_dir / "dist"
     if not dist_dir.exists():
-        print("❌ Error: Frontend build failed - dist directory not found")
+        print("Error: Frontend build failed - dist directory not found")
```
yondonfu: If these changes are env specific eg for CI they shouldn't be necessary as the emojis can be left if proper env vars are set. See how the current GH actions do this https://github.com/daydreamlive/scope/blob/a1ddc9d718b5f284232140f8c6413192d860c97f/.github/workflows/test.yml#L43

frontend/vite.config.ts
```diff
@@ -10,6 +10,7 @@ export default defineConfig({
     },
   },
   server: {
+    host: true,
```
yondonfu: Shouldn't need this as a default because can just run `npm run dev -- --host` if network access is needed.

.github/workflows/docker-build-branch.yml
yondonfu, line 1: This looks pretty similar to the existing Docker build action. Can that one be used with whatever modifications are needed? In practice, I think once dev on this branch is done only need pushes of images when merging to `main`?

frontend/src/components/settings/DaydreamAccountSection.tsx
```diff
@@ -0,0 +1,306 @@
+/**
+ * DaydreamAccountSection - Auth and Cloud Mode UI for Settings
+ *
+ * Displays:
+ * - Not logged in: Sign in/Sign up buttons
```
yondonfu, line 5: Clean up comment needed no sign up button

frontend/src/components/settings/DaydreamAccountSection.tsx
```diff
@@ -0,0 +1,306 @@
...
+  };
+
+  const handleSignIn = () => {
+    redirectToSignIn();
+  };
+
+  const handleSignOut = () => {
+    // Disconnect from cloud if connected before signing out
+    if (status.connected) {
+      handleDisconnect();
```
yondonfu: `await` needed here since I think you want to finish disconnecting before proceeding?

frontend/src/components/Header.tsx
```diff
@@ -1,35 +1,84 @@
...
-    "general" | "api-keys" | "plugins"
+    "general" | "account" | "api-keys" | "plugins"
   >("general");
   const [initialPluginPath, setInitialPluginPath] = useState("");
+  const [cloudConnected, setCloudConnected] = useState(false);
+  const [cloudConnecting, setCloudConnecting] = useState(false);
+
+  // Fetch initial cloud status
+  useEffect(() => {
+    const fetchCloudStatus = async () => {
```
yondonfu: There seems to be redundant polling for cloud status both here and in DaydreamAccountSection. Can that be avoided eg by lifting polling to a single place and passing props so they can be used by both components or by using a context?

frontend/src/hooks/usePipeline.ts
```diff
@@ -10,6 +14,29 @@ interface UsePipelineOptions {
 
 export function usePipeline(options: UsePipelineOptions = {}) {
   const { pollInterval = 2000, maxTimeout = 600000 } = options;
+  const { adapter, isCloudMode } = useCloudContext();
+
+  // Helper functions that use cloud adapter when available
+  const getPipelineStatus =
```
yondonfu: Wouldn't the new useApi hook avoid repeating this cloud check stuff here?

frontend/src/hooks/usePipelines.ts
```diff
@@ -58,21 +57,85 @@ export function usePipelines() {
...
+  );
+
+  // Refresh pipelines (can be called externally)
+  const refreshPipelines = useCallback(async () => {
+    try {
+      setIsLoading(true);
+      setError(null);
+
+      // Use adapter if in cloud mode, otherwise direct API
+      const schemas =
```
yondonfu, line 72: Wouldn't the useApi hook avoid these cloud checks here?

frontend/src/hooks/useStreamState.ts
```diff
@@ -42,6 +43,25 @@ function getFallbackDefaults(mode?: InputMode) {
 }
 
 export function useStreamState() {
+  const { adapter, isCloudMode, isReady } = useCloudContext();
+
+  // Helper functions that use cloud adapter when available
+  const getPipelineSchemas =
+    useCallback(async (): Promise<PipelineSchemasResponse> => {
```
yondonfu, line 50: Wouldn't the useApi hook avoid the cloud checks here?

frontend/src/hooks/useUnifiedWebRTC.ts
```diff
@@ -0,0 +1,531 @@
...
+  onStreamStop?: () => void;
+}
+
+/**
+ * Unified WebRTC hook that works in both local and cloud modes.
+ *
+ * In local mode, uses direct HTTP for signaling.
+ * In cloud mode, uses the CloudAdapter WebSocket for signaling.
+ */
+export function useUnifiedWebRTC(options?: UseUnifiedWebRTCOptions) {
```
yondonfu, line 45: Could prob be unified with existing useWebRTC hook as most of that is duplicated here in a separate pass yeah?

frontend/src/lib/cloudContext.tsx
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * Provides a context for managing cloud deployment mode.
+ * When CLOUD_WS_URL is set, all API calls and WebRTC signaling
+ * go through the CloudAdapter WebSocket connection.
+ */
+
+import React, { createContext, useContext, useEffect, useState } from "react";
+import { CloudAdapter } from "./cloudAdapter";
+
+interface FalContextValue {
```
yondonfu: Is this fal specific?

frontend/src/lib/cloudContext.tsx
```diff
@@ -0,0 +1,97 @@
...
+        console.error("[CloudProvider] Connection failed:", err);
+        setError(err);
+        setIsReady(false);
+      });
+
+    return () => {
+      console.log("[CloudProvider] Disconnecting from cloud");
+      cloudAdapter.disconnect();
+    };
+  }, [wsUrl]);
```
yondonfu: Might want the apiKey here in dep array - at the moment if it changes adapter won't reconnect.

src/scope/cloud/fal_app.py
```diff
@@ -0,0 +1,725 @@
...
+    machine_type = "GPU-H100"
+
+    # Additional requirements needed for the setup code
+    requirements = [
+        "requests",
+        "httpx",  # For async HTTP requests
+        "aiokafka",  # For Kafka event publishing
+    ]
+
+    auth_mode = "public"
```
yondonfu, line 274: Just to clarify there is no auth impl here yet but it needs to be added yes?

src/scope/server/app.py
```diff
@@ -1324,6 +1713,120 @@ async def reload_plugin(
...
+    try:
+        # Use request body credentials if provided, otherwise fall back to CLI/env
+        app_id = request.app_id or os.environ.get("SCOPE_CLOUD_APP_ID")
+        api_key = request.api_key or os.environ.get("SCOPE_CLOUD_API_KEY")
+
+        if not app_id:
+            raise HTTPException(
+                status_code=400,
+                detail="cloud credentials not configured. Use --cloud-app-id and --cloud-api-key CLI args, "
+                "or SCOPE_CLOUD_APP_ID and SCOPE_CLOUD_APP_ID environment variables.",
```
yondonfu: Dup mention of same env var here

src/scope/server/cloud_track.py
```diff
@@ -0,0 +1,255 @@
...
+            await asyncio.sleep(0.01)
+
+        # No frame received, return last frame or black frame
+        if self._last_frame is not None:
+            pts, time_base = await self.next_timestamp()
+            self._last_frame.pts = pts
+            return self._last_frame
+
+        # Return black frame as fallback
+        black = np.zeros((512, 512, 3), dtype=np.uint8)
```
yondonfu: This seems to be the cause of the black frame issue we see at start of stream sometimes.

frontend/src/lib/auth.ts
```diff
@@ -0,0 +1,268 @@
...
+    window.location.href = authUrl;
+  }
+}
+
+/**
+ * Exchange a short-lived token for a long-lived API key
+ */
+export async function exchangeTokenForAPIKey(token: string): Promise<string> {
+  // TODO if this fails, log out and redirect to sign in page
+  // TODO check state queryparam
```
yondonfu: This still needs to be addressed yeah?

src/scope/server/frame_processor.py
```diff
@@ -33,14 +35,32 @@ def to_ndarray(self, format="rgb24"):
...
+    """
+
     def __init__(
         self,
-        pipeline_manager: PipelineManager,
+        pipeline_manager: "PipelineManager | None" = None,
         max_parameter_queue_size: int = 8,
         initial_parameters: dict = None,
         notification_callback: callable = None,
+        cloud_manager: "Any | None" = None,  # CloudConnectionManager for cloud mode
```
yondonfu: Consider using TYPE_CHECKING import for CloudConnectionManager as is done elsewhere. Example https://github.com/daydreamlive/scope/blob/a1ddc9d718b5f284232140f8c6413192d860c97f/src/scope/core/plugins/manager.py#L20

src/scope/server/pipeline_processor.py
```diff
@@ -381,29 +394,28 @@ def process_chunk(self):
...
 
             # Route video input based on VACE status
-            # Don't overwrite if preprocessor already provided vace_input_frames
-            if video_input is not None and "vace_input_frames" not in call_params:
+            # We do not support combining latent initialization and VACE conditioning
+            if video_input is not None:
+                # Check if pipeline actually supports VACE before routing to vace_input_frames
+                from scope.core.pipelines.wan2_1.vace import VACEEnabledPipeline
+
+                pipeline_supports_vace = isinstance(self.pipeline, VACEEnabledPipeline)
```
yondonfu: Can we keep what was previously here as that avoided doing a check per chunk. This looks like a regression

src/scope/server/pipeline_processor.py
```diff
@@ -471,30 +487,26 @@ def process_chunk(self):
...
-                self.output_frame_deltas.append(delta)
-
-            self._last_frame_time = now
+            self.output_frame_times.append(time.time())
 
         self._calculate_output_fps()
 
     def _calculate_output_fps(self):
-        """Calculate FPS from the average inter-frame delta."""
+        """Calculate FPS based on how fast frames are produced into the output queue."""
```
yondonfu: What is the reason for this change? Looks like a regression

src/scope/server/pipeline_processor.py
```diff
@@ -381,29 +394,28 @@ def process_chunk(self):
...
+                    # VACE conditioning: route to vace_input_frames
                     call_params["vace_input_frames"] = video_input
                 else:
+                    # Latent initialization: route to video
                     call_params["video"] = video_input
 
             output_dict = self.pipeline(**call_params)
 
             # Extract video from the returned dictionary
-            output = output_dict.get("video")
```
yondonfu, line 398: Why is this being removed this seems like a regression

frontend/src/lib/cloudAdapter.ts
```diff
@@ -0,0 +1,516 @@
...
+}
+
+// ==================== React Hook ====================
+
+import { useState, useEffect, useRef, useCallback } from "react";
+
+/**
+ * React hook for using the CloudAdapter
+ */
+export function useCloudAdapter(wsUrl: string | null, apiKey?: string) {
```
yondonfu, line 442: Please check given the existence of cloudContext whether this file can be cleaned up as there is a bunch of code here that seems like dup and not used.

src/scope/server/build.py
emranemran: I'll remove this. I think this was added when I tried to use WSL to compile and I forgot to remove.

.github/workflows/docker-build-branch.yml
emranemran: @mjh1 wdyt?

frontend/src/components/Header.tsx
emranemran: @mjh1 can you take a look at this?

src/scope/cloud/fal_app.py
emranemran: There's a bug on fal's end that's blocking the use of jwt token to auth. Until they fix the issue, we have to use public mode here. We'll update this once we hear from them. cc @mjh1 in case i missed anything here.

.github/workflows/docker-build-branch.yml
mjh1: Yea I agree it should be combined, I just had trouble modifying it in the past so did this temporarily really

frontend/src/lib/cloudContext.tsx
mjh1: Renamed to generic "cloud" 👍

src/scope/server/app.py
```diff
@@ -484,7 +575,31 @@ async def get_pipeline_schemas():
...
+    cloud-hosted scope backend to get the available pipelines there.
     """
+    # If connected to cloud, proxy the request to get cloud's available pipelines
+    if cloud_manager.is_connected:
+        logger.info("Proxying pipeline schemas request to cloud")
+        response = await cloud_manager.api_request(
+            method="GET",
+            path="/api/v1/pipelines/schemas",
+            timeout=30.0,
+        )
```
leszko, line 595: This somehow does not work for me without refreshing the page. Look at this recording. I'm running Scope on my local macos, but when I connect to the Cloud, I'd like to see ALL pipelines. However, without refreshing the page, I don't see, e.g., `longlive`.

https://github.com/user-attachments/assets/8f8c1f49-2957-4d1a-b162-0b372e426e7e

src/scope/server/app.py
```diff
@@ -591,12 +748,50 @@ async def download_recording(
...
     This will finalize the current recording and create a copy for download,
-    then continue recording with a new file."""
+    then continue recording with a new file.
+
+    In cloud mode, this proxies the download request to cloud.
+    """
     try:
-        # Get the session by ID
+        # If cloud mode is active, proxy to cloud
+        if cloud_manager.is_connected:
```
leszko, line 767: I see we use this `if cloud_manager.is_connected` everywhere with the same logic. Shouldn't we maybe refactor this to use `@app.middleware("http")` to just handle the same for all http requests? Or if we don't want to proxy ALL requests, then use a decorator?

src/scope/server/app.py
```diff
@@ -591,12 +748,50 @@ async def download_recording(
...
+
+            # Return as streaming response
+            return Response(
+                content=content,
+                media_type="video/mp4",
+                headers={
+                    "Content-Disposition": f'attachment; filename="{filename}"',
+                    "Content-Length": str(len(content)),
+                },
+            )
```
leszko, line 798: Why do we need this custom recording logic? Could we just forward/proxy the download request (the same way as we do with other requests)?

src/scope/server/app.py
```diff
@@ -724,6 +923,39 @@ def process_asset_file(
         )
 
     try:
+        # If cloud mode is active, proxy to cloud
+        if cloud_manager.is_connected:
+            logger.info("[CLOUD] Fetching assets list from cloud")
+            path = "/api/v1/assets"
```
leszko, line 932: Are assets stored in the backend "per user"? I think they should be or alternatively they should be cleaned up for each cloud session. We want to prevent one user seeing assets from another user.

src/scope/server/app.py
```diff
@@ -793,6 +1045,54 @@ async def upload_asset(request: Request, filename: str = Query(...)):
                 detail=f"File size exceeds maximum of {max_size / (1024 * 1024):.0f}MB",
             )
 
+        # If cloud mode is active, upload to cloud AND save locally for thumbnails
```
leszko, line 1041: Why do we need to save thumbnails locally?

src/scope/server/app.py
```diff
@@ -1324,6 +1713,120 @@ async def reload_plugin(
         ) from e
 
 
+# =============================================================================
+# Cloud Integration Endpoints
+# =============================================================================
+
+
+# TODO disconnect from cloud if browser is closed
```
leszko: Don't we actually need to implement it? Or at least some timeout for the Cloud connection? What if a user closes the browser and just keep it open, I think that will be a big cost for us.

src/scope/server/app.py
```diff
@@ -1324,6 +1713,120 @@ async def reload_plugin(
...
+    This establishes a WebSocket connection to the cloud runner,
+    which stays open until disconnect is called. Once connected:
+    - Pipeline loading is proxied to the cloud-hosted scope backend
+    - Video from the browser flows through the backend to cloud
+    - The cloud runner stays warm and ready for video processing
+
+    Credentials can be provided in the request body or via CLI args:
+    --cloud-app-id and --cloud-api-key (or SCOPE_CLOUD_APP_ID/SCOPE_CLOUD_API_KEY env vars).
+
+    Note: The connection may take 1-2 minutes on cold start while the
```
leszko, line 1730: Should we maybe mention that it can take so long in the frontend message when the connection is being established?

src/scope/server/frame_processor.py
```diff
@@ -75,10 +105,15 @@ def __init__(
         # This determines whether we wait for video frames or generate immediately.
         self._video_mode = (initial_parameters or {}).get("input_mode") == "video"
 
-        # Pipeline chaining support
+        # Pipeline chaining support (local mode only)
```
leszko, line 108: So, we don't support pre- and post- processors in the cloud mode?

src/scope/server/frame_processor.py
```diff
@@ -182,7 +299,28 @@ def put(self, frame: VideoFrame) -> bool:
...
+            self._log_frame_stats()
+
+        if self._cloud_mode:
+            # Cloud mode: send frame to cloud (only in video mode)
+            # In text mode, cloud generates video from prompts only - no input frames
+            if not self._video_mode:
+                return True  # Silently ignore frames in text mode
+            if self.cloud_manager:
+                frame_array = frame.to_ndarray(format="rgb24")
+                if self.cloud_manager.send_frame_to_fal(frame_array):
```
leszko: nit: maybe just `self.cloud_manager.send_frame(frame_array)`, I don't think we need to mention everywhere that it's send to FAL

frontend/src/lib/auth.ts
mjh1: ah yep, done now

src/scope/server/webrtc.py
```diff
@@ -317,6 +338,134 @@ def on_data_channel_message(message):
...
+                    exclude_none=True
+                )
+            logger.info(f"[CLOUD] Received offer with parameters: {initial_parameters}")
+
+            # Create new RTCPeerConnection with configuration
+            pc = RTCPeerConnection(self.rtc_config)
+            session = Session(pc, user_id=request.user_id)
+            self.sessions[session.id] = session
+
+            # Create CloudTrack instead of VideoProcessingTrack
```
leszko, line 373: Why do we actually need a separate class `CloudTrack`? I always thought we will be able to use `VideoProcessingTrack` and just handle put/get frame differently in `FrameProcessor`.

src/scope/server/cloud_track.py
```diff
@@ -0,0 +1,237 @@
...
+from aiortc.mediastreams import VIDEO_CLOCK_RATE, VIDEO_TIME_BASE, MediaStreamError
+from av import VideoFrame
+
+if TYPE_CHECKING:
+    from .cloud_connection import CloudConnectionManager
+
+logger = logging.getLogger(__name__)
+
+
+class CloudTrack(MediaStreamTrack):
```
leszko, line 33: It seems there is quite a lot of logic duplicated with `VidoeProcessingTrack`. Could we somehow re-use it or maybe extract somewhere the common logic?

src/scope/cloud/fal_app.py
```diff
@@ -0,0 +1,782 @@
...
+        return False, "User is not a cohort participant"
+    except urllib.error.HTTPError as e:
+        if e.code == 404:
+            return False, "User not found"
+        return False, f"Failed to fetch user: {e.code}"
+    except Exception as e:
+        return False, f"Error validating user: {e}"
+
+
+class KafkaPublisher:
```
leszko, line 67: Why do we have a second `KafkaPublisher` implementation here? Isn't the one from `kafka_publisher.py` the only one needed?

src/scope/server/app.py
mjh1: this was a regression somewhere, it's fixed that now 👍

frontend/src/components/Header.tsx
mjh1: Implemented this now 👍

frontend/src/hooks/useUnifiedWebRTC.ts
mjh1: Similar to `cloudAdapter.ts` these implementations were for the front end direct to fal integration for a hosted deployment. We decided to keep these in so that this feature could be used if needed. Maybe we should just forget about that and remove the direct from front end implementation, but perhaps we could do that in a follow up PR to avoid making big changes at this stage. cc @emranemran

src/scope/server/app.py
mjh1: Yes that's right, there is only one connection allowed per runner on the Fal side, and we clean up the directory after every connection.

src/scope/server/app.py
mjh1: I guess claude thought it might just be useful so that the user has them available when they go back to local mode too, but no need for it.

src/scope/server/app.py
mjh1: Yeah I don't think we should implement this since it's nice to still have the connection there if you refresh your page or it gets closed for some reason. And the overall timeout is already implemented yes, we give users a maximum connection duration of 1hr, they can then reconnect if needed.

src/scope/cloud/fal_app.py
mjh1: We couldn't get the import to work annoyingly, we'll follow up on this in another PR, i've raised a ticket.

src/scope/server/frame_processor.py
mjh1: It should be supported, I think I remember we tested it together so not sure what this comment is about, I'll take a look

src/scope/server/cloud_track.py
emranemran: That was intentional. We didn't want to modify the local code paths in any way for this Beta launch. Also, a lot of these code changes might go away for the real cloud version so it would be easier to revert at that point.

frontend/src/hooks/useUnifiedWebRTC.ts
emranemran: Yes, we can remove it as a follow-up but there are some product ideas that will be tested with direct mode so we should leave it as-is for now if that's alright. Removing it now might introduce breaking changes and we're tight on time.

src/scope/server/app.py
mjh1: I think that's another good one for a follow up PR if that's ok, feels too risky to change all of that at this late stage.

src/scope/server/app.py
leszko: @mjh1 @emranemran you haven't addressed this one. Could you explain somewhere why we need custom logic for downloading recordings?

mjh1: https://linear.app/livepeer/issue/INF-464/dry-up-the-api-call-proxying#comment-3db9196f

mjh1: We don't 👍 good spot

PR #403, opened 2026-02-05
Fix CUDA OOM on pipeline switching

emranemran:
## Summary

Fixes CUDA out-of-memory errors when switching between pipelines (e.g. longlive → krea).

## Root Cause

When `_unload_pipeline_by_id_unsafe()` removes a pipeline from the manager's `_pipelines` dict, active WebRTC sessions still hold references to the old pipeline through:

```
WebRTCManager.sessions → Session → VideoProcessingTrack → FrameProcessor
  → PipelineProcessor.pipeline → old pipeline object (still on GPU)
```

Because `PipelineProcessor` stores a direct reference (`self.pipeline = pipeline`), the old pipeline and all its GPU tensors survive `gc.collect()`. Additionally, pinned memory buffers in `FrameProcessor` were never released during pipeline switching.

## Changes

- **Add `cleanup()` to Pipeline base class** — calls `gc.collect()` and `torch.cuda.empty_cache()` to free GPU memory. Does NOT clear component/state dicts to avoid race conditions with in-flight frame processing.
- **Call `cleanup()` during pipeline unload** — in `_unload_pipeline_by_id_unsafe()`, invoke cleanup before removing references
- **Add GPU memory logging** — log free GPU memory before/after unload and after load, wrapped in try/except to prevent logging failures from disrupting pipeline switching
- **Clear pinned buffer cache on frame processor stop** — release DMA transfer buffers that were never freed
- **Release pipeline reference on processor stop** — set `self.pipeline = None` to break the reference chain and allow GC to reclaim GPU memory

## Test plan

- [x] Deploy to fal.ai staging and verify GPU memory logs appear
- [x] Switch pipelines (longlive → krea) without crashes or `KeyError: 'vae'`
- [x] Confirm GPU memory is reclaimed between pipeline switches
- [ ] `uv run pytest` passes
- [ ] `uv run daydream-scope` starts without errors

### DIFF
--- a/src/scope/cloud/fal_app.py
+++ b/src/scope/cloud/fal_app.py
@@ -169,7 +169,7 @@ def cleanup_session_data():
 # fal deploy fal_app.py --auth public
 
 # Configuration
-DOCKER_IMAGE = "daydreamlive/scope:6334b27"
+DOCKER_IMAGE = "daydreamlive/scope:5340825"
 
 # Create a Dockerfile that uses your existing image as base
 dockerfile_str = f"""


yondonfu:
It seems like the only guaranteed clean way to make sure all GPU mem is freed when unloading a pipeline, regardless of how it is implemented, is to isolate it in a subprocess. @leszko looked into this previously, but we tabled because didn't want to take on the complexity (particularly inter-process comms for an already latency sensitive code path). I would treat that path as a separate thing to consider and discuss.

This leaves the question - what are the practical steps that can be taken to minimize change of this type of OOM during pipeline switching in the short term?

A few ideas from a chat with Claude that is a superset of what is already done in this PR and @varshith15's suggestions:

```
Step 1: Stop active processors and break ref chains BEFORE gc.collect

 The single highest-impact change. Currently _unload_pipeline_by_id_unsafe does:
 1. del self._pipelines[pipeline_id] — drops manager's ref
 2. gc.collect() — but processors still hold refs, so nothing is freed

 Reorder to:
 1. Notify all active FrameProcessors that use this pipeline_id to stop their PipelineProcessors
 2. Each PipelineProcessor.stop() should set self.pipeline = None and drain queues (freeing GPU tensors)
 3. FrameProcessor should clear _pinned_buffer_cache
 4. THEN del self._pipelines[pipeline_id]
 5. THEN gc.collect() + torch.cuda.empty_cache()

 This is similar to PR #403's callback approach, but must be synchronous/blocking — the unload must wait until all processors have actually released their refs.

 Files: pipeline_manager.py, frame_processor.py, pipeline_processor.py

 Step 2: Explicit teardown on Pipeline base class (generic, no plugin work needed)

 Add a teardown() method to Pipeline base class in interface.py that:
 - Iterates self.__dict__ and deletes any torch.nn.Module instances found (catches RIFE's self.rife_interpolator, standard self.components, etc.)
 - Clears self.components, self.state, self.blocks if they exist
 - This is generic — handles any pipeline without requiring custom cleanup
 - Called by pipeline_manager after all external refs are broken (Step 1) but before gc.collect

 This addresses the RIFE-specific problem without requiring each pipeline to implement its own cleanup. The base class just introspects __dict__ for GPU-holding objects.

 Files: interface.py, pipeline_manager.py

 Step 3: Properly drain GPU tensors from queues

 PipelineProcessor.stop() currently drains queues with get_nowait() but the tensors go out of scope one by one. For deterministic cleanup:
 - Drain all tensors into a list, then del the list explicitly
 - Alternatively, just ensure self.pipeline = None is set in stop() (as PR #403 proposes)

 The queue draining in stop() already exists (lines 171-185 of pipeline_processor.py) — just ensure that after draining, there's no lingering ref.

 Files: pipeline_processor.py

 Step 4: Clear pinned buffer cache on stop

 FrameProcessor._pinned_buffer_cache holds pinned CUDA memory buffers that are never released during pipeline switching. Add to FrameProcessor.stop():

 with self._pinned_buffer_lock:
     self._pinned_buffer_cache.clear()

 This is small but pinned memory can be non-trivial.
```

I think what feels reasonable is for there to be a generic cleanup fn that does a best effort cleanup for all pipelines by introspecting `__dict__` and it just leaves possibility of edge cases right now.

yondonfu:
On the backend-fal-v6 branch I can't seem to repro the issue of memory not being freed with the following combos:

video-depth-anything + longlive + rife -> passthrough
reward-forcing -> passthrough
memflow -> passthrough

In all cases, when the switch to passthrough is completed I see the VRAM drop down to baseline levels.

I did not get a chance to try with Krea though. So, I wonder if it is Krea specific.

If testing on a H100 one thing to try that comes to mind is changing this line https://github.com/daydreamlive/scope/blob/0febd6a4a84665d22f6a720866ebba3c10f5b83d/src/scope/server/pipeline_manager.py#L759 to compile=False as this is one of the bigger differences relative to other pipelines that is specific to H100s. Perhaps the torch.compile happening under the hood is causing some caching that prevents GPU mem from being freed...

PR #406, opened 2026-02-06
kafka integration

emranemran:
### Summary                                                                                                                                                                                                     
  - Add `connection_id` parameter to the Kafka publisher API so all server-side events can be correlated 
  with the fal.ai WebSocket connection that triggered them
  - Thread `connection_id` from the WebRTC offer through Session, tracks, FrameProcessor,                
  PipelineProcessor, and PipelineManager
  - Remove the duplicate `KafkaPublisher` class from `fal_app.py` and replace it with a lazy import of   
  the server's standardized publisher

  ### Problem

  - `fal_app.py` had its own `KafkaPublisher` class with a different API (`publish(event_type,
  data_dict)`) and a different default topic (`"network_events"`)
  - Server-side Kafka events had no `connection_id`, making it impossible to correlate them with the fal 
  WebSocket connection
  - Two publishers meant two different envelope formats for the same system

  ### Changes

  **Commit 1: Add `connection_id` to kafka publisher API and schemas**
  - `connection_id: str | None = None` added to all 4 publish methods in `kafka_publisher.py`
  - `connection_id` field added to `WebRTCOfferRequest` and `PipelineLoadRequest` schemas

  **Commit 2: Thread `connection_id` through server components (8 files)**
  - `webrtc.py` → `tracks.py` / `cloud_track.py` → `frame_processor.py` → `pipeline_processor.py`        
  - `pipeline_manager.py`: pipeline load/unload/error events include `connection_id` when triggered from 
  cloud mode
  - `fal_app.py`: injects `connection_id` into offer and pipeline load requests to the scope server      

  **Commit 3: Remove duplicate KafkaPublisher from `fal_app.py`**
  - Deleted ~100 lines of duplicate publisher class
  - Lazy imports server's `KafkaPublisher`; all events now use `KAFKA_TOPIC` env var (default
  `"scope-events"`)

  ### Test plan

  - [ ] `uv run daydream-scope` starts without errors
  - [ ] With `KAFKA_BOOTSTRAP_SERVERS` set, verify events include `connection_id` when available
  - [ ] Verify fal-side events (`websocket_connected`, `websocket_disconnected`) use the standardized    
  envelope

### DIFF
--- a/src/scope/cloud/fal_app.py
+++ b/src/scope/cloud/fal_app.py
@@ -147,7 +147,12 @@ class KafkaPublisher:
             "id": event_id,
             "type": "stream_trace",
             "timestamp": timestamp_ms,
-            "data": {"type": event_type, "client_source": "scope", "timestamp": timestamp_ms, **data},
+            "data": {
+                "type": event_type,
+                "client_source": "scope",
+                "timestamp": timestamp_ms,
+                **data,
+            },
         }
 
         try:
--- a/src/scope/server/frame_processor.py
+++ b/src/scope/server/frame_processor.py
@@ -515,7 +515,9 @@ class FrameProcessor:
                 "fps_in": round(fps_in, 1),
                 "fps_out": round(fps_out, 1),
                 "elapsed_ms": int(elapsed * 1000),
-                "since_last_heartbeat_ms": int((now - self._last_heartbeat_time) * 1000),
+                "since_last_heartbeat_ms": int(
+                    (now - self._last_heartbeat_time) * 1000
+                ),
             }
             if self._cloud_mode:
                 heartbeat_metadata["frames_to_cloud"] = self._frames_to_cloud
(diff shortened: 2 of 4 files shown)


src/scope/cloud/fal_app.py
```diff
@@ -166,10 +166,10 @@ def cleanup_session_data():
 # switch to python 3.10 to match the scope image
 # pip install fal
 # fal auth login
-# fal deploy fal_app.py --auth public
+# fal deploy --env (main/staging/prod) (--app-name X) fal_app.py --auth public
 
 # Configuration
-DOCKER_IMAGE = "daydreamlive/scope:6334b27"
+DOCKER_IMAGE = "daydreamlive/scope:b7c2bff"
```
yondonfu: Are you going to need to commit a change every time to bump the image used?

mjh1: That's been removed now 👍 we can get it automatically based on what's currently checked out

PR #412, opened 2026-02-06
feat: enable plugin params on stream start

ryanontheinside:


yondonfu (commented):
Mind just giving a quick explanation of what these changes do?

Eg not immediately clear to me what that extra model_config field does in Parameters.

PR #413, opened 2026-02-06
fix: stringify enum option values in plugin settings dropdown

ryanontheinside:
The EnumField component stringified the selected value but not the
  dropdown options, causing Radix Select's strict equality check to fail
  for non-string Literal types (e.g. Literal[0, 1]). The dropdown would
  render options when opened but show nothing when closed.


yondonfu (approved):
LGTM!

PR #428, opened 2026-02-10
Cuda oom rebase

mjh1:
This is a reopened version of https://github.com/daydreamlive/scope/pull/403 which got auto-closed unintentionally. This new one is now rebased against main.

### DIFF
--- a/src/scope/core/pipelines/interface.py
+++ b/src/scope/core/pipelines/interface.py
@@ -1,5 +1,6 @@
 """Base interface for all pipelines."""
 
+import gc
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING
 
@@ -76,3 +77,17 @@ class Pipeline(ABC):
             The video tensor is in THWC format and [0, 1] range.
         """
         pass
+
+    def cleanup(self) -> None:
+        """Release GPU resources. Called before pipeline is unloaded.
+
+        Subclasses may override for custom cleanup behavior.
+        """
+        gc.collect()
+        try:
+            import torch
+
+            if torch.cuda.is_available():
+                torch.cuda.empty_cache()
+        except ImportError:
+            pass
--- a/src/scope/core/pipelines/interface.py
+++ b/src/scope/core/pipelines/interface.py
@@ -80,6 +80,7 @@ class Pipeline(ABC):
 
     def cleanup(self) -> None:
         """Release GPU resources. Called before pipeline is unloaded."""
+        import torch
 
         if hasattr(self, "components"):
             try:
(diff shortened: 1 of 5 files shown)


src/scope/core/pipelines/interface.py
```diff
@@ -76,3 +77,41 @@ def __call__(self, **kwargs) -> dict:
...
+            except Exception:
+                pass
+
+        if hasattr(self, "state"):
+            try:
+                if hasattr(self.state, "values"):
+                    self.state.values.clear()
+                del self.state
+            except Exception:
+                pass
```
leszko, line 101: Isn't this part specific to some pipelines. Wouldn't it be possible to just iterate over all fields (like below) and delete rather than assuming that some pipelines have `components` and `state` fields?

src/scope/server/pipeline_processor.py
```diff
@@ -179,25 +179,34 @@ def stop(self):
...
+        # Release pipeline reference to allow GC of GPU resources
+        self.pipeline = None
+
         logger.info(f"PipelineProcessor stopped for pipeline: {self.pipeline_id}")
 
+    def release_pipeline(self):
+        """Release pipeline reference to allow GC of GPU resources."""
+        logger.info(f"Releasing pipeline reference for {self.pipeline_id}")
+        self.pipeline = None
+
```
leszko, line 209: This should not be needed, because we set `self.pipeline = None` in `stop()` and we always call `stop()` and `release_pipeline()` together.

src/scope/server/pipeline_processor.py
```diff
@@ -179,25 +179,34 @@ def stop(self):
...
             while not self.output_queue.empty():
                 try:
-                    self.output_queue.get_nowait()
+                    frame = self.output_queue.get_nowait()
+                    del frame
                 except queue.Empty:
                     break
 
+        # Release pipeline reference to allow GC of GPU resources
+        self.pipeline = None
```
leszko, line 201: I think we should not need to explicitly set `self.pipeline = None`. The whole PipelineProcessor object should be cleaned after the stream is stopped.

I think the flow should be:
1. We start the stream, there are the following objects created: `VideoTrack/CloudTrack`, `FrameProcessor`, and (a few) `PipelineProcessor` objects.
2. We stop the stream, all these objects are deleted
So effectively, the only object that is preserved between stream start/stop, is `PipelineManager`, because we want to avoid reloading the pipelines if they are already loaded. All the rest should be cleaned up in between stream start/stop.

leszko (changes requested):
Added comments. In general, if it solves the OOM issue, than it's great 🎉  In that case, I'll hold off with the Subprocess work.

One general comment is that I think that there are too many changes in this PR and I think that not all of them are needed. I guess the only really needed change is the `cleanup()` function in `interface.py` and it's usage in `pipeline_manager.py`. Maybe also these `del frame`. All the rest should not be needed because it should be handled by GC.

PR #429, opened 2026-02-10
Instant pipeline list reload on cloud connection

mjh1:
Make the pipeline list reload instant when toggling remote inf on/off. Also clean up the polling so we're not constantly polling cloud status when disconnected.

You may want to just review the second commit since git is rendering the renamed file `frontend/src/hooks/useCloudStatus.tsx` as deleted and created.


frontend/src/pages/StreamPage.tsx
```diff
@@ -1228,7 +1235,6 @@ export function StreamPage() {
     <div className="h-screen flex flex-col bg-background">
       {/* Header */}
       <Header
-        onCloudStatusChange={setIsBackendCloudConnected}
```
leszko, line 1231: If we remove it from here, could you remove it also from `Header.tsx`. I think `onCloudStatusChange` property is not used anywhere anymore.

The same probably with `onCloudConnectingChange`.

Also please change if its needed in the components where we pass it, like, `SettingsDialog` or `AccountTab`.

frontend/src/hooks/useCloudStatus.tsx
```diff
@@ -0,0 +1,163 @@
...
+ * Shared cloud status context and hook.
+ *
+ * Provides a single source of truth for cloud connection status across all components.
+ * Uses a React context to share state, so when one component refreshes the status,
+ * all components see the updated value immediately.
+ *
+ * Polling is smart and based on current status:
+ * - Disconnected: no polling (state changes only via explicit user actions)
+ * - Connecting: poll every 1s to quickly detect when connection completes
+ * - Connected: poll every 30s to detect unexpected disconnection
```
leszko: I think it's every `5s`.

PR #439, opened 2026-02-11
feat: NDI support & input sources abstraction

gioelecerati:
**Changes**
- Add NDI support with ctypes
  - NDI sdk detection for darwin, linux, win
  - Use native resolution
- InputSource abstraction used by both the new NDI receiver and the existing spout receiver
- OutputSink abstraction - spout sender moved into it
- Frontend changes

**How to test**
- Install the NDI sdk or NDI tools on your machine
- Expose a NDI source (you can use for example OBS with the DIstroAV plugin enabling the "Preview" output, or install Resolume Arena and enable NDI output from the menu)
- Start scope
- Video input -> NDI -> Select the desired source
- Play

**In a follow-up PR**
- NDI sender for the OutputSink

### DIFF
--- a/frontend/src/components/InputAndControlsPanel.tsx
+++ b/frontend/src/components/InputAndControlsPanel.tsx
@@ -214,15 +214,6 @@ export function InputAndControlsPanel({
     }
   }, [mode, selectedNdiSource, fetchNdiPreview]);
 
-  // Periodically refresh NDI preview while streaming
-  useEffect(() => {
-    if (!isStreaming || mode !== "ndi" || !selectedNdiSource) return;
-    const interval = setInterval(() => {
-      fetchNdiPreview(selectedNdiSource);
-    }, 3000);
-    return () => clearInterval(interval);
-  }, [isStreaming, mode, selectedNdiSource, fetchNdiPreview]);
-
   // Helper function to determine if playhead is at the end of timeline
   const isAtEndOfTimeline = () => {
     if (_timelinePrompts.length === 0) return true;


src/scope/server/frame_processor.py
```diff
@@ -584,6 +628,11 @@ def update_parameters(self, parameters: dict[str, Any]):
             spout_config = parameters.pop("spout_receiver")
             self._update_spout_receiver(spout_config)
 
+        # Handle generic input source settings
+        if "input_source" in parameters:
+            input_source_config = parameters.pop("input_source")
+            self._update_input_source(input_source_config)
```
leszko, line 571: Do we plan to allow changing the input_source when the stream is ongoing?

These parameters here are like "runtime" parameters. There are also the "intial_parameters" used during the initialization.

src/scope/server/frame_processor.py
```diff
@@ -28,6 +30,27 @@
...
+        from scope.core.inputs.spout import SpoutInputSource
+
+        sources[SpoutInputSource.source_id] = SpoutInputSource
+    except Exception:
+        pass
+
+    return sources
+
+
 class _SpoutFrame:
```
leszko: If it's generic now, then I think we should rename it. Also, I think you can move it out of `frame_processor.py`

leszko (commented):
I skimmed through the backend part. Nice work @gioelecerati I think it's in the good direction, I'll do the complete review when you finish your PR.

Some early comments:
1. I see that right now you copy a lot of logic from the Spout received logic, but do I understand correctly that you plan to _replace_ the spout logic and now we'll only use the input source?
2. Now you've implemented the input, but I guess you're planning to do the same for the output. Is that correct?
3. As a side note: `frame_processor.py` is pretty overloaded. I see that you created a separate package `core/inputs`. That's great. Also when you work on `_SpoutFrame` class (or whatever you rename it to, consider moving it to a separate file. The same with some functions that are not strictly related to FrameProcessor. Thanks.

src/scope/server/frame_processor.py
gioelecerati: On the UI I am disallowing to switch from NDI to other inputs and viceversa, but it's just a patch because I am not sure how resolution changes are handled - for example, my NDI source was 1920x1080, switching to the 512x512 example video kept the output video as 1920x1080 but stretched, if you have any insight about this I will work to improve it

frontend/src/components/InputAndControlsPanel.tsx
```diff
@@ -152,6 +165,55 @@ export function InputAndControlsPanel({
...
+    } catch (e) {
+      console.error("Failed to discover NDI sources:", e);
+      setNdiSources([]);
+    } finally {
+      setIsDiscoveringNdi(false);
+    }
+  }, []);
+
+  // NDI preview thumbnail
+  const [ndiPreviewUrl, setNdiPreviewUrl] = useState<string | null>(null);
```
leszko, line 186: This is not being refreshed in realtime. Look at the screenshot, I updated the image, it's correctly changed in the output, but the input is not changed.

Image

frontend/src/lib/api.ts
```diff
@@ -225,6 +225,103 @@ export const getHardwareInfo = async (): Promise<HardwareInfoResponse> => {
...
+
+  if (!response.ok) {
+    const errorText = await response.text();
+    throw new Error(
+      `Probe resolution failed: ${response.status} ${response.statusText}: ${errorText}`
+    );
+  }
+
+  return response.json();
+};
```
leszko, line 316: How did we handle the resolution before? I mean the resolution that the Spout sent, didn't we need a similar logic?

src/scope/server/app.py
```diff
@@ -1131,6 +1131,204 @@ def is_spout_available() -> bool:
     return sys.platform == "win32"
 
 
+@app.get("/api/v1/input-sources")
+async def list_input_source_types():
+    """List available input source types with their availability status."""
+    from scope.core.inputs.ndi import NDIInputSource
```
leszko: Any reason why we use lazy imports here? I guess this endpoint `/api/v1/input-sources` will always be called, so I don't see any benefits of lazy loading.

src/scope/server/frame_processor.py
```diff
@@ -577,12 +563,12 @@ def update_parameters(self, parameters: dict[str, Any]):
         # Handle Spout output settings
         if "spout_sender" in parameters:
             spout_config = parameters.pop("spout_sender")
-            self._update_spout_sender(spout_config)
+            self._update_output_sink_from_spout_config(spout_config)
```
leszko, line 566: Do I understand correctly that you plan to send a separate PR to abstract the output/sink, and then FrameProcessor will be completely unaware of Spout. Is that correct?

src/scope/core/frames.py
```diff
@@ -0,0 +1,17 @@
+"""Lightweight frame wrappers for non-WebRTC input sources."""
+
+
+class RawFrame:
+    """Wraps a raw numpy array to match the VideoFrame interface.
+
+    Allows code that expects a VideoFrame (with .to_ndarray()) to work
+    with plain numpy arrays from input sources like Spout and NDI.
```
leszko: I wonder for Spout, do we need the conversion from Tensor to VideoFrame? Isn't Spout all about keeping everything in GPU?

src/scope/server/app.py
```diff
@@ -1131,6 +1131,204 @@ def is_spout_available() -> bool:
...
+                "source_name": cls.source_name,
+                "source_description": cls.source_description,
+                "available": cls.is_available(),
+            }
+        )
+
+    return {"input_sources": sources}
+
+
+@app.get("/api/v1/input-sources/{source_type}/sources")
```
leszko, line 1177: Do I understand correctly that currently we call `/sources` , `/resolution`, and `/preview` endpoints only for NDI? And for Spout UI does not call them, right?

src/scope/server/app.py
```diff
@@ -1131,6 +1131,204 @@ def is_spout_available() -> bool:
...
+    source_type: str, identifier: str, timeout_ms: int = Query(5000)
+):
+    """Probe the native resolution of a specific input source."""
+    from scope.core.inputs.ndi import NDIInputSource
+    from scope.core.inputs.spout import SpoutInputSource
+
+    source_classes = {
+        NDIInputSource.source_id: NDIInputSource,
+        SpoutInputSource.source_id: SpoutInputSource,
+    }
```
leszko: Can't we use here (and in other places in thsi file), `get_intput_source_classes()`?

src/scope/server/app.py
```diff
@@ -1131,6 +1131,204 @@ def is_spout_available() -> bool:
     return sys.platform == "win32"
 
 
+@app.get("/api/v1/input-sources")
+async def list_input_source_types():
+    """List available input source types with their availability status."""
+    from scope.core.inputs.ndi import NDIInputSource
+    from scope.core.inputs.spout import SpoutInputSource
+
+    source_classes = [SpoutInputSource, NDIInputSource]
```
leszko: Shouldn't we get it from `get_input_source_classes()` instead of hardcoding here?

src/scope/core/frames.py
```diff
@@ -0,0 +1,17 @@
+"""Lightweight frame wrappers for non-WebRTC input sources."""
+
+
+class RawFrame:
```
leszko: Is this used anywhere?

src/scope/server/app.py
```diff
@@ -1131,6 +1131,204 @@ def is_spout_available() -> bool:
...
+        raise HTTPException(
+            status_code=400,
+            detail=f"Input source '{source_type}' is not available on this system. "
+            f"Required dependencies may not be installed.",
+        )
+
+    try:
+        instance = source_class()
+        try:
+            discovered = instance.list_sources(timeout_ms=timeout_ms)
```
leszko, line 1192: I wonder how time consuming this operation is. If it takes a long time to discover all NDI inputs, then you can consider caching it (here in the frontend).

frontend/src/components/InputAndControlsPanel.tsx
gioelecerati: Preview is supposed to be just an image which gets loaded on source selection and when you click on the preview image it gets refreshed

Will make it refreshing automatically instead of manual with an interval

frontend/src/lib/api.ts
gioelecerati: No Spout wasn't probing, it just had 512x512 hardcoded. Which I tried to do the same with NDI, but I was having the output trimmed or stretched depending on how I was applying it

This is also part of the reason why I went for not allowing source switch while you're on NDI, since the output will have the initial parameters resolution set as the NDI source resolution, ending up with a stretched output if you switch to camera or other source

Is there a way to change the resolution mid stream or are we planning to have that hot reloading parameters for w & h?

Either case, if it's possible to change res, I will address it in a follow up PR

src/scope/server/app.py
gioelecerati: Good point, it was a carry over from the bridge - addressing it

src/scope/server/frame_processor.py
gioelecerati: Yes!

src/scope/core/frames.py
gioelecerati: Ooof actually this rawframe is not used anymore, I will delete it

Not sure if I am knowledgeable enough about spout to answer your question tho, can you elaborate?

src/scope/server/app.py
gioelecerati: Correct, but potentially the could be used by spout as well

src/scope/server/app.py
gioelecerati: Good point! Will do

src/scope/core/frames.py
gioelecerati: Right, as mentioned before it's not good catch :)

PR #440, opened 2026-02-11
Fix none error on session close

mjh1:
The culprit ended up being:
https://github.com/daydreamlive/scope/blob/6d504dcb0713bd7c6accb3db23cd6781876ac454/src/scope/server/tracks.py#L188

This was not an async function so await was returning `object NoneType can't be used in 'await' expression`

### DIFF
--- a/src/scope/server/frame_processor.py
+++ b/src/scope/server/frame_processor.py
@@ -115,7 +115,7 @@ class FrameProcessor:
         # This determines whether we wait for video frames or generate immediately.
         self._video_mode = (initial_parameters or {}).get("input_mode") == "video"
 
-        # Pipeline chaining support (local mode only)
+        # Pipeline chaining support
         self.pipeline_processors: list[PipelineProcessor] = []
         self.pipeline_ids: list[str] = []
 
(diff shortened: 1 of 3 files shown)


src/scope/server/tracks.py
```diff
@@ -185,4 +185,7 @@ async def stop(self):
         if self.frame_processor is not None:
             self.frame_processor.stop()
 
-        await super().stop()
+        # MediaStreamTrack.stop() may be sync or async depending on aiortc version
```
leszko: Shouldn't we instead of handling both aiortc versions, jus pin to the specific aiortc version?

mjh1: I don't know if we want to pin the exact version just for this? We have `aiortc>=1.13.0` set right now. Is there some static analysis tool we can run to check it? Anyways, I've removed the extra check and just made it work with the version we're on.

src/scope/server/frame_processor.py
```diff
@@ -115,7 +115,7 @@ def __init__(
         # This determines whether we wait for video frames or generate immediately.
         self._video_mode = (initial_parameters or {}).get("input_mode") == "video"
 
-        # Pipeline chaining support (local mode only)
```
mjh1, line 118: I verified this works for remote inference 👍

PR #442, opened 2026-02-11
Add LoRA support in cloud relay mode

mjh1:
Enable LoRA browsing, selection, and install-by-URL in both cloud and local modes. Previously the LoRA UI was completely hidden in cloud mode and the backend had no cloud proxy for the LoRA list endpoint.

Backend: add cloud proxy to GET /api/v1/lora, add POST /api/v1/lora endpoint with filename inference, extension validation, and duplicate detection. Add 5-min timeout for LoRA installs in fal_app.py.

Frontend: remove isCloudMode gates in SettingsPanel and ComplexFields, switch LoRAManager to useApi hook for cloud-aware routing, add installLoRAFile to api/cloudAdapter/useApi, add install URL input UI.

### DIFF
--- a/frontend/src/components/Header.tsx
+++ b/frontend/src/components/Header.tsx
@@ -87,7 +87,9 @@ export function Header({
       if (openSettingsTab === "plugins") {
         setPluginsOpen(true);
       } else {
-        setInitialTab(openSettingsTab as "general" | "account" | "api-keys" | "loras");
+        setInitialTab(
+          openSettingsTab as "general" | "account" | "api-keys" | "loras"
+        );
         setSettingsOpen(true);
       }
       onSettingsTabOpened?.();
--- a/frontend/src/components/SettingsDialog.tsx
+++ b/frontend/src/components/SettingsDialog.tsx
@@ -8,11 +8,7 @@ import { ReportBugDialog } from "./ReportBugDialog";
 import { usePipelinesContext } from "@/contexts/PipelinesContext";
 import { useLoRAsContext } from "@/contexts/LoRAsContext";
 import { LoRAsTab } from "./settings/LoRAsTab";
-import {
-  installLoRAFile,
-  deleteLoRAFile,
-  getServerInfo,
-} from "@/lib/api";
+import { installLoRAFile, deleteLoRAFile, getServerInfo } from "@/lib/api";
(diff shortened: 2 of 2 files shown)


frontend/src/components/LoRAManager.tsx
```diff
@@ -99,6 +130,31 @@ export function LoRAManager({
...
+      await installPromise;
+      setInstallUrl("");
+      await loadAvailableLoRAs();
+    } catch (error) {
+      setInstallError(
+        error instanceof Error ? error.message : "Install failed"
+      );
+    } finally {
+      setIsInstalling(false);
+    }
```
leszko: Nit: I think you'll display the toast with error twice here.

frontend/src/components/LoRAManager.tsx
leszko: `loadAvailableLoRAs` should be here in the dependency array.

yondonfu (changes requested):
This doesn't seem to support CivitAI right now. Using the download URL for https://civitai.com/models/2383884?modelVersionId=2680702 results in an error about an invalid file extension.

Since the current LoRA [user guide](https://github.com/daydreamlive/scope/blob/main/docs/lora.md) calls out that LoRAs can be sourced from HuggingFace and CivitAI and given that the latter is popular for LoRAs, I think we should make sure that is supported before introducing the LoRA download feature.

The linked LoRA guide contains guidance on how to programatically get the LoRA from CivitAI which might be a useful reference here. It will require a CivitAI API key which can be configured in the API Keys tab and the existing API key handling code might be specific to HuggingFace so it would just need to be tweaked a bit to handle CivitAI API keys too https://github.com/daydreamlive/scope/blob/1a7052fde6fbfbfb76f35aab9745c77fcd2d8a79/src/scope/server/app.py#L1352.

frontend/src/components/LoRAManager.tsx
mjh1: this has moved but it's fixed now 👍

frontend/src/components/LoRAManager.tsx
mjh1: Fixed in https://github.com/daydreamlive/scope/pull/442/commits/76002729f50845132249166ad72d897b549106db

PR #443, opened 2026-02-11
Avoid sending local image paths for cloud mode

mjh1:
Fix the situation where we might reference local image paths on the cloud server, we now:
- Clear previously selected images on cloud connection state change
- Refresh the list in the image picker dynamically when the connection state changes rather than only when opening the picker

### DIFF
--- a/frontend/src/components/InputAndControlsPanel.tsx
+++ b/frontend/src/components/InputAndControlsPanel.tsx
@@ -27,6 +27,7 @@ import {
   parseInputFields,
 } from "../lib/schemaSettings";
 import { SchemaPrimitiveField } from "./PrimitiveFields";
+import { useCloudStatus } from "../hooks/useCloudStatus";
 
 interface InputAndControlsPanelProps {
   className?: string;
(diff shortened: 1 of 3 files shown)


frontend/src/components/InputAndControlsPanel.tsx
```diff
@@ -174,6 +175,47 @@ export function InputAndControlsPanel({
     }
   }, [localStream]);
 
+  // Track cloud connection state and clear images when it changes
+  // (switching between local/cloud means different asset lists)
+  const { isConnected: isCloudConnected, isConnecting: isCloudConnecting } =
```
leszko: Do we need this `isCloudConnecting`, It's not checked anywhere and it seems it could be enough to have `isCloudConnected`.

frontend/src/components/MediaPicker.tsx
```diff
@@ -45,6 +46,25 @@ export function MediaPicker({
...
+      return;
+    }
+
+    // If connection state changed and picker is open, reload images
+    if (prevCloudConnectedRef.current !== isCloudConnected && isOpen) {
+      loadImages();
+    }
+
+    prevCloudConnectedRef.current = isCloudConnected;
+  }, [isCloudConnected, isOpen]);
```
leszko: Shouldn't you add `loadImages` in the dependency array?

frontend/src/components/InputAndControlsPanel.tsx
mjh1: Good spot, yep that shouldn't be needed

PR #444, opened 2026-02-11
Add Electron App for Mac build

leszko:
## Summary
- Add macOS build job to the Electron app CI workflow, producing signed and notarized artifacts for every release
- Include Apple notarization via a new `afterSign` hook so the app passes Gatekeeper on user machines
- Extend `update-latest-yml.js` to generate `latest-mac.yml` for the macOS auto-updater
- Improve auto-updater error handling with user-facing dialogs on download failure and update errors

## Release artifacts
Each release will now include 2 additional macOS files:
| Artifact | Purpose |
|---|---|
| `DaydreamScope-arm64.dmg` | Installer for Apple Silicon Macs |
| `DaydreamScope-*-arm64.zip` | Update payload for Apple Silicon (required by electron-updater) |

The ZIP files are required because `electron-updater` on macOS downloads ZIP archives (not DMGs) when performing in-app auto-updates.

All files are signed with Apple Developer ID credentials and notarized, so users will not see Gatekeeper warnings.

## Testing
Tested all features on macOS including pipelines, WebRTC streaming, settings, plugins, and the auto-updater — everything works correctly.

### DIFF
--- a/.github/workflows/build-electron.yml
+++ b/.github/workflows/build-electron.yml
@@ -105,3 +105,123 @@ jobs:
             app/dist/**/*
           retention-days: 30
           if-no-files-found: error
+
+  build-macos:
+    name: Build macOS
+    runs-on: macos-latest
+    env:
+      # Map CI_MACOS_* secrets to env vars used by certificate import and electron-builder
+      MACOS_SIGNING_IDENTITY: ${{ secrets.CI_MACOS_CERTIFICATE_ID }}
+      MACOS_SIGNING_CERT: ${{ secrets.CI_MACOS_CERTIFICATE_BASE64 }}
+      MACOS_SIGNING_CERT_PASSWORD: ${{ secrets.CI_MACOS_CERTIFICATE_PASSWORD }}
+      MACOS_NOTARIZATION_USERNAME: ${{ secrets.CI_MACOS_NOTARIZATION_USER }}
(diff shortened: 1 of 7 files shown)


yondonfu (commented):
There were a few updates to the Windows desktop app recently:

https://github.com/daydreamlive/scope/pull/435
https://github.com/daydreamlive/scope/pull/448

Can you double check if anything needs to be ported to Mac builds as well before starting to support them?

Also, while I think having a Mac build makes sense...I'm a bit wary of starting to support it without first considering these questions:

1. What if a plugin includes a pipeline that won't run on Mac? How should that be handled?
2. What if a plugin includes a pipeline that runs on Mac, but runs poorly? How should that be handled?

Curious if you have thoughts on this.

yondonfu:
@leszko

> IMO this is not a blocker for releasing the Electron app on macOS. However, it’s something we should consider more broadly in the context of plugins and environment compatibility.

I think your points make sense. But, I'd like to double check that we've tested the installation and runtime failure scenarios before moving forward with supporting Mac releases. It's ok if certain plugins/pipelines don't work on Mac, but I'd like to verify that in the scenarios that you describe the installation/use of the plugins don't break the app installation for Mac users in a non-recoverable way (eg no way to get back to working install without needing to go into the CLI) as that would then become a pain to deal with support wise.

PR #445, opened 2026-02-11
feat: add dynamic parameter UI for pre/postprocessor plugins

### LINKED ISSUE #414: Post-processor plugins: dynamic parameter sliders not visible in UI
### Summary

I built a VFX plugin (scope-vfx) with GPU-accelerated chromatic aberration + VHS/retro CRT effects using pure PyTorch ops. It registers via pluggy like any other pipeline and has 8 custom parameters (toggles + sliders) using ui_field_config.

When I register it as a main pipeline (no usage set), everything works great. The dynamic parameter sliders show up in the settings panel and I can tweak all the effects live. Screenshot for reference:

Image

Then I switched it to usage = [UsageType.POSTPROCESSOR] so I could chain it after generative models like LongLive and apply the VFX on top of AI-generated video instead of raw camera input. The effects apply correctly in the chain, but the parameter sliders don't show up for post-processors.

The post-processor dropdown selects the pipeline, but there's no settings panel that expands for it.

Image

Workaround I found: select VFX Pack as the main pipeline first, dial in the settings, then switch back to the generative model with VFX Pack as post-processor. The parameter values persist and get applied. Works, but not exactly discoverable.

I think the gap is that the frontend renders dynamic parameters for the main pipeline slot but doesn't have an equivalent section for pre/post-processor slots. The backend side is fine, schemas are served correctly and parameters flow through the chain. It's just the UI that doesn't expose them.

### Platform

Win 11, 64GB RAM

### Nvidia GPU

NVIDIA GeForce RTX 5090, 32607 MiB

### Scope Version

daydream-scope: 0.1.0b3 git commit: 0febd6a

### uv Version

uv 0.9.8 (85c5d3228 2025-11-07)

### node Version

v24.11.0

ryanontheinside:
resolves #414

### DIFF
--- a/src/scope/server/frame_processor.py
+++ b/src/scope/server/frame_processor.py
@@ -877,13 +877,11 @@ class FrameProcessor:
             # Get pipeline instance from manager
             pipeline = self.pipeline_manager.get_pipeline_by_id(pipeline_id)
 
-            proc_params = self.parameters.copy()
-
             # Create processor with its own queues
             processor = PipelineProcessor(
                 pipeline=pipeline,
                 pipeline_id=pipeline_id,
-                initial_parameters=proc_params,
+                initial_parameters=self.parameters.copy(),
                 session_id=self.session_id,
                 user_id=self.user_id,
                 connection_id=self.connection_id,


frontend/src/pages/StreamPage.tsx
```diff
@@ -657,14 +673,47 @@ export function StreamPage() {
...
-  const handlePostprocessorIdsChange = (ids: string[]) => {
-    updateSettings({ postprocessorIds: ids });
+  type PluginKind = keyof typeof pluginSettingsKeys;
+
+  const makePluginIdsHandler = (kind: PluginKind) => (ids: string[]) => {
+    const k = pluginSettingsKeys[kind];
+    updateSettings({ [k.ids]: ids, [k.overrides]: {} });
   };
 
+  const makePluginOverrideHandler =
```
leszko: nit: I think we should call it here Pipeline (not Plugin). This override is related to pipelines, no matter if they are built-in or plugins.

src/scope/server/frame_processor.py
```diff
@@ -584,7 +584,14 @@ def update_parameters(self, parameters: dict[str, Any]):
             spout_config = parameters.pop("spout_receiver")
             self._update_spout_receiver(spout_config)
 
-        # Update parameters for all pipeline processors
+        # Route per-pipeline params to the correct processor
+        per_pipeline = parameters.pop("pipeline_params", None) or {}
+        for processor in self.pipeline_processors:
+            extra = per_pipeline.get(processor.pipeline_id)
+            if extra:
+                processor.update_parameters(extra)
```
leszko: I’m not a fan of treating pre/post-processor parameters differently from the main pipeline. It doesn’t actually solve the name collision issue and adds extra complexity. I suggest one of these approaches:
1. **Simple (good enough for now)**  
   Send all params for pre/post-processors and the main pipeline together. Name collisions will still break things, but no backend changes are needed in this PR.
2. **More robust (valid, but more complex)**  
   Include the main pipeline parameters inside `pipeline_params`. Complexity arises because we need to:  
   - Handle initial parameters (e.g., `height`, `width`)  
   - Distinguish params unrelated to a specific pipeline (like "pause" events)  
   - Maintain backward compatibility until the cloud deploys the latest version  

Why the current solution doesn’t solve name collisions?  => If, for example, both a pre-processor and the main pipeline have a `scale` parameter, changing it for the main pipeline will also affect the pre-processor.

leszko (approved):
LGTM 👍

PR #446, opened 2026-02-11
Prevalidate plugin entrypoints to check for broken plugins

yondonfu:
## Summary
- **Plugin manager**: Added `_prevalidate_entrypoints()` which enforces a single entry point per plugin, records failures as `FailedPluginInfo`, and blocks broken entry points from being registered in pluggy
- **API**: `GET /api/v1/plugins` now returns `failed_plugins` alongside `plugins` so the frontend can surface errors
- **Frontend**: `PluginsTab` shows a warning banner and per-plugin error details when plugins fail to load
- **Registry**: Wraps plugin pipeline loading in try/except so built-in pipelines survive plugin failures
- **Tests**: 8 new tests covering error handling, blocking, failure recording, and pipeline registration resilience

## Test plan
- [x] Run `uv run pytest tests/test_plugin_manager.py` — all 8 new tests pass
- [x] Install a plugin with a broken entry point (e.g. bad import) and confirm it appears in the failed plugins UI
- [x] Confirm healthy plugins still load and built-in pipelines are unaffected

### DIFF
--- a/frontend/src/components/SettingsDialog.tsx
+++ b/frontend/src/components/SettingsDialog.tsx
@@ -15,6 +15,7 @@ import {
   restartServer,
   waitForServer,
   getServerInfo,
+  type FailedPluginInfo,
 } from "@/lib/api";
 import { toast } from "sonner";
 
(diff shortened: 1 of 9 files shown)


yondonfu:
Screenshot 2026-02-11 153046

This is how the UI shows broken plugins.

PR #457, opened 2026-02-13
feat: ndi output support & ndi live preview

gioelecerati:
**Summary**
- NDI output support
- `OutputSink` -> `OutputSinks` to allow having both Spout and NDI producing outputs in separate threads
- UI updates
- MJPEG live preview (10fps - 320px)

### DIFF
--- a/frontend/src/components/InputAndControlsPanel.tsx
+++ b/frontend/src/components/InputAndControlsPanel.tsx
@@ -8,6 +8,7 @@ import {
   SelectValue,
 } from "./ui/select";
 import { ToggleGroup, ToggleGroupItem } from "./ui/toggle-group";
+import { Switch } from "./ui/switch";
 import { Badge } from "./ui/badge";
 import { Input } from "./ui/input";
 import { Upload, ArrowUp, RefreshCw } from "lucide-react";
@@ -184,8 +185,9 @@ export function InputAndControlsPanel({
   }, []);
 
   // Live MJPEG preview URL
+  const [showNdiPreview, setShowNdiPreview] = useState(false);
   const ndiStreamUrl =
-    mode === "ndi" && selectedNdiSource
+    mode === "ndi" && selectedNdiSource && showNdiPreview
       ? getInputSourceStreamUrl("ndi", selectedNdiSource)
       : null;
   const [isStreamLoaded, setIsStreamLoaded] = useState(false);
@@ -396,8 +398,25 @@ export function InputAndControlsPanel({
                     />
                   </Button>
                 </div>
-                {/* Live NDI preview (MJPEG stream) */}
+                {/* Preview toggle */}
                 {selectedNdiSource && (
+                  <div className="flex items-center gap-2 text-xs">
+                    <Switch
+                      id="ndi-preview-toggle"
+                      checked={showNdiPreview}
+                      onCheckedChange={setShowNdiPreview}
+                      disabled={isStreaming}
+                    />
+                    <label
+                      htmlFor="ndi-preview-toggle"
+                      className="text-muted-foreground cursor-pointer select-none"
+                    >
+                      Show live preview
+                    </label>
+                  </div>
+                )}
+                {/* Live NDI preview (MJPEG stream) */}
+                {selectedNdiSource && showNdiPreview && (
                   <div className="relative rounded-md overflow-hidden border border-border bg-muted min-w-0">
                     {ndiStreamUrl ? (
                       <img


src/scope/server/app.py
```diff
@@ -1297,6 +1303,94 @@ async def get_input_source_preview(
...
+        return None
+    for session in webrtc_manager.sessions.values():
+        vt = getattr(session, "video_track", None)
+        fp = getattr(vt, "frame_processor", None) if vt else None
+        if fp and fp.running and fp.input_source_type == source_type:
+            return fp
+    return None
+
+
+@app.get("/api/v1/input-sources/{source_type}/sources/{identifier:path}/stream")
```
leszko, line 1249: Is this used instead of `/preview`? Can we then remove the `/preview` endpoint?

src/scope/server/app.py
```diff
@@ -1297,6 +1303,94 @@ async def get_input_source_preview(
...
+async def stream_input_source_preview(
+    source_type: str, identifier: str, fps: int = Query(10, ge=1, le=30)
+):
+    """MJPEG stream of an input source for live preview.
+
+    During an active stream, reads from the FrameProcessor's raw-frame buffer
+    (zero extra receiver overhead).  Otherwise spins up a temporary receiver.
```
leszko: Couldn't we _always_ spin up a separate NDI receiver? I find it weird and architecturally "not nice" that we use frame processor here.

src/scope/server/schema.py
```diff
@@ -114,9 +114,13 @@ class Parameters(BaseModel):
...
         description="Update scales for loaded LoRA adapters. Each entry updates a specific adapter by path.",
     )
+    output_sinks: "dict[str, OutputSinkConfig] | None" = Field(
+        default=None,
+        description="Output sinks config keyed by type, e.g. {'spout': {'enabled': true, 'name': 'ScopeOut'}, 'ndi': {'enabled': true, 'name': 'Scope'}}",
+    )
     spout_sender: "SpoutConfig | None" = Field(
         default=None,
-        description="Spout output configuration for sending frames to external apps",
+        description="(Deprecated, use output_sinks) Spout output configuration",
```
leszko: Should we maybe remove everything about Spout from this file? If we have this abstraction, then we should not need anything about spout from now on.

src/scope/server/frame_processor.py
```diff
@@ -560,7 +573,12 @@ def _get_pipeline_dimensions(self) -> tuple[int, int]:
 
     def update_parameters(self, parameters: dict[str, Any]):
         """Update parameters that will be used in the next pipeline call."""
-        # Handle Spout output settings
+        # Handle generic output sinks config
+        if "output_sinks" in parameters:
+            sinks_config = parameters.pop("output_sinks")
+            self._update_output_sinks_from_config(sinks_config)
+
+        # Handle legacy Spout output settings (backward compat)
```
leszko: Do we need to support this legacy stuff here? I think the Scope is always delivered frontend+backend together, so that shouldn't be needed, wdyt?

I'd love to get rid of all mentions of Spout in Frame Processor.

src/scope/server/frame_processor.py
```diff
@@ -238,20 +234,20 @@ def stop(self, error_message: str = None):
...
-        # Clean up output sink
-        self.output_sink_enabled = False
-        if self.output_sink_thread and self.output_sink_thread.is_alive():
+        # Clean up all output sinks
+        for sink_type, entry in list(self.output_sinks.items()):
             try:
-                self.output_sink_queue.put_nowait(None)
+                entry["queue"].put_nowait(None)
             except queue.Full:
                 pass
```
leszko: Not sure how critical it is, but if you signal with `None` to stop the thread, then if the queue is full, then the thread will never be stopped.

src/scope/server/frame_processor.py
```diff
@@ -238,20 +234,20 @@ def stop(self, error_message: str = None):
...
-            self.output_sink_thread.join(timeout=2.0)
-        if self.output_sink is not None:
+            thread = entry.get("thread")
+            if thread and thread.is_alive():
+                thread.join(timeout=2.0)
```
leszko, line 249: What is it takes more than 2s for the tread to complete? If it's problematic then at least I'd check if the thread is still active and print some warning. But maybe there is a better way to handle this.

leszko (approved):
Added 2 comments. Other than that LGTM 👍  Great work @gioelecerati 🎉 

One thing you can just double-check is if you enable both NDI in and NDI out. If it works smoothly and it the resources it consumes look ok. Saying that, because I was mainly testing it on a Windows TensorDock machine and at some point it performed very poorly. Then, after the machine restart it was good again. So, I'd assume it's related to TensorDock stability. But worth double-checking.

PR #480, opened 2026-02-17
Audio Support for Scope

BuffMcBigHuge:
# Audio Support for Scope

## Summary

Adds end-to-end audio support to Scope's WebRTC streaming pipeline. Pipelines can now return audio alongside video in their output dict; the server buffers, resamples, and streams audio over WebRTC and NDI. A shared media clock keeps audio and video synchronized.

### Backend

- **Pipeline interface**: Pipelines may return `{"video": ..., "audio": ..., "audio_sample_rate": ...}`. Audio keys are optional; pipelines that don't produce audio are unchanged.
- **PipelineProcessor**: New `audio_output_queue` for audio chunks from the pipeline output dict.
- **FrameProcessor**: Audio drain thread reads from the last processor's audio queue, resamples to 48 kHz (WebRTC standard), mixes to mono, and buffers 20 ms chunks. `get_audio()` drains the buffer for the audio track.
- **MediaClock**: Shared clock for A/V sync. Both video and audio tracks derive PTS from `get_media_time()` so RTCP Sender Reports map correctly to NTP.
- **AudioProcessingTrack**: aiortc `MediaStreamTrack` that produces 20 ms frames at 48 kHz. Returns silence when no audio is available to keep the track alive.
- **NDI**: `NDIOutputSink.send_audio()` sends float32 audio via `NDIlib_send_send_audio_v2`.

### Frontend

- **VideoOutput**: Mute/unmute toggle (speaker icon). Starts muted to satisfy browser autoplay policy; user can unmute once the stream is playing.
- **useUnifiedWebRTC**: Merges video and audio tracks into a single `MediaStream`. Adds a recvonly audio transceiver so the SDP offer includes an audio m-line for the backend to attach its track.

### WebRTC Handshake

The browser adds `addTransceiver("audio", { direction: "recvonly" })` so the offer includes an audio m-line. After `setRemoteDescription`, the backend finds the audio transceiver, attaches its `AudioProcessingTrack`, and sets direction to `sendonly`. The answer then indicates that the server will send audio.

## Architecture

```
Pipeline.__call__() → {"video": tensor, "audio": tensor, "audio_sample_rate": int}
    │
    ▼
PipelineProcessor.audio_output_queue
    │
    ▼
FrameProcessor._audio_drain_loop (resample to 48 kHz, buffer)
    │
    ├──► AudioProcessingTrack.recv() → WebRTC
    └──► NDIOutputSink.send_audio() → NDI
```

## Related

- Architecture doc.
- Pipelines that produce audio (e.g. [LTX-2](https://github.com/daydreamlive/scope-ltx-2)) are wired separately; this PR provides the infrastructure.

### DIFF
--- a/src/scope/server/frame_processor.py
+++ b/src/scope/server/frame_processor.py
@@ -553,7 +553,7 @@ class FrameProcessor:
                 with self._audio_buffer_lock:
                     self._audio_buffer.append(audio_mono)
                     self._audio_buffer_samples += len(audio_mono)
-                    logger.info(
+                    logger.debug(
                         f"[FRAME-PROCESSOR] Audio buffered: {len(audio_mono)} samples "
                         f"(total: {self._audio_buffer_samples}, sr={sample_rate})"
                     )
--- a/src/scope/server/pipeline_processor.py
+++ b/src/scope/server/pipeline_processor.py
@@ -476,7 +476,11 @@ class PipelineProcessor:
 
             # Forward extra params to downstream pipeline (dual-output pattern)
             # Preprocessors return {"video": frames, "vace_input_frames": ..., "vace_input_masks": ...}
-            extra_params = {k: v for k, v in output_dict.items() if k != "video"}
+            # Audio keys are handled separately via audio_output_queue, not as pipeline params.
+            _non_param_keys = {"video", "audio", "audio_sample_rate"}
+            extra_params = {
+                k: v for k, v in output_dict.items() if k not in _non_param_keys
+            }
             if extra_params and self.next_processor is not None:
                 self.next_processor.update_parameters(extra_params)
 
@@ -548,7 +552,7 @@ class PipelineProcessor:
             if audio_output is not None and audio_sample_rate is not None:
                 # Detach and move to CPU for downstream consumption
                 audio_output = audio_output.detach().cpu()
-                logger.info(
+                logger.debug(
                     f"[PIPELINE-PROC] Audio from {self.pipeline_id}: "
                     f"shape={audio_output.shape}, sr={audio_sample_rate}"
                 )
--- a/src/scope/server/tracks.py
+++ b/src/scope/server/tracks.py
@@ -31,6 +31,7 @@ class VideoProcessingTrack(MediaStreamTrack):
         fps: int = 30,
         initial_parameters: dict = None,
         notification_callback: callable = None,
+        media_clock: MediaClock | None = None,
         session_id: str | None = None,
         user_id: str | None = None,
         connection_id: str | None = None,
@@ -40,6 +41,7 @@ class VideoProcessingTrack(MediaStreamTrack):
         self.pipeline_manager = pipeline_manager
         self.initial_parameters = initial_parameters or {}
         self.notification_callback = notification_callback
+        self.media_clock = media_clock
         self.session_id = session_id
         self.user_id = user_id
         self.connection_id = connection_id
@@ -54,6 +56,8 @@ class VideoProcessingTrack(MediaStreamTrack):
         self._paused = False
         self._paused_lock = threading.Lock()
         self._last_frame = None
+        self._last_send_time: float | None = None
+        self._clock_started = False
 
         # Server-side input mode - when enabled, frames come from the backend
         # instead of WebRTC (no browser video track needed)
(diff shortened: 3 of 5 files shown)


src/scope/core/outputs/ndi.py
```diff
@@ -162,6 +163,61 @@ def send_frame(self, frame: np.ndarray | torch.Tensor) -> bool:
             logger.error(f"Error sending NDI frame: {e}")
             return False
 
+    def send_audio(
```
leszko, line 166: Nice that you added the audio for NDI as well 🏅

src/scope/server/frame_processor.py
```diff
@@ -474,6 +507,133 @@ def get(self) -> torch.Tensor | None:
...
+    def _audio_drain_loop(self):
+        """Background thread that drains audio from the last pipeline processor's
+        audio_output_queue, resamples to 48kHz, and appends to the audio buffer.
+        """
+        logger.info("[FRAME-PROCESSOR] Audio drain thread started")
+
+        while self.running:
+            if not self.pipeline_processors:
+                time.sleep(0.01)
+                continue
```
leszko, line 519: Can we maybe start the audio loop AFTER the pipeline processors are created, then we wouldn't need to have this check here.

src/scope/server/frame_processor.py
```diff
@@ -474,6 +507,133 @@ def get(self) -> torch.Tensor | None:
...
+                )
+            except queue.Empty:
+                continue
+
+            try:
+                # Convert torch tensor to numpy float32
+                if isinstance(audio_tensor, torch.Tensor):
+                    audio_np = audio_tensor.float().numpy()
+                else:
+                    audio_np = np.asarray(audio_tensor, dtype=np.float32)
```
leszko, line 534: Do we accept the pipelines to create audio in two formats? Or if not, then why would need this check?

src/scope/server/frame_processor.py
```diff
@@ -474,6 +507,133 @@ def get(self) -> torch.Tensor | None:
...
+
+                # Ensure shape is [C, S] (channels, samples)
+                if audio_np.ndim == 1:
+                    audio_np = audio_np[np.newaxis, :]  # mono -> [1, S]
+
+                # Mix down to mono for WebRTC (average channels)
+                if audio_np.shape[0] > 1:
+                    audio_mono = audio_np.mean(axis=0)
+                else:
+                    audio_mono = audio_np[0]
```
leszko, line 544: Similar question, we have a lot of checks for the audio tensor shape, do we allow returning different formats?

src/scope/server/frame_processor.py
```diff
@@ -474,6 +507,133 @@ def get(self) -> torch.Tensor | None:
 
         return frame
 
+    def _audio_drain_loop(self):
```
leszko, line 510: Nit: I wonder if we could/should move the audio related code into a separate file, like audio.py. The main reason is that frame_processor.py becomes busy.

src/scope/server/pipeline_processor.py
```diff
@@ -509,6 +529,25 @@ def process_chunk(self):
                     )
                     continue
 
+            # Extract audio from pipeline output and queue it
+            audio_output = output_dict.get("audio")
+            audio_sample_rate = output_dict.get("audio_sample_rate")
+            if audio_output is not None and audio_sample_rate is not None:
+                # Detach and move to CPU for downstream consumption
+                audio_output = audio_output.detach().cpu()
+                logger.info(
```
leszko: I think it should be at the debug log, because the logs get too noisy.

src/scope/server/frame_processor.py
```diff
@@ -474,6 +507,133 @@ def get(self) -> torch.Tensor | None:
...
+                    audio_mono = self._resample_audio(
+                        audio_mono, sample_rate, WEBRTC_AUDIO_SAMPLE_RATE
+                    )
+
+                # Append to buffer
+                with self._audio_buffer_lock:
+                    self._audio_buffer.append(audio_mono)
+                    self._audio_buffer_samples += len(audio_mono)
+                    logger.info(
+                        f"[FRAME-PROCESSOR] Audio buffered: {len(audio_mono)} samples "
```
leszko, line 557: I think this should be a the the `debug` level, because the logs get too noisy.

src/scope/server/tracks.py
```diff
@@ -189,3 +197,69 @@ async def stop(self):
...
+
+        # Create AudioFrame: shape must be (1, num_samples) for mono s16 layout
+        frame = AudioFrame.from_ndarray(
+            audio_int16.reshape(1, -1), format="s16", layout="mono"
+        )
+        frame.sample_rate = AUDIO_CLOCK_RATE
+
+        # Set PTS from shared media clock for A/V sync
+        media_time = self.media_clock.get_media_time()
+        frame.pts = self.media_clock.media_time_to_audio_pts(media_time)
```
j0sh: I think the corresponding call for video is missing?

nit: since media_time_to_audio_pts is called immediately is after get_media_time, I think this API can be simplified to something like

```
frame.pts = self.media_clock.to_pts(AUDIO_CLOCK_RATE)
```

Since AUDIO_CLOCK_RATE is being manually set elsewhere anyway, I am not sure if there really needs to be separate API entry points for audio / video

j0sh (commented):
One note on timing / synchronization. Do the audio and video pipelines run independently?

As far as I can tell, media is timestamped right before sending out to WebRTC, in which case there will likely be desync if the pipeline for one track is delayed compared to another. You might want to propagate a reference timestamp at the beginning of both pipelines earlier in the process, but I don't really know this codebase well enough to suggest where.

Re: sync, there might also be more subtle WebRTC usage issues but I am not sure yet; things look _mostly_ fine from what I can tell. WebRTC playback sync is complex and some of the knobs in frameworks like Pion are a little unintuitive; I'm not too familiar with aiortc at the moment.

src/scope/server/frame_processor.py
BuffMcBigHuge: I had struggled with this actually - I'll look into your suggestion.

PR #499, opened 2026-02-20
Plugin discovery in app

thomshutt:
Since we're seeing a lot of uptake of plugins, it makes sense to break them out to a top-level nav item rather than hiding within Settings.

Also implements plugin discovery within Scope, by calling the Daydream API.

Navbar:
Screenshot 2026-02-20 at 12 28 52

Dialog:
image

### DIFF
--- a/frontend/src/components/PluginsDialog.tsx
+++ b/frontend/src/components/PluginsDialog.tsx
@@ -293,7 +293,9 @@ export function PluginsDialog({
             <TabsContent value="discover" className="mt-0">
               <DiscoverTab
                 onInstall={handleInstallPlugin}
-                installedPluginNames={plugins.map(p => p.name)}
+                installedRepoUrls={plugins
+                  .map(p => p.package_spec)
+                  .filter((s): s is string => !!s)}
                 isInstalling={isInstalling}
               />
             </TabsContent>
--- a/frontend/src/components/settings/DiscoverTab.tsx
+++ b/frontend/src/components/settings/DiscoverTab.tsx
@@ -27,6 +27,7 @@ function GitHubIcon({ className }: { className?: string }) {
 interface DaydreamPlugin {
   id: string;
   creatorId: string;
+  creatorUsername: string;
   name: string;
   slug: string;
   description: string | null;
@@ -46,15 +47,23 @@ interface DiscoverResponse {
   hasMore: boolean;
 }
 
+function normalizeRepoUrl(url: string): string {
+  return url
+    .replace(/^git\+/, "")
+    .replace(/\.git$/, "")
+    .replace(/\/$/, "")
+    .toLowerCase();
+}
+
 interface DiscoverTabProps {
   onInstall: (packageSpec: string) => void;
-  installedPluginNames: string[];
+  installedRepoUrls: string[];
   isInstalling?: boolean;
 }
 
 export function DiscoverTab({
   onInstall,
-  installedPluginNames,
+  installedRepoUrls,
   isInstalling = false,
 }: DiscoverTabProps) {
   const [plugins, setPlugins] = useState<DaydreamPlugin[]>([]);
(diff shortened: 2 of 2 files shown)


frontend/src/components/settings/DiscoverTab.tsx
```diff
@@ -0,0 +1,256 @@
+import { useState, useEffect, useCallback } from "react";
+import { Download, ExternalLink, Search, Loader2 } from "lucide-react";
+import { Button } from "../ui/button";
+import { Input } from "../ui/input";
+import { Badge } from "../ui/badge";
+
+const DAYDREAM_API = "https://api.daydream.live";
+const DAYDREAM_APP = "https://app.daydream.live";
```
leszko: In `auth.ts` this is based in the env variable like:
```ts
const DAYDREAM_API_BASE =
  (import.meta.env.VITE_DAYDREAM_API_BASE as string | undefined) ||
  "https://api.daydream.live";
```

No strong feelings about it, but it would be nice to be consistent. So either change `auth.ts` or change it here.

frontend/src/components/settings/DiscoverTab.tsx
```diff
@@ -0,0 +1,256 @@
...
+    const timer = setTimeout(() => setDebouncedSearch(search), 300);
+    return () => clearTimeout(timer);
+  }, [search]);
+
+  const fetchDiscoverPlugins = useCallback(async () => {
+    setIsLoading(true);
+    setError(null);
+    try {
+      const params = new URLSearchParams({
+        limit: "50",
```
leszko, line 85: What if we have more than 50 entries? Should we have some pagination?

frontend/src/components/settings/DiscoverTab.tsx
```diff
@@ -0,0 +1,256 @@
...
+                    <Button
+                      variant="ghost"
+                      size="icon"
+                      className="h-8 w-8"
+                      title={`Install ${plugin.name}`}
+                      disabled={isInstalling}
+                      onClick={() => {
+                        const spec = plugin.repositoryUrl
+                          ? `git+${plugin.repositoryUrl}`
+                          : plugin.name;
```
leszko: Why do we use `plugin.name` if `repositoryUrl` is not specified. I guess that instead we should print an error.

frontend/src/components/PluginsDialog.tsx
```diff
@@ -0,0 +1,308 @@
...
+      );
+    } finally {
+      isModifyingPluginsRef.current = false;
+    }
+  };
+
+  return (
+    <Dialog open={open} onOpenChange={isOpen => !isOpen && onClose()}>
+      <DialogContent className="sm:max-w-[600px] p-0 gap-0">
+        <DialogHeader className="sr-only">
```
leszko: Do we need this `sr-only`? Do we even care about screen readers? 🙃

leszko (changes requested):
Nice work @thomshutt ! I really like how now the plugins start to be the central thing in Scope!

A few functional things (I tried to present in the demo recording):
1. When you make the window wider, it would be maybe nice to make the dialog with plugins wider as well, becuase it looks squeezed even if you have a big screen. The same actually with the height, it could use more of the screen.
2. The link from the plugin should go directly to the specific plugin, not the main plugin page
3. After the plugin is installed it should be displayed as "Installed"; Currently after the server refresh, nothing changes
4. Question: how to do you decide if the plugin is "tool" or "effect"? Do we have such distinction on the plugin page?

https://github.com/user-attachments/assets/31a182da-3ba9-47a4-9fac-fd068738cd52

frontend/src/components/settings/DiscoverTab.tsx
thomshutt: There is pagination, the 50 is actually specifying how many per page

frontend/src/components/PluginsDialog.tsx
thomshutt: This is just copied from the Settings dialog but you're right that we don't need it - I'll remove from both

frontend/src/components/settings/DiscoverTab.tsx
thomshutt: Good call, updated to be consistent with that

frontend/src/components/settings/DiscoverTab.tsx
thomshutt: Good catch, this should never happen but I've updated to not show the Download button if it does

PR #516, opened 2026-02-23
feat : expose temporal chunking metadata in pipeline schemas (api)

### LINKED ISSUE #470: Add frames_per_chunk and temporal_downsample_factor to pipeline schema metadata
Currently `frames_per_chunk` and `temporal_downsample_factor` are computed internally from pipeline block configs (`num_frame_per_block` × `vae_temporal_downsample_factor`) but aren't       surfaced through the `/api/v1/pipelines/schemas` endpoint.                                                                                                                                  
                                                                                                                                                                                              
  Exposing these in `get_schema_with_metadata()` would allow clients (e.g. ComfyUI nodepack) to display or use these values without hardcoding pipeline-specific constants.

  ### Suggested approach
  - Add `frames_per_chunk` and `temporal_downsample_factor` fields to the schema metadata dict returned by `BasePipelineConfig.get_schema_with_metadata()`
  - Derive values from the pipeline's existing config (e.g. `num_frame_per_block`, `vae_temporal_downsample_factor`

RUFFY-369:
**Fixes #470** 

This PR decouples external clients like comfyui by surfacing temporal chunking metadata directly via the /schemas API, enabling safe & dynamic buffer allocation at runtime.

**Validation**
- Verified the payload via direct REST queries against the uvicorn backend. AR diffusion models return their exact latent requirements, while spatial pre/post processors rightly return null

Screenshot from 2026-02-23 15-38-41

cc @ryanontheinside @leszko

### DIFF
--- a/frontend/src/components/Header.tsx
+++ b/frontend/src/components/Header.tsx
@@ -1,7 +1,8 @@
 import { useState, useEffect, useRef } from "react";
-import { Settings, Cloud, CloudOff } from "lucide-react";
+import { Settings, Cloud, CloudOff, Plug } from "lucide-react";
 import { Button } from "./ui/button";
 import { SettingsDialog } from "./SettingsDialog";
+import { PluginsDialog } from "./PluginsDialog";
 import { toast } from "sonner";
 import { useCloudStatus } from "../hooks/useCloudStatus";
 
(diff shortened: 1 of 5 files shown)


src/scope/core/pipelines/krea_realtime_video/schema.py
```diff
@@ -48,6 +48,8 @@ class KreaRealtimeVideoConfig(BasePipelineConfig):
     min_dimension = 16
     modified = True
     recommended_quantization_vram_threshold = 40.0
+    num_frame_per_block = 3
```
leszko, line 51: Now we have this info duplicated, it's here and also at `model.yaml`. We should either remove it from `model.yaml` or maybe just move everything that exists now in `model.yaml` into schema.yaml. The later would be my preference. @ryanontheinside Wdyt?

PR #518, opened 2026-02-23
01 feat: LoRA manifest tracking and multi-source downloader

ryanontheinside:
## Summary

- Adds a manifest system (`lora_manifest.json`) that tracks LoRA provenance (source, repo, version), SHA256 hashes, and metadata
- Adds a downloader supporting HuggingFace, CivitAI, and direct URL sources with hash verification
- Integrates manifest into all existing LoRA endpoints (list, install, delete)

### New endpoints
- `POST /api/v1/lora/download` — download from HF/CivitAI/URL with structured source fields and provenance tracking
- `PUT /api/v1/lora/{filename}/provenance` — retroactively tag existing LoRAs

### Updated endpoints (existing, modified in this PR)
- `GET /api/v1/loras` — response now includes `sha256` and `provenance` from manifest
- `POST /api/v1/loras` — now also writes manifest entry after install
- `DELETE /api/v1/loras/{name}` — now cleans up manifest entry

### Relationship to #442
PR #442 added the existing LoRA install-by-URL UI and `POST /api/v1/loras` endpoint. This PR extends that endpoint with manifest tracking and adds the richer `POST /api/v1/lora/download` API (structured source fields, provenance, SHA256 verification). A frontend follow-up is needed to integrate the new download API into the LoRAsTab UI — e.g., parsing pasted HF/CivitAI URLs into structured requests and displaying provenance metadata in the LoRA list.

### Note
CivitAI public models download without an API token. Gated/early-access models require a user-provided token via `CIVITAI_API_TOKEN` env var or stored in Settings > API Keys.

## Test plan
- [x] `uv run pytest tests/test_lora_manifest.py tests/test_lora_downloader.py` — 16 unit tests pass
- [x] `uv run daydream-scope` — server starts without errors
- [x] `uv run python scripts/test_lora_download.py` — real downloads from HF and CivitAI verified
- [x] Verified `lora_manifest.json` created with correct entries after download
- [x] Verified `GET /api/v1/loras` returns provenance and sha256 data for manifested LoRAs
- [x] Verified CivitAI public model downloads work without API token

### DIFF
--- a/src/scope/server/app.py
+++ b/src/scope/server/app.py
@@ -1187,7 +1187,9 @@ async def tag_lora_provenance(
 ) -> LoRAManifestEntry:
     """Retroactively tag a local LoRA with provenance info."""
     lora_dir = get_lora_dir()
-    file_path = lora_dir / filename
+    file_path = (lora_dir / filename).resolve()
+    if not file_path.is_relative_to(lora_dir.resolve()):
+        raise HTTPException(status_code=400, detail="Invalid filename")
     if not file_path.exists():
         raise HTTPException(status_code=404, detail=f"LoRA file '{filename}' not found")
 
--- a/src/scope/server/lora_downloader.py
+++ b/src/scope/server/lora_downloader.py
@@ -160,7 +160,9 @@ async def download_lora(
         dest_dir = lora_dir / request.subfolder
     else:
         dest_dir = lora_dir
-    dest_path = dest_dir / filename
+    dest_path = (dest_dir / filename).resolve()
+    if not dest_path.is_relative_to(lora_dir.resolve()):
+        raise ValueError("Invalid destination path")
 
     # Download
     await loop.run_in_executor(None, http_get, url, dest_path)


src/scope/server/app.py
```diff
@@ -1070,6 +1133,40 @@ async def delete_lora_file(
         raise HTTPException(status_code=500, detail=str(e)) from e
 
 
+@app.post("/api/v1/lora/download")
```
leszko, line 1166: Is this endpoind to download LoRA or to download LoRA manifest? I think it's a little confusing, because we already have some other logic for downloading LoRAs and it's used in UI. 

If it's downloading lora, then I think we should have only 1 way to download loras, we already download lora in `/api/v1/loras`. If it's downloading lora manifest, then I'd update the endpoint name and the related functions to be clear about it (then also change it in the PR Description, because it's confusing).

src/scope/server/app.py
```diff
@@ -989,15 +1025,35 @@ async def install_lora_file(
...
+
+        source: str
+        if is_civitai:
+            source = "civitai"
+        elif hostname == "huggingface.co" or hostname.endswith(".huggingface.co"):
+            source = "huggingface"
+        else:
+            source = "url"
+
+        provenance = LoRAProvenance(source=source, url=url)
```
leszko: I think the `url` may contain the API TOKEN for CivitAI, because we attach it earlier in `?token={api-key}`. This is problematic, because now we'll store that token in the manifest entry.

leszko (commented):
@ryanontheinside I guess you checked these scenarios, but just to double-check:
1. Version does not match sha256 => What happens then?
2. There is no `version_id`, but there is sha256, but the latest has changed. What should we do now? This is actually what I got in my environment (please check JSON below).
3. A version was removed from HF or CivitAI

```json
        {
          "id": null,
          "filename": "daydream-scope-dissolve.safetensors",
          "weight": 1.0,
          "merge_mode": "permanent_merge",
          "provenance": {
            "source": "civitai",
            "repo_id": null,
            "hf_filename": null,
            "model_id": null,
            "version_id": null,
            "url": "https://civitai.com/api/download/models/2680702"
          },
          "expected_sha256": "fd373e0991a33df28f6d0d4a13d8553e2c9625483e309e8ec952a96a2570bec9"
        }
```

ryanontheinside:
> 1. Version does not match sha256 => What happens then?

The file is deleted and an error is raised.

> 2. There is no `version_id`, but there is sha256, but the latest has changed.

Expected for LoRAs installed through the old `POST /api/v1/loras` endpoint, which only has a URL to work with. The new `POST /api/v1/lora/download` accepts structured fields (`model_id`, `version_id`). A frontend follow-up will wire the UI to the new endpoint.

> 3. A version was removed from HF or CivitAI

If the file is already local, it works fine. On a fresh machine, the resolver reports it as missing, and the download attempt fails with an HTTP error. No special handling needed since both cases present identically (file not on disk).

src/scope/server/app.py
ryanontheinside: These serve different roles. `POST /api/v1/loras` is the existing URL-based install used by the current UI. `POST /api/v1/lora/download` accepts structured source fields (HF `repo_id`, CivitAI `version_id`, etc.) and is what the workflow import system uses. The old endpoint was enhanced with manifest writes so existing installs also get tracked, but the richer API is needed for reliable re-downloading from provenance. A frontend follow-up will migrate the UI to the new endpoint.

src/scope/server/app.py
```diff
@@ -1070,6 +1147,40 @@ async def delete_lora_file(
...
+
+
+@app.put("/api/v1/lora/{filename:path}/provenance")
+async def tag_lora_provenance(
+    filename: str,
+    provenance: LoRAProvenance,
+) -> LoRAManifestEntry:
+    """Retroactively tag a local LoRA with provenance info."""
+    lora_dir = get_lora_dir()
+    file_path = lora_dir / filename
```
leszko:
```suggestion
  file_path = (lora_dir / filename).resolve()
  if not file_path.is_relative_to(lora_dir.resolve()):
      raise HTTPException(status_code=400, detail="Invalid filename")
```

I'd add this check. Otherwise someone using the endpoint could try to hack some random directory like `/etc/passwd` by passing it as a path parameter.

src/scope/server/lora_downloader.py
```diff
@@ -0,0 +1,195 @@
...
+    if not filename:
+        filename = await loop.run_in_executor(None, _filename_from_response, url)
+
+    # Build destination path
+    if request.subfolder:
+        dest_dir = lora_dir / request.subfolder
+    else:
+        dest_dir = lora_dir
+    dest_path = dest_dir / filename
+
```
leszko, line 166: The same a before, we should check if the `dest_pat` is actually something we expect it to be. Otherwise user may inject some random directory. I'm not too concerned about the local runs, but since we run it as well in the cloud we need this kind-of protections.

```suggestion
 dest_path = dest_path.resolve()
  if not dest_path.is_relative_to(lora_dir.resolve()):
      raise ValueError("Invalid destination path")
```

PR #522, opened 2026-02-23
02 feat: add workflow schema and export API

ryanontheinside:
## Summary
- Introduces `schema.py`: Pydantic models for `.scope-workflow.json`, the long-term contract for shareable workflows
- Adds `__init__.py` re-exports for all schema types
- Adds `GET /api/v1/workflow/schema` (returns JSON Schema) and `POST /api/v1/workflow/validate` (validates a document against the schema)
- Renames `PluginManager._list_plugins_sync` to `list_plugins_sync` (public API, 5 internal callers)

**Depends on #518** (LoRA manifest & downloader)

## Design decisions
- All schema models use `extra="ignore"` for forward compatibility (including `WorkflowLoRAProvenance` subclass)
- Pipeline source type is `"builtin" | "pypi" | "git" | "local"` for actionable import resolution
- `pipeline_version` is optional (builtins ship with Scope, so version mismatch is handled by `min_scope_version` in #526)
- Optional `WorkflowLoRA.id` field reserved for future graph-mode per-node LoRA assignment
- Serialization uses `exclude_none=True` to keep exported JSON clean

## Test plan
- [x] `uv run pytest tests/test_workflow_schema.py -v` -- 38 tests covering:
  - Schema round-trip serialization
  - Forward compatibility (unknown fields at every nesting level)
  - Minimal document validation
  - Negative validation (invalid source type, missing required fields)
  - Serialization stability (exact key sets, JSON file round-trip)
  - Provenance subclass behavior
  - FastAPI endpoint tests (schema + validate)
- [x] `uv run pytest tests/test_plugin_manager.py -v` -- 99 tests pass (rename verification)


src/scope/server/app.py
```diff
@@ -1167,6 +1167,44 @@ async def tag_lora_provenance(
...
+class WorkflowExportRequest(BaseModel):
+    name: str
+    description: str = ""
+    author: str = ""
+    frontend_params: dict[str, dict] | None = None
+    """Per-pipeline frontend params, keyed by pipeline_id."""
+
+
+@app.post("/api/v1/workflow/export")
+async def export_workflow(
```
leszko:
```suggestion
def export_workflow(
```

I think that if you use `async`, then it will block the main event loop.

src/scope/server/app.py
```diff
@@ -1167,6 +1167,44 @@ async def tag_lora_provenance(
     return add_manifest_entry(lora_dir, filename, provenance, sha256, size_bytes)
 
 
+# ---------------------------------------------------------------------------
+# Workflow export
+# ---------------------------------------------------------------------------
+
+
+class WorkflowExportRequest(BaseModel):
+    name: str
```
leszko: What do you think about having a default here as well? 

I can imagine someone calling the endpoint, even from the API, and not care about the name, so it would make their life simpler.

Can be `name: str = ""` or `name: str = "Untitled Workflow"`.

src/scope/server/app.py
```diff
@@ -1167,6 +1167,44 @@ async def tag_lora_provenance(
...
+    description: str = ""
+    author: str = ""
+    frontend_params: dict[str, dict] | None = None
+    """Per-pipeline frontend params, keyed by pipeline_id."""
+
+
+@app.post("/api/v1/workflow/export")
```
leszko: What about the cloud support for this? 

CC: @mjh1

src/scope/core/workflows/schema.py
```diff
@@ -0,0 +1,77 @@
...
+    params: dict[str, Any] = {}
+
+
+class ScopeWorkflow(BaseModel, extra="ignore"):
+    """Root schema for a ``.scope-workflow.json`` file."""
+
+    format: Literal["scope-workflow"] = "scope-workflow"
+    format_version: str = WORKFLOW_FORMAT_VERSION
+    metadata: WorkflowMetadata
+    pipelines: list[WorkflowPipeline]
```
leszko, line 75: If it's long-term contract, I think this will not survive moving to non-linear (graph) mode. How do you think we should approach it?

src/scope/core/workflows/schema.py
```diff
@@ -0,0 +1,77 @@
...
+    manifest model strict.
+    """
+
+
+class WorkflowMetadata(BaseModel, extra="ignore"):
+    """Authorship and tooling metadata."""
+
+    name: str
+    description: str = ""
+    author: str = ""
```
leszko: I wonder if we want to have all these 3 fields. In some sense, it's simpler to add later fields than to make some fields deprecated. As a rule of thumb right now, I'd say that if we only have an idea how it will look like in UI, then it's fine to keep them. If not, then I'd start with fields that we need right now.

src/scope/core/workflows/schema.py
```diff
@@ -0,0 +1,77 @@
...
+    scope_version: str
+
+
+class WorkflowPipelineSource(BaseModel, extra="ignore"):
+    """Where a pipeline comes from."""
+
+    type: Literal["builtin", "pypi", "git", "local"]
+    plugin_name: str | None = None
+    plugin_version: str | None = None
+    package_spec: str | None = None
```
leszko, line 45: Nit: How about making some fields optional, I found it less readable with all these `null` values in the generated JSON.

```json
      "source": {
        "type": "builtin",
        "plugin_name": null,
        "plugin_version": null,
        "package_spec": null
      },
```

src/scope/core/workflows/schema.py
```diff
@@ -0,0 +1,77 @@
...
+
+class WorkflowLoRA(BaseModel, extra="ignore"):
+    """A LoRA adapter used by the workflow."""
+
+    id: str | None = None
+    filename: str
+    weight: float = 1.0
+    merge_mode: str = "permanent_merge"
+    provenance: WorkflowLoRAProvenance | None = None
+    expected_sha256: str | None = None
```
leszko: Nit: maybe just `sha256`?

src/scope/core/workflows/schema.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class WorkflowPipeline(BaseModel, extra="ignore"):
+    """A single pipeline within the workflow."""
+
+    pipeline_id: str
+    pipeline_version: str
```
leszko:
```json
    {
      "pipeline_id": "video-depth-anything",
      "pipeline_version": "1.0.0",
      "source": {
        "type": "builtin",
        "plugin_name": null,
        "plugin_version": null,
        "package_spec": null
      },
```

Where does the version come from for the built-in pipelines? In general, I wonder if we should even include the version for built-in pipelines, they come with the scope tool, so what should happen if there is a version mismatch?

src/scope/core/workflows/schema.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class WorkflowMetadata(BaseModel, extra="ignore"):
+    """Authorship and tooling metadata."""
+
+    name: str
+    description: str = ""
+    author: str = ""
+    created_at: datetime
+    scope_version: str
```
leszko, line 36: How do we plan to use the `scope_version`?

src/scope/core/workflows/export.py
leszko, line 1: There is something wrong with the export. I've created a workflow with: 
- video-depth - anything preprocessor
- VACE ON, Input video ON
- main pipeline - longlive

Here's what I have generated:
```json
{
  "format": "scope-workflow",
  "format_version": "1.0",
  "metadata": {
    "name": "My Workflow",
    "description": "Exported session",
    "author": "Me",
    "created_at": "2026-02-26T08:58:19.958641Z",
    "scope_version": "0.1.4"
  },
  "pipelines": [
    {
      "pipeline_id": "video-depth-anything",
      "pipeline_version": "1.0.0",
      "source": {
        "type": "builtin",
        "plugin_name": null,
        "plugin_version": null,
        "package_spec": null
      },
      "loras": [
        {
          "id": null,
          "filename": "daydream-scope-dissolve.safetensors",
          "weight": 1.0,
          "merge_mode": "permanent_merge",
          "provenance": {
            "source": "civitai",
            "repo_id": null,
            "hf_filename": null,
            "model_id": null,
            "version_id": null,
            "url": "https://civitai.com/api/download/models/2680702"
          },
          "expected_sha256": "fd373e0991a33df28f6d0d4a13d8553e2c9625483e309e8ec952a96a2570bec9"
        }
      ],
      "params": {
        "height": 512,
        "width": 512,
        "quantization": null,
        "vace_enabled": true
      }
    }
  ]
}
```

**Comments**
1. There is no longlive pipeline
2. Loras are embedded into video-depth-anything
3. There is nothing about the input mode; also nothing about the prompt

src/scope/core/workflows/export.py
```diff
@@ -0,0 +1,154 @@
...
+)
+
+logger = logging.getLogger(__name__)
+
+# Maps plugin_manager source strings to WorkflowPipelineSource.type values.
+_SOURCE_TYPE_MAP: dict[str, str] = {
+    "pypi": "pypi",
+    "git": "git",
+    "local": "local",
+}
```
leszko: Do we need it? Don't we actually always have `pypi` right now?

src/scope/core/workflows/export.py
```diff
@@ -0,0 +1,154 @@
...
+        top-level ``WorkflowLoRA`` list).
+    """
+    from scope.core.lora.manifest import load_manifest
+    from scope.core.pipelines.registry import PipelineRegistry
+
+    manifest = load_manifest(lora_dir)
+
+    plugins = get_plugin_list(plugin_manager)
+
+    snapshot = pipeline_manager.get_load_snapshot()
```
leszko: I think that getting these should actually happen in the frontend, not in the backend. I think the whole export should basically be:
1. Get everything from the frontend
2. Only enrich the data from the backend when really needed; e.g. provenance for Lora, plugin versions, etc.

The main reasoning is that I think treating backend as the source of truth will result a lot of problems, because we update backend during the stream start operation. So, this is what can happen:
1. User starts longlive, user starts the stream
2. User stops longlive
3. User selects krea, specifies some params
4. User clicks export

I think the expectation would be that user gets workflow they current see, not the last loaded workflow.

leszko (changes requested):
Added inline comments, but the main things to think about:
1. Pipeline schema, this will be hard to change later
2. Export happening in the frontend and only enrich it in the backend

src/scope/server/app.py
ryanontheinside: Good catch, no reason for this to be async. Addressed in PR #525 but can change it here as well since this makes no sense

src/scope/server/app.py
ryanontheinside: yea i think untitled workflow will be good

src/scope/server/app.py
ryanontheinside: the refactor to front end driven export will largely address this

src/scope/core/workflows/schema.py
ryanontheinside: Let's revisit after the refactor to frontend-driven export, since the schema will naturally follow whatever topology the frontend represents

src/scope/core/workflows/schema.py
ryanontheinside: good call

src/scope/core/workflows/schema.py
ryanontheinside: Agreed

src/scope/core/workflows/schema.py
ryanontheinside: #526  adds min_scope_version. scope_version is meant to be metadata for diagnostics, and min_scope_version is meant to be a compatibility gate

src/scope/core/workflows/export.py
ryanontheinside: i will revisit this after the front end export refactor as i think they will naturally resolve

ryanontheinside: though this is fixed in a later PR at the moment - it should not be this fragile

src/scope/core/workflows/export.py
ryanontheinside: I think installing from local paths and git URLs is definitely something we should support, even if we encourage pypi. As a developer, I'll be saving workflows that I never share with anyone but myself using lots of forks of plugins for example.

src/scope/core/workflows/export.py
ryanontheinside: Agreed. This should have been the obvious thing to do, and it will probably resolve most of your comments.

src/scope/core/workflows/schema.py
ryanontheinside: this feels like we cant solve for this without graph mode implemented, but we do have format_version which gives room to evolve. wdyt?

leszko: I think it's ok. The only drawback I see is that if we will want to support old workflows exported, then we'll need to implement this compatibility handling logic. But I think it's ok. I share your thinking that it's hard to design now for Graph Mode that will be in the future.

src/scope/server/app.py
```diff
@@ -1181,6 +1181,49 @@ async def tag_lora_provenance(
...
+class ExportPipelineInput(BaseModel):
+    pipeline_id: str
+    params: dict = {}
+    loras: list[ExportLoRAInput] = []
+
+
+class WorkflowExportRequest(BaseModel):
+    name: str = "Untitled Workflow"
+    pipelines: list[ExportPipelineInput]
+    lora_merge_mode: str = "permanent_merge"
```
leszko: Do we need this `lora_merge_mode` passed as a param? This seems like something we should get from inside the pipeline, from `ExportLoRAInput`.

src/scope/core/workflows/export.py
leszko: ok, that's fine. But if it's fixed in the other, then let's not merge this PR to `main` before the others.

PR #531, opened 2026-02-24
Add Syphon input support for MacOS

thomshutt:
Only shows up for users on MacOS. Tested with the Syphon test tool and TouchDesigner.

image

### DIFF
--- a/frontend/src/lib/api.ts
+++ b/frontend/src/lib/api.ts
@@ -319,7 +319,7 @@ export const getInputSourceResolution = async (
 export const getInputSourceStreamUrl = (
   sourceType: string,
   identifier: string,
-  fps = 15
+  fps = 2
 ): string =>
   `/api/v1/input-sources/${sourceType}/sources/${encodeURIComponent(identifier)}/stream?fps=${fps}`;
 
--- a/src/scope/server/app.py
+++ b/src/scope/server/app.py
@@ -1493,7 +1493,7 @@ def get_input_source_resolution(
 
 @app.get("/api/v1/input-sources/{source_type}/sources/{identifier:path}/stream")
 async def stream_input_source_preview(
-    source_type: str, identifier: str, fps: int = Query(15, ge=1, le=30)
+    source_type: str, identifier: str, fps: int = Query(2, ge=1, le=30)
 ):
     """MJPEG stream of an input source for live preview."""
     source_class = _resolve_input_source_class(source_type)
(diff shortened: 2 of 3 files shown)


frontend/src/lib/api.ts
```diff
@@ -319,7 +319,7 @@ export const getInputSourceResolution = async (
 export const getInputSourceStreamUrl = (
   sourceType: string,
   identifier: string,
-  fps = 2
+  fps = 15
```
leszko: What is this change for?

src/scope/server/app.py
```diff
@@ -1493,7 +1493,7 @@ def get_input_source_resolution(
 
 @app.get("/api/v1/input-sources/{source_type}/sources/{identifier:path}/stream")
 async def stream_input_source_preview(
-    source_type: str, identifier: str, fps: int = Query(2, ge=1, le=30)
+    source_type: str, identifier: str, fps: int = Query(15, ge=1, le=30)
```
leszko: Won't it be a problem with the NDI support input preview? Aren't there any performance concerns about it? CC: @gioelecerati

gioelecerati: Yeah I specifically set it at 2 fps because having it even at 10 FPS was consuming too much resources on lower end machines or non gpu machines

It's fine for battle machines with a GPU, but if you're running remote inference on a macbook, having it at 15fps will risk to cause performance issues

gioelecerati: Note: This might be valid for NDI and not Syphon. I suspect the main issue with NDI was spinning up a separate receiver

src/scope/core/outputs/spout.py
```diff
@@ -12,15 +12,15 @@
 
 
 class SpoutOutputSink(OutputSink):
-    """Output sink that sends video frames via Spout (Windows) or Syphon (macOS).
+    """Output sink that sends video frames via Spout on Windows.
```
gioelecerati, line 15: Will the output sink for Syphon be implemented in a follow up PR?

src/scope/server/app.py
thomshutt: Ah okay, I just tested locally on my macbook and it seemed fine, so I'll split it to use different values based on input source

src/scope/core/outputs/spout.py
thomshutt: Yep!

src/scope/server/app.py
thomshutt: Done!

PR #540, opened 2026-02-26
feat: add support for OSC

vcashwin:
## Architecture

The OSC server is a standalone background service (like Kafka publisher) managed in the FastAPI lifespan. It listens on a UDP port, maps incoming OSC addresses to Scope parameter updates, and pushes them into the existing frame_processor.update_parameters() path -- no new pipeline plumbing needed.

image

## API Endpoints

```
GET /api/v1/osc/status -- Returns {"enabled": bool, "port": int, "listening": bool}

POST /api/v1/osc/config -- Start/stop OSC server or change port at runtime
```

image

### DIFF
--- a/frontend/src/components/OutputsPanel.tsx
+++ b/frontend/src/components/OutputsPanel.tsx
@@ -1,5 +1,7 @@
 import { Card, CardContent, CardHeader, CardTitle } from "./ui/card";
+import { Toggle } from "./ui/toggle";
 import { OutputSinkToggle } from "./OutputSinkToggle";
+import { LabelWithTooltip } from "./ui/label-with-tooltip";
 import { PARAMETER_METADATA } from "../data/parameterMetadata";
 import type { SettingsState } from "../types";
 
@@ -13,6 +15,8 @@ interface OutputsPanelProps {
   spoutAvailable?: boolean;
   ndiAvailable?: boolean;
   syphonAvailable?: boolean;
+  oscEnabled?: boolean;
+  onOscToggle?: (enabled: boolean) => void;
   isStreaming?: boolean;
 }
 
@@ -23,6 +27,8 @@ export function OutputsPanel({
   spoutAvailable = false,
   ndiAvailable = false,
   syphonAvailable = false,
+  oscEnabled = false,
+  onOscToggle,
   isStreaming = false,
 }: OutputsPanelProps) {
   return (
@@ -72,6 +78,25 @@ export function OutputsPanel({
             isStreaming={isStreaming}
           />
         )}
+
+        <div className="space-y-1">
+          <div className="flex items-center justify-between gap-2">
+            <LabelWithTooltip
+              label="OSC Control"
+              tooltip="Open Sound Control server for external parameter control from TouchDesigner, Resolume, Max/MSP, etc."
+              className="text-sm font-medium"
+            />
+            <Toggle
+              pressed={oscEnabled}
+              onPressedChange={onOscToggle ?? (() => {})}
+              variant="outline"
+              size="sm"
+              className="h-7"
+            >
+              {oscEnabled ? "ON" : "OFF"}
+            </Toggle>
+          </div>
+        </div>
       </CardContent>
     </Card>
   );
--- a/frontend/src/components/SettingsPanel.tsx
+++ b/frontend/src/components/SettingsPanel.tsx
@@ -314,8 +314,6 @@ interface SettingsPanelProps {
   inputMode?: InputMode;
   // Whether this pipeline supports noise controls in video mode (schema-derived)
   supportsNoiseControls?: boolean;
-  // OSC
-  oscEnabled?: boolean;
   // VACE settings
   vaceEnabled?: boolean;
   onVaceEnabledChange?: (enabled: boolean) => void;
@@ -383,7 +381,6 @@ export function SettingsPanel({
   loraMergeStrategy = "permanent_merge",
   inputMode,
   supportsNoiseControls = false,
-  oscEnabled = false,
   vaceEnabled = true,
   onVaceEnabledChange,
   vaceUseInputVideo = true,
@@ -1188,23 +1185,6 @@ export function SettingsPanel({
             </>
           );
         })()}
-
-        {/* OSC Status */}
-        {oscEnabled && (
-          <div className="space-y-1">
-            <div className="flex items-center justify-between gap-2">
-              <LabelWithTooltip
-                label="OSC Control"
-                tooltip="Open Sound Control server for external parameter control from TouchDesigner, Resolume, Max/MSP, etc."
-                className="text-sm font-medium"
-              />
-              <span className="inline-flex items-center gap-1.5 text-xs text-emerald-500 font-medium">
-                <span className="h-1.5 w-1.5 rounded-full bg-emerald-500" />
-                Listening
-              </span>
-            </div>
-          </div>
-        )}
       </CardContent>
     </Card>
   );
--- a/frontend/src/pages/StreamPage.tsx
+++ b/frontend/src/pages/StreamPage.tsx
@@ -33,7 +33,7 @@ import type {
   DownloadProgress,
 } from "../types";
 import type { PromptItem, PromptTransition } from "../lib/api";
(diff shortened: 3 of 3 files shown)


docs/osc.md
```diff
@@ -0,0 +1,180 @@
+# OSC (Open Sound Control) Integration
+
+Scope includes an optional OSC server that allows external applications to control pipeline parameters in real-time over UDP. This enables integration with creative tools like TouchDesigner, Resolume, Max/MSP, Ableton Live, and hardware MIDI controllers.
+
+## Prerequisites
+
+Install the `python-osc` dependency:
+
+```bash
+uv pip install python-osc
```
leszko: Since `python-osc` if very small, could we just install and start it by default? I guess there is no need for the `extra`.

docs/osc.md
```diff
@@ -0,0 +1,180 @@
...
+```bash
+uv sync --extra osc
+```
+
+## Enabling the OSC Server
+
+### CLI Flag
+
+```bash
+uv run daydream-scope --osc
```
leszko: I think there should not be a need to pass the `osc` argument. Otherwise, it won't work in the Electron App.

docs/osc.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Runtime API
+
+You can also start/stop the OSC server at runtime without restarting Scope:
+
+```bash
+# Start
+curl -X POST http://localhost:8000/api/v1/osc/config \
+  -H "Content-Type: application/json" \
+  -d '{"enabled": true, "port": 9000}'
```
leszko: Shouldn't it be the only option to start OSC and we can have a toggle in UI for this?

src/scope/server/osc.py
```diff
@@ -0,0 +1,262 @@
...
+
+        dispatcher.map("/scope/prompt", self._handle_prompt)
+        dispatcher.map("/scope/prompt/weight", self._handle_prompt_weight)
+        dispatcher.map("/scope/noise", self._handle_noise)
+        dispatcher.map("/scope/denoise", self._handle_denoise)
+        dispatcher.map("/scope/cache/reset", self._handle_cache_reset)
+        dispatcher.map("/scope/cache/bias", self._handle_cache_bias)
+        dispatcher.map("/scope/output/*/enable", self._handle_output_enable)
+        dispatcher.map("/scope/output/*/disable", self._handle_output_disable)
+        dispatcher.map("/scope/param/*", self._handle_generic_param)
```
leszko: Could we just have endpoint instead and pass which parameter to modify as a parameter?

docs/osc.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Enabling the OSC Server
+
+### CLI Flag
+
+```bash
+uv run daydream-scope --osc
+```
+
+With a custom port (default is 9000):
```
leszko: Do we actually need a custom port? Can't we use the same port as scope?

docs/osc.md
```diff
@@ -0,0 +1,180 @@
...
+|---|---|---|
+| `/scope/prompt` | `<string>` | Set the first prompt text |
+| `/scope/prompt/weight` | `<float 0-100>` | Set first prompt weight |
+| `/scope/noise` | `<float 0.0-1.0>` | Set noise scale |
+| `/scope/denoise` | `<int> [<int>...]` | Set denoising step list |
+| `/scope/cache/reset` | *(none)* | Trigger a cache reset |
+| `/scope/cache/bias` | `<float 0.01-1.0>` | Set KV cache attention bias |
+| `/scope/output/<type>/enable` | `[<string name>]` | Enable an output sink (e.g., `/scope/output/ndi/enable`) |
+| `/scope/output/<type>/disable` | *(none)* | Disable an output sink |
+| `/scope/param/<key>` | `<value>` | Generic parameter passthrough |
```
leszko: Params should not be hardcoded, users should be able to modify any arbitrary parameters.

docs/osc.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Generic Parameters
+
+The `/scope/param/<key>` address is a catch-all that lets you set any pipeline parameter by name. The `<key>` becomes the parameter name and the argument becomes the value. This means new parameters are automatically controllable via OSC without code changes.
```
leszko: I think this is the only thing that we need . A generic endpoint which covers all params. We could also consider adding pipeline_id as the optional path param (but can add it later in the follow-up PR).

src/scope/server/osc.py
```diff
@@ -0,0 +1,262 @@
...
+        self._webrtc_manager = webrtc_manager
+
+        try:
+            from pythonosc.osc_server import AsyncIOOSCUDPServer
+
+            dispatcher = self._setup_dispatcher()
+            self._server = AsyncIOOSCUDPServer(
+                ("0.0.0.0", self.port),
+                dispatcher,
+                asyncio.get_event_loop(),
```
leszko: Nit: This is deprecated. Use `asyncio.get_running_loop` instead.

src/scope/server/osc.py
```diff
@@ -0,0 +1,262 @@
...
+            return False
+
+        self._webrtc_manager = webrtc_manager
+
+        try:
+            from pythonosc.osc_server import AsyncIOOSCUDPServer
+
+            dispatcher = self._setup_dispatcher()
+            self._server = AsyncIOOSCUDPServer(
+                ("0.0.0.0", self.port),
```
leszko: I think it should bind to the same `host` as specified in `app.py`. Not always `0.0.0.0`.

docs/osc.md
vcashwin: Added a default

docs/osc.md
vcashwin: Removed any CLI based toggle, controllable via config API now

src/scope/server/osc.py
vcashwin: done

docs/osc.md
vcashwin: Had it earlier to remove any confusion since I was newly exploring this protocol. Should be fine to run on the same port. Updated.

docs/osc.md
vcashwin: Added a generic API that can take in params.

src/scope/server/osc.py
vcashwin: Done.

src/scope/server/osc.py
vcashwin: Added a global var for port and passing it along here now.

docs/osc.md
vcashwin: Done.

src/scope/server/app.py
```diff
@@ -272,7 +280,12 @@ async def lifespan(app: FastAPI):
     from .webrtc import WebRTCManager
 
     # Startup
-    global webrtc_manager, pipeline_manager, cloud_connection_manager, kafka_publisher
+    global \
+        webrtc_manager, \
+        pipeline_manager, \
+        cloud_connection_manager, \
+        kafka_publisher, \
+        osc_manager
```
leszko, line 288: Why this change?

frontend/src/lib/api.ts
```diff
@@ -208,8 +208,36 @@ export interface HardwareInfoResponse {
...
+  osc_enabled: boolean;
 }
 
+export interface OSCStatusResponse {
+  enabled: boolean;
+  port: number;
+  listening: boolean;
+}
+
+export const getOSCStatus = async (): Promise<OSCStatusResponse> => {
```
leszko, line 220: Is this used anywhere? If not, please remove.

src/scope/server/osc.py
```diff
@@ -0,0 +1,136 @@
...
+    def _get_active_frame_processors(self) -> list:
+        """Get frame processors from all active WebRTC sessions."""
+        if not self._webrtc_manager:
+            return []
+
+        processors = []
+        for session in self._webrtc_manager.sessions.values():
+            if (
+                session.video_track
+                and hasattr(session.video_track, "frame_processor")
```
leszko, line 55: I think you also need to check fot the `cloud_track` so that it works in the cloud.

PR #541, opened 2026-02-26
On demand plugins

mjh1:
### Summary

- **Add a dedicated cloud Docker image** (`Dockerfile.cloud`) and update CI to build both standard and cloud variants using a matrix strategy, with `-cloud` suffixed image tags
  - Main reason was to speed up cloud load times, i found that it was faster to not install any of the dependencies into the image and download them at runtime instead. They total about 8gb so it really slows down the docker image pull (and github build times). I also omit the frontend from this image.
- **Enable whitelisted plugin installation in cloud mode** — instead of blocking all plugin installs, plugins are now allowed if they match an entry in the `CLOUD_PLUGINS_WHITELIST` environment variable (with URL normalization for flexible matching)
- **Proxy plugin API endpoints through the cloud WebSocket** — the `/health`, `/api/v1/restart`, and all `/api/v1/plugins` endpoints now use the `@cloud_proxy()` decorator so the frontend can manage plugins on the remote cloud instance
- **Clean up plugins between sessions** — on WebSocket disconnect, all installed plugins are uninstalled and session data is wiped, with a gate (`_cleanup_event`) that blocks the next connection's "ready" message until cleanup finishes
- **Harden cloud subprocess environment** — the scope subprocess now receives only whitelisted environment variables instead of inheriting the full parent environment

### Changes

| File | What changed |
|---|---|
| `Dockerfile.cloud` | New cloud-specific Dockerfile based on `nvidia/cuda:12.8.0-cudnn-runtime-ubuntu24.04` |
| `.github/workflows/docker-build.yml` | Matrix strategy for `standard` / `cloud` builds with per-variant tags, dockerfiles, and GHA cache scopes |
| `src/scope/cloud/fal_app.py` | Whitelist-based plugin install gating, plugin cleanup on disconnect, env var whitelisting, `uv run --extra kafka` instead of pre-install, assets dir moved to `/tmp/` |
| `src/scope/server/app.py` | Added `@cloud_proxy()` to health, restart, and all plugin CRUD endpoints; renamed `request` -> `http_request` to avoid shadowing |

Screenshot 2026-02-26 at 18 22 42

You can now connect to cloud and go to the plugin settings discover tab and click install, same as the local experience.

### DIFF
--- a/.github/workflows/fal-deploy.yml
+++ b/.github/workflows/fal-deploy.yml
@@ -8,9 +8,9 @@ on:
         required: true
         type: choice
         options:
-          - staging
+          - main
           - prod
-        default: 'staging'
+        default: 'main'
   workflow_run:
     workflows: ["Build and Push Docker Image"]
     types:
@@ -48,13 +48,13 @@ jobs:
           # Use the SHA from the triggering workflow for workflow_run events
           SHA="${{ github.event.workflow_run.head_sha || github.sha }}"
           SHORT_SHA=$(echo "$SHA" | cut -c1-7)
-          
-          # Use input environment for manual dispatch, default to staging for workflow_run
+
+          # Use input environment for manual dispatch, default to main for workflow_run
           if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
             ENV="${{ github.event.inputs.environment }}"
           else
-            ENV="staging"
+            ENV="main"
           fi
-          
+
           echo "Deploying to $ENV with tag $SHORT_SHA"
           SCOPE_DEPLOY_TAG=$SHORT_SHA fal deploy --env $ENV --auth public src/scope/cloud/fal_app.py
--- a/src/scope/cloud/fal_app.py
+++ b/src/scope/cloud/fal_app.py
@@ -12,8 +12,10 @@ Based on:
 import asyncio
 import json
 import os
+import queue
 import shutil
 import subprocess as _subprocess
+import threading
 import time
 import uuid
 from typing import Any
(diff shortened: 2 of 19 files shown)


src/scope/cloud/fal_app.py
```diff
@@ -275,37 +352,53 @@ def setup(self):
...
-                check=True,
-                env=scope_env,
-            )
-            print("✅ daydream-scope[kafka] installed")
-        except Exception as e:
-            print(f"Failed to install daydream-scope[kafka]: {e}")
+        scope_env["UV_CACHE_DIR"] = "/tmp/uv-cache"
 
         # Start the scope server in a background thread
+        # TODO disable plugin install while stream running?
```
mjh1: i noticed that it was still possible to install a plugin while streaming which would reload server and break the stream, I think this is probably an existing edge case but we can probably fix easily in the front end

src/scope/cloud/fal_app.py
```diff
@@ -604,14 +674,52 @@ async def handle_api_request(payload: dict):
...
+
+                # Check if the requested package is in the cloud plugins whitelist
+                def normalize_plugin_url(url: str) -> str:
+                    """Normalize a plugin URL for comparison."""
+                    import re
+
+                    normalized = url.lower().strip()
+                    # Remove URL protocols
+                    normalized = re.sub(r"^git\+https?://", "", normalized)
+                    normalized = re.sub(r"^https?://", "", normalized)
```
leszko, line 828: So, we support only URLs for plugins, right? Not local path and not PyPI packages. If that's the intention, then I'd add some comment about it. Maybe even in UI, it should be explicit that if you're in the cloud mode, you can only install from URLs.

src/scope/cloud/fal_app.py
```diff
@@ -198,6 +212,44 @@ def cleanup_session_data():
         print(f"Warning: Session cleanup failed: {e}")
 
 
+async def cleanup_installed_plugins():
+    """Uninstall all plugins installed during the session via the Scope API."""
+    import httpx
+
+    scope_url = "http://localhost:8000"
```
leszko: Do we always hardcode scope port to `8000` in FAL? I think it's ok, just want to make sure. And if that's the case, then consider extracting as a constant to be explicit.

src/scope/cloud/fal_app.py
```diff
@@ -198,6 +212,44 @@ def cleanup_session_data():
...
+                        f"{scope_url}/api/v1/plugins/{name}", timeout=60.0
+                    )
+                    if resp.status_code == 200:
+                        print(f"Cleanup: uninstalled plugin '{name}'")
+                    else:
+                        print(
+                            f"Warning: Failed to uninstall plugin '{name}': {resp.status_code}"
+                        )
+                except Exception as e:
+                    print(f"Warning: Failed to uninstall plugin '{name}': {e}")
```
leszko, line 336: I wonder if it would make sense to delete plugins in parallel?

src/scope/cloud/fal_app.py
```diff
@@ -296,37 +348,52 @@ def setup(self):
...
+            "NVIDIA_VISIBLE_DEVICES",
+            "NVIDIA_DRIVER_CAPABILITIES",
+            "LD_LIBRARY_PATH",
+            # Daydream API
+            "DAYDREAM_API_BASE",
+            # Kafka
+            "KAFKA_BOOTSTRAP_SERVERS",
+            "KAFKA_TOPIC",
+            "KAFKA_SASL_USERNAME",
+            "KAFKA_SASL_PASSWORD",
```
leszko, line 457: Ok, since we pass credentials into Scope, do we already have some safeguard for the plugins that are added into Daydream Plugin registry?

src/scope/cloud/fal_app.py
```diff
@@ -604,14 +674,52 @@ async def handle_api_request(payload: dict):
...
+                    normalized = re.sub(r"^https?://", "", normalized)
+                    # Remove .git suffix
+                    if normalized.endswith(".git"):
+                        normalized = normalized[:-4]
+                    # Remove trailing slashes
+                    normalized = normalized.rstrip("/")
+                    return normalized
+
+                def is_plugin_whitelisted(package: str) -> bool:
+                    raw = os.getenv("CLOUD_PLUGINS_WHITELIST", "")
```
leszko, line 837: How are you going to update this plugin list env variable?

.github/workflows/fal-deploy.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.10'
+          python-version: '3.12'
```
leszko, line 39: Are we sure that using different Python version won't have any effect?

src/scope/cloud/fal_app.py
```diff
@@ -604,14 +674,52 @@ async def handle_api_request(payload: dict):
             body = payload.get("body")
             request_id = payload.get("request_id")
 
-            # Block plugin installation in cloud mode (security: prevent arbitrary code execution)
             if method == "POST" and path == "/api/v1/plugins":
```
leszko: I wonder if this check is good enough. The path comes from the WebSocket message payload, so someone could hack like `/api/v1/plugins?hack=true` and then this check will not apply. Isn't that an issue?

src/scope/cloud/fal_app.py
mjh1: Are you sure that will be ok? I can test it out anyway

src/scope/cloud/fal_app.py
mjh1: This is a variable that lives in fal, it can be updated through their CLI or web dashboard.

.github/workflows/fal-deploy.yml
mjh1: It looks good in testing, I can do some more exhaustive testing though. In our pyproject file we specify python 3.12 so I think this was just an accidental case where the docker images were on 3.10 because that was the default installed with the older ubuntu version. I updated that which means it also needs to be updated here for the fal deploy.

src/scope/cloud/fal_app.py
mjh1: hmm maybe i'll avoid it for now:
```
There are a few concurrency hazards:
- pipeline_manager.unload_pipeline_by_id() mutates shared state (_pipelines, _pipeline_statuses, _pipeline_load_params dicts, and main pipeline state). The internal helper is even named _unload_pipeline_by_id_unsafe -- it must be called with the lock held. Running multiple unloads concurrently risks corrupting that shared state.
- CUDA cleanup -- each uninstall calls torch.cuda.empty_cache() and torch.cuda.synchronize(). Concurrent calls to these while pipelines are being torn down could interfere with each other.
- pip uninstall -- the sync implementation ultimately shells out to pip. Running multiple pip processes concurrently is not safe; pip uses a single shared site-packages directory and doesn't handle concurrent modifications.
- The plugin manager lock (self._lock) only guards the registry unregister section, not the entire uninstall operation, so it doesn't provide end-to-end protection.
```
I was thinking it sounded risky to have concurrent pip commands.

src/scope/cloud/fal_app.py
mjh1: oh yep good spot!

src/scope/cloud/fal_app.py
mjh1: I'll add a note to the UI 👍 in cloud mode we're only allowing installation from the discover tab for now, url entry is hidden.

src/scope/cloud/fal_app.py
mjh1: My plan is to create a write only user for this so that if they ever did leak it would be much less of a concern, they wouldn't be able to read our data. Then we can also have a scan over the current plugins to check nothing looks dangerous.

src/scope/cloud/fal_app.py
leszko: ok 👌

src/scope/cloud/fal_app.py
leszko: ok, but who will update it? Will we have some automation which will take our plugins and update this variable?

PR #544, opened 2026-02-26
Make Syphon input discovery more reliable

thomshutt:
Syphon server discovery relies on NSDistributedNotificationCenter, which delivers server announcement notifications exclusively to the main thread's NSRunLoop. In the FastAPI/uvicorn server, the main thread runs asyncio's event loop, so the NSRunLoop was never pumped. When ensure_directory_initialized() was called from a FastAPI worker thread, it pumped that worker thread's run loop instead — which receives zero notifications. This caused _get_raw_servers() to always return an empty list.

### DIFF
--- a/src/scope/server/app.py
+++ b/src/scope/server/app.py
@@ -318,7 +318,10 @@ async def lifespan(app: FastAPI):
     # Subsequent refreshes pump on demand in the list_input_sources endpoint.
     if sys.platform == "darwin":
         try:
-            from .syphon.receiver import ensure_directory_initialized, drain_notifications
+            from .syphon.receiver import (
+                drain_notifications,
+                ensure_directory_initialized,
+            )
 
             ensure_directory_initialized()
             drain_notifications(0.1)


src/scope/server/app.py
```diff
@@ -313,8 +313,43 @@ async def lifespan(app: FastAPI):
...
+            pump_run_loop(1.0)
+            logger.info(
+                "Syphon directory initialized with %d server(s)",
+                len(
+                    __import__("objc")
+                    .lookUpClass("SyphonServerDirectory")
+                    .sharedDirectory()
+                    .servers()
+                    or []
+                ),
```
leszko: This is weird, if that's so difficult to get the n of servers, then maybe just log without the number of servers?

src/scope/server/app.py
```diff
@@ -313,8 +313,43 @@ async def lifespan(app: FastAPI):
             kafka_publisher = None
             logger.warning("Kafka publisher failed to start")
 
+    # Syphon server discovery (macOS only): initialise the ObjC singleton and
```
leszko: I understand this part is needed, but I wonder if we could make it a little less "invasive":
1. Run it ONLY when user really uses Syphon (when the click on Syphon Input or when they refresh the Syphon inputs)
2. Nit: Move this whole code somewhere inside "syphon" code, just keep `_syphon_pump_task` here.

src/scope/server/frame_processor.py
```diff
@@ -92,6 +92,7 @@ def __init__(
         self.input_source: InputSource | None = None
         self.input_source_enabled = False
         self.input_source_type = ""
+        self.input_source_name = ""
```
leszko: Why do we need this `input_source_name` here? Is this related to fixing the issue with Syphon Input?

thomshutt: Yeah this wasn't related, it was something Opus had done to fix a speculative "bug" (switching between two sources of the same type is silently ignored) which wasn't related or reachable. Reverted.

src/scope/server/app.py
thomshutt: Yep, both good points - I've simplified it a lot now to just pump the loop when the user clicks the refresh button and also put the init code inside a conditional for MacOS

PR #555, opened 2026-02-28
Add Workflow Export/Import

ryanontheinside:
## Summary

Adds workflow export and import, allowing users to share and apply complete pipeline configurations.

**Backend:**
- `/resolve` endpoint that checks environment requirements (plugins, LoRAs) and reports what needs to be installed before a workflow can be applied

**Frontend:**
- Workflow export dialog with JSON download
- Workflow import dialog with step-by-step dependency resolution UI — shows required plugins, missing LoRAs, and pipeline compatibility before applying
- Extracted plugins and server info into shared React contexts
- Frontend refactoring: simplified prompt timeline, deduplicated pipeline logic, inlined workflow settings state

**Not included (follow-up PRs):**
- Images (reference image, first image, last image)
- Plugin resolution with versions
- Cloud support

### DIFF
--- a/frontend/src/components/WorkflowImportDialog.tsx
+++ b/frontend/src/components/WorkflowImportDialog.tsx
@@ -350,12 +350,7 @@ export function WorkflowImportDialog({
   );
   const installablePlugins = missingPlugins?.filter(i => i.can_auto_resolve);
 
-  const hasUnresolvedDeps = plan?.items.some(i => {
-    if (i.status !== "missing") return false;
-    if (i.kind === "lora") return loras.downloads[i.name] !== "done";
-    if (i.kind === "plugin") return plugins.installs[i.name] !== "done";
-    return true; // pipeline or other kinds — always unresolved if missing
-  });
+  const hasUnresolvedDeps = plan?.items.some(i => i.status === "missing");
 
   // -----------------------------------------------------------------------
   // Render


frontend/src/hooks/useLoRAFiles.ts
```diff
@@ -16,17 +16,19 @@ export function useLoRAFiles(): UseLoRAFilesReturn {
...
       setLoraFiles(response.lora_files);
+      return response.lora_files;
     } catch (error) {
       console.error("Failed to load LoRA files:", error);
+      return loraFiles;
     } finally {
       setIsLoading(false);
     }
-  }, [listLoRAFiles]);
+  }, [listLoRAFiles, loraFiles]);
```
mjh1: I found this causes the UI to make the loras API call in an infinite loop. this was opus's suggestion: https://github.com/daydreamlive/scope/commit/aec93b90c09883728f1ec88341a3d2ee4f7950a2#diff-2e62185f3def09fcf763dc320d9fbe14139bd00cbb9d98f8cbbd22e72d672a4f

mjh1: @leszko

leszko: Thanks, updated

PR #561, opened 2026-03-02
fix: persist input mode settings when pipeline switched

vcashwin:
## Overview
This PR fixes a regression where switching the selected pipeline/model in the right-side settings panel reset inputMode back to the pipeline default (text). The change preserves the user’s current input mode (text/video) across pipeline switches when that mode is supported, and only falls back to the pipeline default when the current mode is invalid for the newly selected pipeline.
It also updates stream settings normalization so mode is no longer force-reset on every pipeline/schema change, and keeps pre/postprocessor settings consistent by clearing only processors that are incompatible with the resolved mode.

Manual behavior checks covered:
- video mode persists when switching to another pipeline that supports video
- mode safely falls back to text for pipelines that only support text

### DIFF
--- a/frontend/src/hooks/useStreamState.ts
+++ b/frontend/src/hooks/useStreamState.ts
@@ -1,4 +1,4 @@
-import { useState, useCallback, useEffect, useRef } from "react";
+import { useState, useCallback, useEffect } from "react";
 import type {
   SystemMetrics,
   StreamStatus,
@@ -212,11 +212,16 @@ export function useStreamState() {
         ) {
           const firstPipelineId = availablePipelines[0] as PipelineId;
           const firstPipelineSchema = schemas.pipelines[firstPipelineId];
+          const nextMode =
+            prev.inputMode &&
+            firstPipelineSchema.supported_modes.includes(prev.inputMode)
+              ? prev.inputMode
+              : firstPipelineSchema.default_mode;
 
           return {
             ...prev,
             pipelineId: firstPipelineId,
-            inputMode: firstPipelineSchema.default_mode,
+            inputMode: nextMode,
           };
         }
         return prev;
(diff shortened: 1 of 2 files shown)


frontend/src/pages/StreamPage.tsx
```diff
@@ -438,6 +440,14 @@ export function StreamPage() {
         ? capResolution(customVideoResolution, pipelineId, modeToUse)
         : { height: defaults.height, width: defaults.width };
 
+    // Clear pre/postprocessors that don't support the selected mode
+    const preprocessorStillValid = settings.preprocessorIds?.every(id =>
+      pipelines?.[id]?.supportedModes?.includes(modeToUse)
+    );
+    const postprocessorStillValid = settings.postprocessorIds?.every(id =>
+      pipelines?.[id]?.supportedModes?.includes(modeToUse)
+    );
```
ryanontheinside, line 449: Nit: if `preprocessorIds` is `undefined`, `undefined?.every()` returns `undefined` (falsy), so this will spread the clearing fields even when there are no processors to clear. Doesn't cause bugs since the values are already empty, but the semantics are slightly off (this is according to Claude - wdyt?).

```suggestion
    const preprocessorStillValid = !settings.preprocessorIds?.length ||
      settings.preprocessorIds.every(id =>
        pipelines?.[id]?.supportedModes?.includes(modeToUse)
      );
    const postprocessorStillValid = !settings.postprocessorIds?.length ||
      settings.postprocessorIds.every(id =>
        pipelines?.[id]?.supportedModes?.includes(modeToUse)
      );
```

frontend/src/hooks/useStreamState.ts
```diff
@@ -212,11 +212,16 @@ export function useStreamState() {
         ) {
           const firstPipelineId = availablePipelines[0] as PipelineId;
           const firstPipelineSchema = schemas.pipelines[firstPipelineId];
+          const nextMode =
+            prev.inputMode &&
+            firstPipelineSchema.supported_modes.includes(prev.inputMode)
+              ? prev.inputMode
+              : firstPipelineSchema.default_mode;
```
ryanontheinside: Non-blocking: this pattern appears in three places now (here, the initial fetch below, and `handlePipelineIdChange` in StreamPage). Worth extracting a small helper like `resolveInputMode(currentMode, schema)` at some point to keep it DRY.

PR #565, opened 2026-03-02
Outputs UI position

thomshutt:
Move the Outputs section into a dedicated panel above Parameters to ensure it's always visible.

The previous layout often hid the Outputs section below the fold, making it inaccessible without scrolling. This change creates a separate, compact `OutputsPanel` and positions it prominently at the top of the right-side column.

image

### DIFF
--- a/frontend/src/components/OutputsPanel.tsx
+++ b/frontend/src/components/OutputsPanel.tsx
@@ -1,7 +1,5 @@
 import { Card, CardContent, CardHeader, CardTitle } from "./ui/card";
-import { Input } from "./ui/input";
-import { Toggle } from "./ui/toggle";
-import { LabelWithTooltip } from "./ui/label-with-tooltip";
+import { OutputSinkToggle } from "./OutputSinkToggle";
 import { PARAMETER_METADATA } from "../data/parameterMetadata";
 import type { SettingsState } from "../types";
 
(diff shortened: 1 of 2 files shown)


frontend/src/pages/StreamPage.tsx
```diff
@@ -123,6 +124,8 @@ export function StreamPage() {
   const syphonAvailable = availableInputSources.some(
     s => s.source_id === "syphon" && s.available
   );
+  const hasAvailableOutputs =
+    spoutAvailable || ndiOutputAvailable || syphonOutputAvailable;
```
leszko, line 128: Do we need this guard twice, here and in `OutputPanel.tsx`?

frontend/src/components/OutputsPanel.tsx
```diff
@@ -0,0 +1,191 @@
...
+                    });
+                  }}
+                  disabled={isStreaming}
+                  className="h-8 text-sm flex-1"
+                  placeholder="ScopeOut"
+                />
+              </div>
+            )}
+          </div>
+        )}
```
leszko: Super nit (and optional): These blocks look the same for each output, consider extracing to something like `OutputSinkToggle`.

frontend/src/components/SettingsPanel.tsx
```diff
@@ -502,7 +487,7 @@ export function SettingsPanel({
   return (
     <Card className={`h-full flex flex-col ${className}`}>
       <CardHeader className="flex-shrink-0">
-        <CardTitle className="text-base font-medium">Settings</CardTitle>
+        <CardTitle className="text-base font-medium">Parameters</CardTitle>
```
leszko: I see you renamed the Settings to Parameters. I'm not sure it's intentional. In general, I don't have any preferences, it's just that we usually call Parameters as Pipeline Parameters, this one is about the whole settings.

leszko (approved):
Added some comments. Other than that LGTM.

After resolving comments, feel free to click "Squash and Merge" (or rebase manually if you prefer).

frontend/src/components/SettingsPanel.tsx
thomshutt:
```suggestion
        <CardTitle className="text-base font-medium">Settings</CardTitle>
```

frontend/src/pages/StreamPage.tsx
thomshutt: Good call, removed the inner one

frontend/src/components/SettingsPanel.tsx
thomshutt: Yeah that was accidental sorry, reverted

frontend/src/components/OutputsPanel.tsx
thomshutt: Good call, DRYed it up

PR #569, opened 2026-03-03
fix: broken record on/off button, add support for output recording

vcashwin:
The recording button was previously disabled during streaming and relied on server-side recording via the backend API. This PR moves recording to the client, allowing users to record the output video directly from the browser during an active stream.

**Changes:**
- Added `useRecording` hook and `recordingUtils` that use an offscreen canvas + MediaRecorder to capture the output video element and auto-download the file on stop
- Flipped the record button to be enabled during streaming (instead of only before), since client-side recording requires an active stream
- Exposed the `<video>` ref from VideoOutput and wired it into StreamPage, removing the server-side recording parameter from the WebRTC initial parameters

https://github.com/user-attachments/assets/3d1db142-9546-4064-be16-eb5304536d35

### DIFF
--- a/frontend/src/components/PromptTimeline.tsx
+++ b/frontend/src/components/PromptTimeline.tsx
@@ -694,10 +694,16 @@ export function PromptTimeline({
             </Button>
             <Button
               onClick={onRecordingToggle}
-              disabled={disabled || isLoading || isDownloading || isStreaming}
+              disabled={disabled || isLoading || isDownloading || !isStreaming}
               size="sm"
               variant="outline"
-              title={isRecording ? "Stop recording" : "Start recording"}
+              title={
+                !isStreaming
+                  ? "Start streaming to enable recording"
+                  : isRecording
+                    ? "Stop recording"
+                    : "Start recording"
+              }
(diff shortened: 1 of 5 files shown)


leszko (changes requested):
AFAIU This PR does not only fix the UI behavior, but also moves the whole recording from backend to frontend. I'm not saying it's wrong, but it definitely requires some discussions and thinking.

**A few questions:**
1. @yondonfu do you have any thoughts? Do you remember why we chose to implement it in the backend?
2. @vcashwin What is the max recording length frontend can handle? Is it ok with the storage?
3. @ryanontheinside After this change, the recording won't work with the Scope API-only mode. Isn't it a problem? I think nobody is using it Scope API-only. But I'd like to double-check.

As I mentioned somewhere, in general I like the idea of recording being frontend-only, but I want to open this discussion to make sure we don't miss anything. Additionally, if we go that path, we should also clean up the code from the backend-side recording.

Tagging also @BuffMcBigHuge , I remember we had these discussions before, so maybe you remember some rationale behind the server-side recording.

PR #578, opened 2026-03-04
Enable workflows for remote inference

mjh1:
Proxy new API calls to cloud to enable new workflow import/export

### DIFF
--- a/src/scope/server/app.py
+++ b/src/scope/server/app.py
@@ -48,6 +48,7 @@ from scope.core.workflows.resolve import (
 from .cloud_proxy import (
     cloud_proxy,
     get_hardware_info_from_cloud,
+    proxy_with_body,
     recording_download_cloud_path,
     upload_asset_to_cloud,
 )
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-build.yml
```diff
@@ -171,13 +171,12 @@ jobs:
...
           PR_NUMBER="${{ needs.build-and-push.outputs.pr_number }}"
-          SHORT_SHA="${{ needs.build-and-push.outputs.short_sha }}"
 
           APP_NAME="scope-pr-${PR_NUMBER}"
 
-          echo "Deploying PR #${PR_NUMBER} to fal as ${APP_NAME} with tag ${SHORT_SHA}"
+          echo "Deploying PR #${PR_NUMBER} to fal as ${APP_NAME}"
 
-          SCOPE_DEPLOY_TAG=$SHORT_SHA fal deploy \
+          fal deploy \
```
mjh1, line 179: tidied this up, it can just get the tag based on the checked out version itself

src/scope/core/workflows/resolve.py
```diff
@@ -249,7 +249,7 @@ def _check_min_scope_version(
 def resolve_workflow(
     workflow: WorkflowRequest,
     plugin_manager: PluginManager,
-    models_dir: Path,
+    lora_dir: Path,
```
mjh1, line 252: in cloud we have loras stored separate to models so that they're not shared between users, so we need to use `get_lora_dir()` instead of `models_dir / "lora"`

src/scope/server/app.py
```diff
@@ -1210,13 +1220,51 @@ async def delete_lora_file(
...
+            raise HTTPException(
+                status_code=status,
+                detail=response.get("error", "Cloud request failed"),
+            )
+        return response.get("data", {})
 
     lora_dir = get_lora_dir()
     try:
-        return await download_lora(request, lora_dir)
+        return await download_lora(request, lora_dir, civitai_token=civitai_token)
```
mjh1, line 1227: @ryanontheinside @leszko Not sure how you feel about doing it this way but in cloud mode I needed some way of passing the token from the user to the remote server, so it gets added to the request that gets proxied above in the `if cloud_manager.is_connected:` block, then it gets passed into `download_lora` here so that it can override `get_civitai_token()` on the remote side which will return nothing (we only store the api keys locally and send them along with the requests when they're proxied to remote).

leszko: I think this is good.

PR #631, opened 2026-03-09
Add ability to adjust LoRA strength via OSC

thomshutt:
Screenshot 2026-03-09 at 13 27 34

https://github.com/user-attachments/assets/4be822cb-d79c-42ba-802c-863767c2830f

## Summary by CodeRabbit

* **New Features**
  * Real-time LoRA scale control via OSC and a new API endpoint to list loaded LoRA adapters.

* **Improvements**
  * Consolidated pipeline refresh that also updates pipeline status and keeps cloud/direct mode in sync.
  * OSC advertising now exposes LoRA adapter control paths.
  * Positional adapter matching and adapter-name support for more accurate scale updates.
  * Concurrent WebRTC session cleanup for faster shutdowns.

* **Bug Fixes**
  * More reliable discovery and reporting of loaded LoRA adapters across local/cloud flows.

### DIFF
--- a/frontend/src/pages/StreamPage.tsx
+++ b/frontend/src/pages/StreamPage.tsx
@@ -13,6 +13,7 @@ import type { TimelinePrompt } from "../components/PromptTimeline";
 import { StatusBar } from "../components/StatusBar";
 import { LogPanel } from "../components/LogPanel";
 import { useUnifiedWebRTC } from "../hooks/useUnifiedWebRTC";
+import { MIDIProvider } from "../contexts/MIDIContext";
 import { useVideoSource } from "../hooks/useVideoSource";
 import { useWebRTCStats } from "../hooks/useWebRTCStats";
 import { useControllerInput } from "../hooks/useControllerInput";
@@ -36,6 +37,7 @@ import type {
   LoRAConfig,
   LoraMergeStrategy,
   DownloadProgress,
+  SettingsState,
 } from "../types";
 import type { PromptItem, PromptTransition } from "../lib/api";
 import { getInputSourceResolution, fetchDaydreamWorkflow } from "../lib/api";
@@ -94,7 +96,6 @@ export function StreamPage() {
   const {
     isConnected: isBackendCloudConnected,
     isConnecting: isBackendCloudConnecting,
-    connectStage,
   } = useCloudStatus();
 
   // Combined cloud mode: either frontend direct-to-cloud or backend relay to cloud
@@ -294,7 +295,6 @@ export function StreamPage() {
     error: pipelineError,
     loadPipeline,
     pipelineInfo,
-    loadingStage,
     checkStatus: checkPipelineStatus,
   } = usePipeline();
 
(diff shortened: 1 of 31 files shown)


src/scope/server/webrtc.py
```diff
@@ -652,8 +652,13 @@ def broadcast_parameter_update(self, parameters: dict) -> None:
...
                 session.video_track.pause(parameters["paused"])
-            if session.video_track and hasattr(session.video_track, "frame_processor"):
-                session.video_track.frame_processor.update_parameters(parameters)
+            fp = (
+                getattr(session.video_track, "frame_processor", None)
+                if session.video_track
+                else None
+            )
+            if fp is not None:
+                fp.update_parameters(parameters)
```
leszko, line 661: What is this change about? Isn't the logic equal to what was before?

src/scope/server/pipeline_manager.py
```diff
@@ -100,6 +100,61 @@ def get_loaded_pipeline_ids(self) -> list[str]:
...
+        """Store pipeline IDs and load params locally without loading pipelines.
+
+        Used in cloud inference mode so the local OSC server can discover LoRA
+        adapters (via the load_params fallback) even though the actual pipeline
+        runs on a remote host.
+        """
+        with self._lock:
+            self._load_params = load_params
+            for pid in pipeline_ids:
+                self._pipeline_load_params[pid] = load_params or {}
```
leszko, line 122: If we use cloud, do we ever clear these values? I wonder if you load some loras once, and then some other loras, won't you have both here?

src/scope/server/osc_docs.py
```diff
@@ -155,6 +152,35 @@ def _collect_pipeline_paths() -> dict[str, list[dict[str, Any]]]:
...
+    if not pipeline_manager:
+        return []
+    loaded = pipeline_manager.get_loaded_lora_adapters()
+    paths: list[dict[str, Any]] = []
+    for i, adapter in enumerate(loaded):
+        adapter_path = adapter.get("path", "")
+        filename = (
+            adapter_path.rsplit("/", 1)[-1]
+            if "/" in adapter_path
+            else (adapter_path or f"slot {i}")
```
leszko, line 177: Will the hardcoding of `/` work in Windows? Maybe use some `pathlib.PurePosixPath` or `os.path.basename` to handle both OS?

frontend/src/pages/StreamPage.tsx
```diff
@@ -1152,6 +1152,20 @@ export function StreamPage() {
           break;
         }
         default: {
+          // LoRA adapter scale updates (e.g. "lora/0/scale")
+          const loraMatch = key.match(/^lora\/(\d+)\/scale$/);
```
leszko, line 1300: Nit: Not sure how often this is called, but if often, then you may extract `const loraMatch = key.match(/^lora\/(\d+)\/scale$/);` outside the function, otherwise the regex is complied on every SSE message.

src/scope/server/app.py
```diff
@@ -586,6 +587,20 @@ async def load_pipeline(
         # load_params is already a dict (or None)
         load_params_dict = request.load_params
 
+        # Always store load context locally so the OSC server can discover
+        # LoRA adapters even when the pipeline runs on a remote host.
+        pipeline_manager.store_load_context(pipeline_ids, load_params_dict)
```
leszko, line 624: I think this is a little fishy, won't it be better to expose `get_loaded_lora_adapters` as an endpoint and then proxy this call to the cloud, then we would always have the data directly from the cloud and not sometimes from cloud, sometimes partly from local? 

I think the simple flow is that:
- connected to cloud => Request everything from cloud
- not connected to cloud => Request everything from local

To have a different flow is an exception and, unless there is a good reason, I'd try to avoid it.

thomshutt: I think this is probably the right way to do it, I was just nervous about blowing up the scope of this ticket but I'll try now and see how it goes

PR #632, opened 2026-03-09
longlive: performance optimization

varshith15:
This PR refactors the LongLive causal model to be fully torch.compile-compatible, enabling significant inference speedups. The entire transformer block loop (including VACE blocks and LoRA) is captured in a single compiled region with zero graph breaks, allowing CUDA graph capture for maximum GPU utilization.

**Performance gains**:
- ~40% speedup on RTX 5090
- ~55-60% speedup on H100
- Compile warmup is 5x-8x faster than krea_realtime_video: only 1 iteration needs compilation (vs 4-5 iterations for krea's cache-filling warmup). Subsequent runs reuse the compiled cache so startup is fast.

FP8 + VACE + Compile all work together -- the previous FP8 quantization path broke when VACE was enabled or when compiling. This is no longer an issue.

All changes currently target the LongLive pipeline, but since most pipelines share the same base Wan2.1 model, the same approach can be extended to the others.

#### 1. Compile-friendly model rewrite (`causal_model.py`)

The core model was rewritten to eliminate every `torch.compile` graph break:

- **RoPE uses real-valued math.** `torch.view_as_complex` causes graph breaks. Replaced with `precompute_freqs_i()` (cos/sin outside compiled region) and `causal_rope_apply_precomputed()` (real-valued rotation inside).
- **Cross-attention inlined.** Q/K/V projections and attention are done inline in the block to avoid graph breaks from submodule dispatch.
- **Removed dead code.** Training paths (`_forward_train`), `model_type` param, `init_weights()`, and all FlexAttention mask generation methods.

#### 2. Two-tier forward for compilation

- `_inner_forward()` -- uncompiled, used while the KV cache is filling (dynamic shapes).
- `_compiled_inner_forward` -- `torch.compile(mode="max-autotune-no-cudagraphs")` version, used once the cache is full (static shapes). Avoids recompilation from shape changes.

#### 3. VACE integrated natively

VACE is no longer a separate wrapper (`VACEEnabledPipeline` removed). `CausalWanModel.init_vace()` creates VACE modules directly on the model. A new `VaceBlock` class handles hint generation. VACE embedding runs outside the compiled region, but the block loop and hint injection run inside it.

#### 4. Centralized KV cache management

The old per-layer cache tracking (`local_end_index`, `global_end_index`, `cache_update_info` dicts, `clone()`-based rolling) is replaced by a single `_roll_update_cache()` with a `fill_level` counter. During filling, new KVs append at `fill_level`. Once full, the non-sink portion is rolled via `torch.roll` and new KVs overwrite the tail. Sink tokens are always preserved.

#### 5. LongLive-specific `SetupCachesBlock`

New `longlive/blocks/setup_caches.py` replaces the shared wan2_1 version. Simplified cache init (just `k`/`v` tensors, no index tracking) and helper functions to propagate runtime config (frame_seq_length, attention size, sink tokens) to all submodules.

#### 6. Chunk-based recache

Recache now processes frames in `num_frame_per_block`-sized chunks instead of all at once. Each chunk sees sink frames + previously recached chunks as context for smoother prompt transitions. The old recache logic is preserved in `memflow/blocks/recache_frames.py` for backward compatibility.

#### 7. FP8 + VACE compatibility

New `vace_fp8_compatible` flag on `BasePipelineConfig`. Frontend updated to allow FP8 when VACE is enabled for compatible pipelines (previously always force-disabled).

#### `crossattn_cache` is removed

`crossattn_cache` only optimizes by 0.07% but it breaks the inference graph, so removed it. 

#### 8. Frontend & server

- Compile toggle in settings panel.
- Load timeout extended to 30min server-side (3x on frontend) when compile is enabled.
- `crossattn_cache` made optional across shared blocks (`DenoiseBlock`, `CleanKVCacheBlock`, `RecacheFramesBlock`).

## Summary by CodeRabbit

* **New Features**
  * UI "Compile" toggle to enable model compilation with an informational note about first-time compile vs. cache.
  * Native VACE support plus enhanced cache setup and recache capabilities for improved streaming/video workflows.

* **Improvements**
  * VACE / FP8 compatibility exposed in settings; FP8 controls and warnings now reflect compatibility.
  * Longer load timeouts when compilation is requested.
  * Streamlined attention/cache handling for more reliable performance.


src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
```diff
@@ -179,7 +185,8 @@ def __call__(self, components, state: PipelineState) -> tuple[Any, PipelineState
         has_extension = has_first_frame or has_last_frame
 
         if not has_ref_images and not has_input_frames and not has_extension:
-            block_state.vace_context = None
+            if not hasattr(block_state, 'vace_context'):
+                block_state.vace_context = None
```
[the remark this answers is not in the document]
varshith15: its needed for warmup

src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
```diff
@@ -179,7 +185,8 @@ def __call__(self, components, state: PipelineState) -> tuple[Any, PipelineState
         has_extension = has_first_frame or has_last_frame
 
         if not has_ref_images and not has_input_frames and not has_extension:
-            block_state.vace_context = None
+            if not hasattr(block_state, "vace_context"):
```
ryanontheinside: **Bug:** This guard means once `vace_context` is set on chunk 1 (e.g., from first-frame conditioning), it never gets cleared on subsequent chunks. The old code unconditionally set `block_state.vace_context = None` when no ref images/input frames/extensions were present, which correctly reset it each chunk. With the `hasattr` guard, stale VACE context bleeds into all subsequent chunks.

Reverting to the unconditional assignment seems like the fix, or clearing `vace_context` in `LongLivePipeline._generate()` when it's not in kwargs (matching the existing pattern for `first_frame_image`/`last_frame_image`). wdyt?

src/scope/core/pipelines/longlive/modules/causal_model.py
```diff
@@ -725,608 +413,167 @@ def __init__(
...
+                kv_cache[i]["k"][:, self.sink_tokens:] = torch.roll(
+                    kv_cache[i]["k"][:, self.sink_tokens:], -take, dims=1
+                )
+                kv_cache[i]["v"][:, self.sink_tokens:] = torch.roll(
+                    kv_cache[i]["v"][:, self.sink_tokens:], -take, dims=1
+                )
+                kv_cache[i]["k"][:, -take:] = new_k[:, -take:]
+                kv_cache[i]["v"][:, -take:] = new_v[:, -take:]
+            else:
+                kv_cache[i]["k"][:, self.fill_level:self.fill_level + take] = new_k[:, -take:]
```
ryanontheinside: **Bug:** When using VACE input video, the first chunk produces output successfully but chunk 2+ crashes with:
```
RuntimeError: The expanded size of the tensor (2048) must match the existing size (3072) at non-singleton dimension 1. Target sizes: [1, 2048, 12, 128]. Tensor sizes: [3072, 12, 128]
```
at this line. The KV cache is allocated for 2048 tokens but the new KVs have 3072. Looks like the VACE path is producing KVs with an inflated sequence dimension that doesn't match the cache allocation.

frontend/src/components/ComplexFields.tsx
```diff
@@ -514,5 +516,43 @@ export function SchemaComplexField({
...
+            disabled={(ctx.isStreaming ?? false) || (ctx.isLoading ?? false)}
+          >
+            {compileValue ? "ON" : "OFF"}
+          </Toggle>
+        </div>
+        {compileValue && (
+          <div className="flex items-start gap-1.5 p-2 rounded-md bg-blue-500/10 border border-blue-500/20">
+            <Info className="h-3.5 w-3.5 mt-0.5 shrink-0 text-blue-600 dark:text-blue-400" />
+            <p className="text-xs text-blue-600 dark:text-blue-400">
+              Note: First-time compilation can take several minutes. Subsequent
```
ryanontheinside: The banner says "first-time compilation can take several minutes, subsequent runs will be significantly faster," but some `is_load_param=True` changes (resolution, LoRA add/remove/swap) likely trigger torch.compile recompilation since they change tensor shapes or model graph structure, not just a fast reload with cached compilation. Worth enumerating which settings changes actually trigger recompilation so users know what's expensive?

src/scope/core/pipelines/longlive/pipeline.py
```diff
@@ -198,6 +205,15 @@ def __init__(
         self.first_call = True
         self.last_mode = None  # Track mode for transition detection
 
+        if compile:
+            inner_model = self.components.generator.model
+            if hasattr(inner_model, "get_base_model"):
+                inner_model = inner_model.get_base_model()
+            inner_model._compiled_inner_forward = torch.compile(
```
ryanontheinside, line 207: Any plans to coordinate with remote infra on `torch.compiler.save_cache_artifacts()` or setting `TORCHINDUCTOR_CACHE_DIR` to a stable path (e.g., `~/.daydream-scope/compile_cache/`)? Without this, users lose compilation cache across environments or when the default cache dir gets cleaned up.

src/scope/core/pipelines/longlive/schema.py
```diff
@@ -125,6 +125,15 @@ class LongLiveConfig(BasePipelineConfig):
...
     )
+    compile: bool = Field(
+        default=False,
+        description="Use torch.compile for faster inference. First-time compilation takes several minutes.",
+        json_schema_extra=ui_field_config(
+            order=9, component="compile", is_load_param=True, label="Compile"
+        ),
+    )
+
+    vace_fp8_compatible = True
```
ryanontheinside: Should be `vace_fp8_compatible: ClassVar[bool] = True` to match the parent class annotation in `base_schema.py`?

src/scope/core/pipelines/longlive/pipeline.py
```diff
@@ -249,3 +265,31 @@ def _generate(self, **kwargs) -> dict:
 
         _, self.state = self.blocks(self.components, self.state)
         return {"video": postprocess_chunk(self.state.values["output_video"])}
+
+    def _warmup(self, inner_model):
+        print("Running warmup iteration...")
```
ryanontheinside: These `print()` calls should use `logger.info()` to match the rest of the codebase.

src/scope/core/pipelines/longlive/pipeline.py
```diff
@@ -86,11 +82,22 @@ def __init__(
             **base_model_kwargs,
         )
 
-        # Apply VACE wrapper if vace_path is configured (upfront loading)
-        # This must happen before LoRA to get correct ordering: LoRA -> VACE -> Base
-        generator.model = self._init_vace(
-            config, generator.model, device=device, dtype=dtype
-        )
+        # Initialize VACE natively on the model (no wrapper needed)
```
ryanontheinside, line 85: I saw your visual comparisons for the raw model, but could you share similar for (same seed/prompt/input) confirming equivalent output quality between the old wrapper path and this native integration with VACE? I saw no real degradation with the raw model, so excited to see this with conditioning frames, inpainting, and first frame conditioning

src/scope/core/pipelines/memflow/blocks/recache_frames.py
```diff
@@ -0,0 +1,222 @@
...
+                required=True,
+                type_hint=list[dict],
+                description="Initialized KV cache",
+            ),
+            InputParam(
+                "crossattn_cache",
+                required=True,
+                type_hint=list[dict],
+                description="Initialized cross-attention cache",
+            ),
```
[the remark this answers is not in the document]
varshith15: its only linked to longlive, this change is not needed for now

src/scope/core/pipelines/longlive/modules/causal_model.py
```diff
@@ -725,608 +413,167 @@ def __init__(
...
-            KV_LEN=total_length + padded_length,
-            _compile=False,
-            device=device,
-        )
+        c = [self.vace_patch_embedding(u.unsqueeze(0).to(dtype=target_dtype)) for u in vace_context]
+        c = [u.flatten(2).transpose(1, 2) for u in c]
+        return torch.cat([
+            torch.cat([u, u.new_zeros(1, max(0, seq_len - u.size(1)), u.size(2))], dim=1)
+            for u in c
+        ])
```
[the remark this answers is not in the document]
varshith15: wont be the case with vace_blocks if it is, its handled in the vace_encoding.py block

src/scope/core/pipelines/longlive/pipeline.py
varshith15: yup, definitely
the best way is to create a .bin of the compiled artifacts and use it as a model to be downloaded and loaded
because this PR is already huge, i thought i'll add it in a separate one.

varshith15: for 5090 and h100 atleast, we can add for others as and when there is a requirement
because its a uv project, i think is should be fine

src/scope/core/pipelines/wan2_1/vace/blocks/vace_encoding.py
varshith15: removed it, passing dummy vace_input_frames now instead for warmup

frontend/src/components/ComplexFields.tsx
```diff
@@ -514,5 +515,43 @@ export function SchemaComplexField({
...
```
leszko: Nit: maybe keep the style of other banners (so make it yellow like the banner for the resoultion).

Image

Image

src/scope/server/pipeline_processor.py
```diff
@@ -19,14 +19,31 @@
 logger = logging.getLogger(__name__)
 
 # Multiply the # of output frames from pipeline by this to get the max size of the output queue
-OUTPUT_QUEUE_MAX_SIZE_FACTOR = 2
+OUTPUT_QUEUE_MAX_SIZE_FACTOR = 3
+
+
+def _has_native_vace(pipeline) -> bool:
```
leszko: I think it does not fit. Plus, heads-up, when we merge the Graph Mode backend PR, the concept of VACE will not even be in the Pipeline Processor. Pipeline Processor will not be aware of VACE at all.

src/scope/core/pipelines/memflow/blocks/recache_frames.py
leszko, line 1: Why does this PR modify memflow code if it does not enable compilation for memflow (but only for longlive)? If this code is shared between memflow and longlive, then we should move it outside memflow package.

PR #534, opened 2026-02-24
Audio Support for Scope Rework

BuffMcBigHuge:
# Audio Support for Scope (Reworked)

Overall, this approach is simplified and seems to solve audio quality issues. The complex gating mechanics and resampling have been removed, leveraging WebRTC aiortc-style timing.

## Summary

Adds end-to-end audio support to Scope's WebRTC streaming pipeline. Pipelines can return audio alongside video in their output dict; the server streams audio over WebRTC. This is a **simplified rewrite** of the audio path that fixes clipping and audio quality issues reported in [PR #480](https://github.com/daydreamlive/scope/pull/480).

### Backend

- **Pipeline interface**: Pipelines may return `{"video": ..., "audio": ..., "audio_sample_rate": ...}`. Audio keys are optional; pipelines that don't produce audio are unchanged.
- **PipelineProcessor**: Uses an `audio_callback` instead of a queue. Only the last processor in a chain receives the callback.
- **FrameProcessor**: Simple `audio_queue` for raw `(audio_tensor, sample_rate)` tuples. No background drain thread, no video-gated release, no resampling.
- **AudioProcessingTrack**: Receives raw audio from FrameProcessor, resamples to 48 kHz when needed (per-channel, preserves stereo), buffers samples, and delivers 20ms stereo frames for WebRTC/Opus. Uses aiortc-style monotonic pacing.

### Frontend

- **VideoOutput**: Mute/unmute toggle (speaker icon). Starts muted to satisfy browser autoplay policy; user can unmute once the stream is playing.
- **useUnifiedWebRTC**: Merges video and audio tracks into a single `MediaStream`. Adds a recvonly audio transceiver so the SDP offer includes an audio m-line for the backend to attach its track.

### WebRTC Handshake

The browser adds `addTransceiver("audio", { direction: "recvonly" })` so the offer includes an audio m-line. After `setRemoteDescription`, the backend finds the audio transceiver, attaches its `AudioProcessingTrack`, and sets direction to `sendonly`. The answer then indicates that the server will send audio.

## Why This Version (audio-sync-2)

[PR #480](https://github.com/daydreamlive/scope/pull/480) received feedback about:
- **Audio quality / clipping** – video and audio not synchronized, poor experience
- **Code complexity** – video-gated release, drain thread, MediaClock coupling

This branch is a simplified rewrite that:

[...]

PR #639, opened 2026-03-09
feat: audio-only pipeline support, decouple FrameProcessor from VideoProcessingTrack

ryanontheinside:
## Summary

- **Decouple FrameProcessor from VideoProcessingTrack**: Session now owns the FrameProcessor lifecycle and injects it into both VideoProcessingTrack and AudioProcessingTrack. This removes the awkward coupling where audio had to route through the video track to reach the shared processor.
- **Conditional video track creation**: Pipelines can declare `produces_video = False` in their config. When the last pipeline in a chain is audio-only, the server skips VideoProcessingTrack, MediaRelay, and RecordingManager entirely.
- **Fix audio delivery for audio-only pipelines**: Audio callback now fires before the video early-return in PipelineProcessor, so pipelines that return only `{"audio": ..., "audio_sample_rate": ...}` actually deliver audio.
- **Add `requires_audio_input` ClassVar**: Forward-looking flag on BasePipelineConfig for pipelines that expect browser audio input (microphone). Not wired yet; placeholder for the audio input workstream.
- **Frontend audio-only indicator**: VideoOutput component detects absence of video tracks and shows an audio-only UI with a hidden media element for playback.

## Depends on

- #534 (Audio Support for Scope Rework)

## Test plan

- [ ] Video+audio pipeline (e.g. LTX-2) still works as before
- [ ] Audio-only pipeline (e.g. test_tone with `produces_video = False`) delivers audio without creating a video track
- [ ] Frontend shows "Audio Only" indicator when no video track is present
- [ ] Session cleanup properly stops FrameProcessor
- [ ] Parameter updates route through `session.frame_processor` correctly
- [ ] Pipeline schema metadata includes `produces_video` and `requires_audio_input`

### DIFF
--- a/frontend/src/components/VideoOutput.tsx
+++ b/frontend/src/components/VideoOutput.tsx
@@ -54,6 +54,7 @@ export function VideoOutput({
   // User can click the speaker icon to unmute once the stream is playing.
   const [isMuted, setIsMuted] = useState(true);
   const [hasAudioTrack, setHasAudioTrack] = useState(false);
+  const [hasVideoTrack, setHasVideoTrack] = useState(false);
 
   // Use external ref if provided, otherwise use internal
   const containerRef = videoContainerRef || internalContainerRef;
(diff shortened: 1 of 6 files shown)


j0sh (approved):
Seems like a sensible change directionally, although I am not familiar enough with the Scope architecture to review this very deeply.

PR #653, opened 2026-03-10
Restore workflow on cloud reconnection

mjh1:
## Summary by CodeRabbit

* **New Features**
  * Automatic cloud workflow backup now preserves your work while connected to the cloud
  * Restore prompt appears when reconnecting to the cloud after a disconnect, allowing you to recover your previous session

You can test by manually disconnecting from cloud and reconnecting
Screenshot 2026-03-12 at 12 10 51

### DIFF
--- a/frontend/src/components/WorkflowImportDialog.tsx
+++ b/frontend/src/components/WorkflowImportDialog.tsx
@@ -67,6 +67,8 @@ interface WorkflowImportDialogProps {
     promptState: WorkflowPromptState | null
   ) => void;
   initialWorkflow?: ScopeWorkflow | null;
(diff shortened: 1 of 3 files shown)


leszko (changes requested):
I tried but it hasn't worked for me. Here's what I do.
1. Connect to cloud
2. Import some workflow (I import this one attached)
3. Start the stream (all good)
4. Close the app, kill the server
5. Start again
6. Connect to cloud
7. The dialog with Restore appears (all good)
8. I click Restore, but nothing is restored, I see only the default longlive pipeline

Do I maybe need some special FAL App instance? (I used the default one)

[complex_cloud.scope-workflow.json](https://github.com/user-attachments/files/25939795/complex_cloud.scope-workflow.json)

PR #654, opened 2026-03-10
feat: improve cloud button discoverability and header icon visibility

emranemran:
Change disconnected state text from "Disconnected" to "Enable Remote Inference" so new users understand the feature is available. Bump opacity on all three header icons (cloud, plugins, settings) for better visibility.

## Summary by CodeRabbit

* **Style**
  * Updated cloud status indicator labels, tooltips, and styling for improved clarity and guidance
  * Adjusted opacity of cloud, plugins, and settings icons for better visibility

* **Tests**
  * Broadened end-to-end cloud activation checks to cover both cloud and remote inference entry points for more reliable test coverage

### DIFF
--- a/frontend/src/components/Header.tsx
+++ b/frontend/src/components/Header.tsx
@@ -138,17 +138,17 @@ export function Header({
             onClick={handleCloudIconClick}
             className={`hover:opacity-80 transition-opacity h-8 gap-1.5 px-2 ${
               isConnected
-                ? "text-green-500 opacity-80"
+                ? "text-green-500 opacity-100"
                 : isConnecting
-                  ? "text-amber-400 opacity-80"
-                  : "text-muted-foreground opacity-60"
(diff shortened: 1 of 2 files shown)


frontend/src/components/Header.tsx
```diff
@@ -163,14 +163,14 @@ export function Header({
...
-                  : "Disconnected"}
+                  : "Enable Remote Inference"}
             </span>
           </Button>
           <Button
             variant="ghost"
             size="icon"
             onClick={() => setPluginsOpen(true)}
-            className="hover:opacity-80 transition-opacity text-muted-foreground opacity-60 h-8 w-8"
+            className="hover:opacity-80 transition-opacity text-muted-foreground opacity-80 h-8 w-8"
```
thomshutt, line 173: Doesn't this remove the hover effect since it now has the same opacity as the regular one?

PR #660, opened 2026-03-11
Add MCP server for AI-assisted pipeline control

ryanontheinside:
## Summary

- Adds an MCP (Model Context Protocol) server that exposes Scope's API as tools, enabling AI assistants to programmatically manage pipelines, control parameters, capture frames, and interact with a running instance over stdio
- Adds REST endpoints for headless sessions (start/stop, parameter get/set, frame capture, metrics, recording) so the MCP server (and other clients) can drive pipelines without a browser
- Adds `onParametersUpdated` callback to the frontend WebRTC hook so external parameter changes (via MCP, REST, OSC) sync the UI
- Fixes a bug where the MCP subprocess created a competing log file at import time, causing `get_logs` to return empty results instead of the server's actual logs

## Details

**MCP server** (`src/scope/server/mcp_server.py`): Thin stdio client that proxies to the Scope HTTP API. Covers pipeline lifecycle, runtime parameters, frame capture, session metrics, recording, LoRA/plugin/asset management, input sources, OSC, workflows, and API keys. Run with `daydream-scope --mcp --port <port>`.

**New REST endpoints**: `POST/GET /api/v1/session/parameters`, `GET /api/v1/session/frame`, `GET /api/v1/session/metrics`, `POST /api/v1/session/start`, `POST /api/v1/session/{id}/stop`, `POST /api/v1/pipeline/unload`, `POST /api/v1/session/{id}/recording/start|stop`.

**Backend changes**: `HeadlessSession` class and headless session management on `WebRTCManager`, `broadcast_notification()` for pushing parameter changes to frontends, `VideoProcessingTrack.get_last_frame()` public accessor, deferred logging setup to avoid MCP subprocess interference.

**Frontend changes**: `useUnifiedWebRTC` gains `onParametersUpdated` callback; `StreamPage` extracts `applyBackendParamsToSettings` so both local sends and external pushes share the same settings-sync path.

## Test plan

- [ ] Run `daydream-scope` and verify server starts without logging errors
- [ ] Run `daydream-scope --mcp --port 8000` and verify MCP server connects
- [ ] Load a pipeline via MCP, start a headless stream, capture a frame
- [ ] Update parameters via MCP and verify the frontend UI reflects changes
- [ ] Call `get_logs` via MCP and verify it returns actual server logs (not empty)
- [ ] Verify WebRTC streaming still works normally through the browser UI

## Summary by CodeRabbit

* **New Features**
  * Added MCP (Model Context Protocol) server support for remote integration and automation
  * Enabled real-time parameter synchronization from backend to frontend UI
  * Added session management endpoints (start, stop sessions)
  * Added recording management (start, stop, download recordings)
  * Added frame capture and performance metrics endpoints
  * Enabled headless session execution without WebRTC connections

### DIFF
--- a/src/scope/server/mcp_server.py
+++ b/src/scope/server/mcp_server.py
@@ -251,7 +251,7 @@ def create_mcp_server(base_url: str = "http://localhost:8000") -> FastMCP:
             pipeline_id: Pipeline ID to run (must already be loaded)
             input_mode: "text" for prompt-only generation, "video" for input source processing
             prompts: Initial prompts, e.g. [{"text": "a forest", "weight": 100}]
-            input_source: Server-side input source config for video mode. Format: {"enabled": true, "source_type": "<type>", "source_name": "<name>"}
+            input_source: Server-side input source config for video mode. Format: {"enabled": true, "source_type": "<type>", "source_name": "<name>"}. For video_file, source_name can be a full file path or an asset name.
         """
         body: dict = {"pipeline_id": pipeline_id, "input_mode": input_mode}
         if prompts is not None:
--- a/src/scope/core/inputs/__init__.py
+++ b/src/scope/core/inputs/__init__.py
@@ -41,6 +41,13 @@ def get_input_source_classes() -> dict[str, type[InputSource]]:
     except Exception:
         pass
 
+    try:
+        from .video_file import VideoFileInputSource
+
+        sources[VideoFileInputSource.source_id] = VideoFileInputSource
+    except Exception:
+        pass
+
     return sources
 
 
--- a/src/scope/core/inputs/video_file.py
+++ b/src/scope/core/inputs/video_file.py
@@ -1,0 +1,200 @@
+"""Video file input source implementation using PyAV."""
+
+import logging
+from pathlib import Path
+from typing import ClassVar
+
+import numpy as np
+
+from .interface import InputSource, InputSourceInfo
+
+logger = logging.getLogger(__name__)
+
+
+class VideoFileInputSource(InputSource):
+    """Input source that reads video frames from a local file.
+
+    Uses PyAV (FFmpeg wrapper) to decode frames sequentially.
+    Loops back to the beginning when the video ends.
+    """
+
+    source_id: ClassVar[str] = "video_file"
+    source_name: ClassVar[str] = "Video File"
+    source_description: ClassVar[str] = (
+        "Read video frames from a local file (MP4, AVI, MOV, MKV, WebM)."
+    )
+
+    def __init__(self):
+        self._container = None
+        self._stream = None
+        self._frame_iter = None
+        self._connected = False
+        self._file_path: str | None = None
+
+    @classmethod
+    def is_available(cls) -> bool:
+        """Always available (PyAV is a core dependency via aiortc)."""
+        try:
+            import av  # noqa: F401
+
+            return True
+        except ImportError:
+            return False
+
+    def list_sources(self, timeout_ms: int = 5000) -> list[InputSourceInfo]:
+        """List video files from the assets directory."""
+        try:
+            from scope.server.file_utils import VIDEO_EXTENSIONS, iter_files
+            from scope.server.models_config import get_assets_dir
+
+            assets_dir = get_assets_dir()
+            sources = []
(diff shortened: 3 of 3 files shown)


src/scope/server/app.py
```diff
@@ -942,6 +922,268 @@ async def close_webrtc_session(
         raise HTTPException(status_code=500, detail=str(e)) from e
 
 
+@app.post("/api/v1/session/parameters")
```
leszko: I wonder if we should not move it all these functions into a separate router. Plus have them enabled only if daydream-scope is started with the `--mcp` flag. The reason why I mention this is that now w introduce new functions to modify the params (and other data), which brings complexity. E.g. soon we'll not have a concept of broadcasting parameters, because params will be per node.

src/scope/server/app.py
```diff
@@ -942,6 +922,268 @@ async def close_webrtc_session(
...
+    """Get the current runtime parameters from all active sessions.
+
+    Returns the merged parameter state from each session's frame processor.
+    Includes both WebRTC and headless sessions.
+    """
+    all_params = {}
+    count = 0
+    for _sid, fp, _headless in webrtc_manager.iter_frame_processors():
+        all_params.update(fp.parameters)
+        count += 1
```
leszko: I think the `count` does not make much sense. We don't support multiple sessions, so this one will always be `1`.

src/scope/server/app.py
```diff
@@ -942,6 +922,268 @@ async def close_webrtc_session(
...
+        )
+    try:
+        await webrtc_manager.remove_headless_session(session_id)
+        return {"status": "ok", "message": f"Session {session_id} stopped"}
+    except Exception as e:
+        logger.error(f"Error stopping headless session {session_id}: {e}")
+        raise HTTPException(status_code=500, detail=str(e)) from e
+
+
+@app.post("/api/v1/pipeline/unload")
```
leszko: Why do we need an endpoint for the pipeline unload?

src/scope/server/app.py
```diff
@@ -942,6 +922,268 @@ async def close_webrtc_session(
...
+    """Unload all currently loaded pipelines and free GPU memory."""
+    try:
+        pipeline_manager.unload_all_pipelines()
+        return {"status": "ok", "message": "All pipelines unloaded"}
+    except Exception as e:
+        logger.error(f"Error unloading pipelines: {e}")
+        raise HTTPException(status_code=500, detail=str(e)) from e
+
+
+@app.post("/api/v1/session/{session_id}/recording/start")
```
leszko: 1. What's the use for recording start/stop?
2. Why do we use `session_id` param here while for other endpoints we don't have `session_id`? In general I don't think it makes much sense to have session_id since we don't support multi-session setup.

src/scope/server/mcp_server.py
```diff
@@ -0,0 +1,599 @@
...
+async def _json(resp: "httpx.Response") -> str:
+    """Raise on HTTP error, then return formatted JSON."""
+    resp.raise_for_status()
+    return _fmt(resp.json())
+
+
+def create_mcp_server(base_url: str = "http://localhost:8000") -> FastMCP:
+    """Create and configure the MCP server with all Scope tools."""
+
+    mcp = FastMCP(
```
leszko, line 49: Is this FastMCP a separate process? If not, then maybe we don't need to add any new endpoints into `app.py`, but just make code calls. That would keep our API simpler.

src/scope/server/webrtc.py
```diff
@@ -645,25 +721,70 @@ async def add_ice_candidate(
             logger.error(f"Failed to add ICE candidate to session {session_id}: {e}")
             raise ValueError(f"Invalid ICE candidate: {e}") from e
 
+    def iter_frame_processors(self):
```
leszko: We don't need this. We always have just 1 frame processor.

src/scope/server/tracks.py
```diff
@@ -183,6 +183,10 @@ async def recv(self) -> VideoFrame:
 
         raise Exception("Track stopped")
 
+    def get_last_frame(self):
+        """Return the most recently rendered frame, or None."""
+        return self._last_frame
```
leszko: I guess we would need to have some lock for that.

PR #697, opened 2026-03-16
Add onboarding flow for new users

hthillman:
## Summary
- Adds multi-phase onboarding flow: inference mode selection (cloud vs local), cloud auth sign-in, starter workflow picker, and interactive workspace tour
- Backend persistence via `GET/PUT /api/v1/onboarding/status` with dual localStorage + API file storage layer
- `data-tour` attribute anchoring system with custom portal-based tour popover (avoids focus-trap conflicts with graph canvas)
- Analytics stubs and "Show onboarding again" reset in Settings > General

## Remaining TODOs
- [ ] Update welcome/inference step copy to final text
- [ ] Wire real workflow JSON into the 3 starter workflow cards
- [ ] Fix thumbnail images rendering as black
- [ ] Delay tour start until workflow import completes
- [ ] Remove "Skip for now" from CloudAuthStep (all users must sign in)
- [ ] Auto-enable remote inference after cloud sign-in

## Test plan
- [ ] Fresh user (no `~/.daydream-scope/onboarding.json`) sees full onboarding flow
- [ ] Returning user skips directly to workspace
- [ ] Cloud auth → local toggle paths both work
- [ ] Workflow picker imports selected workflow into graph
- [ ] Tour highlights correct elements in sequence
- [ ] "Show onboarding again" in Settings resets flow
- [ ] Build passes (`npm run build`)

### DIFF
--- a/frontend/src/components/onboarding/starterWorkflows.ts
+++ b/frontend/src/components/onboarding/starterWorkflows.ts
@@ -22,12 +22,12 @@ export interface StarterWorkflow {
 
 /** Return workflows filtered for the given onboarding style. */
 export function getWorkflowsForStyle(
-  style: "teaching" | "simple" | "local" | null,
+  style: "teaching" | "simple" | "local" | null
 ): StarterWorkflow[] {
   // Default to teaching workflows when no style is set
   const effectiveStyle = style ?? "teaching";
   return STARTER_WORKFLOWS.filter(
-    wf => wf.onboardingStyle === effectiveStyle || wf.onboardingStyle === "both",
+    wf => wf.onboardingStyle === effectiveStyle || wf.onboardingStyle === "both"
   );
 }
 
@@ -235,7 +235,7 @@ export const STARTER_WORKFLOWS: StarterWorkflow[] = [
                 label: "Note",
                 nodeType: "note",
                 noteText:
-                  "Finally, click the Play button in the upper left corner.\n\nAfter 30-45 seconds, generation will begin playing in the Sink node.",
+                  "Finally, click the Run button in the upper right corner.\n\nAfter 30-45 seconds, generation will begin playing in the Sink node.",
                 locked: true,
                 pinned: true,
               },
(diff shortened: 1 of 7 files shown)


src/scope/server/app.py
```diff
@@ -2957,6 +2958,73 @@ async def get_cloud_stats(
...
+
+class OnboardingStatusUpdate(BaseModel):
+    completed: bool | None = None
+    inference_mode: str | None = None
+    onboarding_style: str | None = None
+    referral_source: str | None = None
+    use_case: str | None = None
+
+
+@app.get("/api/v1/onboarding/status", response_model=OnboardingStatusResponse)
```
leszko, line 2986: @hthillman I wonder, wouldn't it be better to not store the info about the onboarding in the backend server, but rather store it in the frontend browser data? We currently do this for Graph caching, so when a user closes the app and opens again, they see the same graph as before. Wdyt?

frontend/src/components/onboarding/starterWorkflows.ts
```diff
@@ -0,0 +1,1665 @@
...
+              type: "lora",
+              position: { x: -119.22, y: 528.92 },
+              width: 240,
+              height: 293,
+              data: {
+                label: "LoRA",
+                nodeType: "lora",
+                loras: [
+                  {
+                    path: "/tmp/.daydream-scope/assets/lora/diffslime_acidzlime-000016.safetensors",
```
leszko: This should not contain paths like this.

Also I wonder that maybe instead of hardcoding the workflows here, we should take them from https://app.daydream.live/ Wdyt?

hthillman: @leszko 

1) we need the "teaching" workflows, which... I guess we could put on daydream.live?
2) Whoops, nice catch

src/scope/server/app.py
hthillman: @leszko Would we ever want a user to see the onboarding again? what if browser data gets cleared?

frontend/src/components/onboarding/starterWorkflows.ts
hthillman: My other worry about pulling from daydream.live - what if daydream.live is ever down? would that affect local experience?

src/scope/server/app.py
hthillman: Showing onboarding again is a worse UX than the graph cache case, where re-showing a graph is often helpful

PR #738, opened 2026-03-25
Livepeer Cloud Mode

j0sh:
Adds a new cloud mode running on the Livepeer network. This can work both standalone (for local development) and with Fal Serverless.

```mermaid
graph LR
    A["Local Scope"] -- Trickle --> B[Orchestrator]
    B -- WebSocket --> C["Scope Runner (Fal)"]
    B -- Trickle --> C["Scope Runner (Fal)"]
```

This is still an early build. Posting this now to avoid additional reconciliation work from upstream merges, however the changes to existing code should be minimal outside of some small cloud proxy routing adjustments to make Livepeer and the existing CloudConnectionManager interchangeable, plus type annotation updates to cover both modes.

The existing Cloud Relay mode will remain the default for now. Opt-in to Livepeer Cloud Mode via

```
SCOPE_CLOUD_MODE=livepeer LIVEPEER_ORCH_URL=... uv run daydream-scope
```

Also see the PR preview deployment notes for instructions on how to test.

The models I have tested on Fal seem to work okay, as do LoRAs.

Not yet implemented and will come in follow-up PRs:
* ~~Plugins~~
* Recordings
* Audio
* ~~Log forwarding~~
* Kafka reporting
* ~~Disk state cleanup between sessions~~

There is also a documentation page on basic usage, an architecture doc for Livepeer Cloud Mode, and a new `~/.agents` folder with a SKILLS.md for Fal which has been handy for debugging deployment issues.

### DIFF
--- a/src/scope/server/livepeer_client.py
+++ b/src/scope/server/livepeer_client.py
@@ -280,7 +280,7 @@ class LivepeerClient:
             raise RuntimeError("stream_started response missing in/out channels")
 
         publisher = MediaPublish(input_url, config=MediaPublishConfig(fps=self._fps))
-        media_output = MediaOutput(output_url)
+        media_output = MediaOutput(output_url, start_seq=-1)
         self._media_connected = True
         subscriber = asyncio.create_task(self._receive_loop(media_output))
 
--- a/uv.lock
+++ b/uv.lock
@@ -1332,7 +1332,7 @@ wheels = [
 [[package]]
 name = "livepeer-gateway"
 version = "0.1.0"
-source = { git = "https://github.com/livepeer/livepeer-python-gateway#5007e2cd9bbed0ed39fbb9ed7f5afffdcbaf348f" }
+source = { git = "https://github.com/livepeer/livepeer-python-gateway#a4670584558f607f418f855b0c849bfdb302dcf2" }
 dependencies = [
     { name = "aiohttp" },
     { name = "av" },


src/scope/server/livepeer.py
```diff
@@ -0,0 +1,288 @@
...
+    ) -> None:
+        """Create and connect a persistent Livepeer LV2V job."""
+        # Keep connect signature compatible with cloud-style connect requests.
+        # Livepeer token-based auth ignores api_key, while app_id can be used as
+        # optional runner routing config (derived into a fal ws_url in the client).
+        _ = api_key
+        self._user_id = user_id
+
+        if self.is_connected:
+            return
```
ryanontheinside, line 100: **Bug: `_connecting` stuck after double-connect**

If `/cloud/connect` is called while already connected, `connect_background()` sets `self._connecting = True` (line 171), then `connect()` early-returns here before the `try/finally` block that resets it (lines 111-142). `_connecting` stays `True` permanently. The `/cloud/status` endpoint then reports `connected: true, connecting: true` indefinitely.

Could we reset `_connecting` before the early return, or guard in `connect_background()` so it doesn't set the flag when already connected?

src/scope/server/livepeer_client.py
```diff
@@ -0,0 +1,745 @@
...
+        )
+        if self._ws_url and "ws_url" not in params:
+            params["ws_url"] = self._ws_url
+
+        request = StartJobRequest(
+            model_id=self._model_id,
+            params=params or None,
+        )
+
+        self._job = await asyncio.to_thread(
```
ryanontheinside, line 130: **Reconnect after disconnect: runner session not cleaned up**

After disconnect + reconnect, the orchestrator returns HTTP 500: `"unexpected message type "error" between ready and started"`. The error persists across retries, so once a session goes stale on the runner side, the client can't recover without the runner being recycled. Would retry-with-backoff or a session cleanup request before reconnecting help here?

Also, during the successful connection, `livepeer_gateway` logs: `"No running event loop; per-segment payments not started"`. May be intentional for now but worth a TODO?

src/scope/server/livepeer_client.py
```diff
@@ -0,0 +1,745 @@
...
+        except asyncio.CancelledError:
+            pass
+        except Exception as e:
+            unexpected_reason = f"Livepeer events loop failed: {e}"
+            logger.error(f"Events loop failed: {e}")
+        finally:
+            if self._connected and not self._shutdown_started:
+                if unexpected_reason is None:
+                    unexpected_reason = "Livepeer events loop stopped unexpectedly"
+                await self._shutdown(
```
ryanontheinside, line 453: **Events loop failure triggers full shutdown without reconnect attempt**

If the events loop encounters an error or stops unexpectedly, `_shutdown()` is called immediately. Combined with `stop_media()` being called from the output loop on unexpected exit (line 389), this can cascade. A transient network blip would tear down the entire session rather than attempting reconnection. Could this try a reconnect before going to full shutdown?

src/scope/cloud/livepeer_app.py
```diff
@@ -0,0 +1,1034 @@
...
+                if cdn_result.error_response is not None:
+                    return cdn_result.error_response
+                response = await client.post(
+                    path,
+                    content=cdn_result.content,
+                    headers={"Content-Type": content_type},
+                    timeout=60.0,
+                )
+            elif is_binary_upload:
+                binary_content = base64.b64decode(body["_base64_content"])
```
ryanontheinside, line 364: **No size limits on base64 content**

`base64.b64decode(body["_base64_content"])` has no size check. A large payload could exhaust memory before reaching the API handler. Would it make sense to cap the encoded length before decoding?

src/scope/server/cloud_proxy.py
```diff
@@ -383,6 +390,12 @@ async def wrapper(*args: Any, **kwargs: Any) -> Any:
             if path is None:
                 actual_path = http_request.url.path
             elif callable(path):
+                if not isinstance(cloud_manager, CloudConnectionManager):
```
ryanontheinside, line 393: **Non-blocking: Recording download blocked in Livepeer mode**

This `isinstance` check raises a 400 for Livepeer. There's a TODO here already, just flagging for tracking.

src/scope/server/livepeer_client.py
j0sh: From what I can gather from the orchestrator logs, this seems like a user error, because the Daydream user ID was not being sent. I am curious how you were triggering it?

Disconnect + reconnect with a properly formed request works OK for me. 

I've had a TODO here specifically to improve orchestrator behavior around handshake / auth failures (eg, returning a 403 instead of a 500) but was planning to do that as a follow-up.

src/scope/server/livepeer_client.py
j0sh: Retries are handled at lower layers ... there's probably still a few things I should do there, but if we hit this point of exiting the event or media loops, we've determined they are pretty much unrecoverable.

src/scope/cloud/livepeer_app.py
j0sh: The event payload where this comes from is limited to 1 MB ([source](https://github.com/livepeer/livepeer-python-gateway/blob/main/src/livepeer_gateway/channel_reader.py#L94))

src/scope/server/livepeer.py
j0sh: Good spot, fixed in https://github.com/daydreamlive/scope/pull/738/commits/6611bf3f48a11c8108a485c1519499c5e062cc9f

PR #877, opened 2026-04-08
Post-session NPS survey

hthillman:
## Summary

Implements DAY-56 — post-session survey UX for PMF + advocacy signal. Splits the combined Sean Ellis + NPS flow into **two independent surveys** with proper eligibility gating, denominator tracking, and separated visual design.

### Why split them

A combined Sean Ellis → NPS flow contaminates both metrics: Sean Ellis's emotional "how disappointed would you be" framing primes the NPS response (response-set bias). Coupling them to the same one-shot lifecycle also makes NPS impossible to run as a recurring pulse.

### What ships

**Two surveys, two lifecycles:**
- **Sean Ellis (PMF snapshot)** — one-shot per install. Vertical stack of 3 outline buttons (`max-w-md` modal).
- **NPS (recurring pulse)** — cooldown-gated. Horizontal 0–10 bar, connected radii, `flex-nowrap` so it never wraps, hover-only red/yellow/green bucket tints (`max-w-lg` modal).

**"Meaningful use" gating** — surveys only fire for users who've actually experienced the product. Defined as a stream session ≥ 30s. Tracked with two `localStorage` keys:
- `scope_first_meaningful_use_at` — written once, never overwritten
- `scope_last_meaningful_use_at` — updated every qualifying session

**Independent eligibility rules:**

| | Sean Ellis | NPS |
|---|---|---|
| Min install age | 45 days since first meaningful use | 60 days |
| Active window | Used in last 14 days | Last 14 days |
| Cadence | One-shot (`scope_sean_ellis_shown_at`) | Every 90 days (`scope_nps_last_shown_at`) |

If both are eligible the same session, Sean Ellis wins; NPS defers to the next eligible stop. Never stacked back-to-back.

**Telemetry-gated at two layers** — surveys never render for opted-out users:
1. `isTelemetryEnabled` is the first check in both `isSeanEllisEligible` / `isNpsEligible` — short-circuits before any other condition
2. `trackEvent` routes through the telemetry module, which itself gates on opt-in

**Denominator tracking** — every modal lifecycle now fires an event so response rate and drop-off are measurable:
- `sean_ellis_shown` / `sean_ellis_response {response}` / `sean_ellis_skipped`
- `nps_shown` / `nps_response {score}` / `nps_skipped`

**Other polish:**
- 750ms delay after stream stop so the modal doesn't interrupt the natural review moment
- `role="radiogroup"` + `aria-label` on both; NPS anchors wired via `aria-describedby`
- All `localStorage` access wrapped in try/catch (Safari private mode / quota safety)
- Eligibility rules extracted to pure functions in `frontend/src/lib/surveyEligibility.ts`

### Files

- **New:** `frontend/src/lib/surveyEligibility.ts` — pure eligibility rules + safe localStorage helpers + tunable constants
- **Rewritten:** `frontend/src/components/SurveyModal.tsx` — `SurveyShell` primitive + `SeanEllisSurvey` + `NpsSurvey`
- **Modified:** `frontend/src/pages/StreamPage.tsx` — meaningful-use tracking on stream stop + two independent eligibility effects

## Test plan

Time-gated flows require `localStorage` overrides to test end-to-end:

- [ ] **Meaningful use** — stop stream in <30s, confirm no `scope_first_meaningful_use_at`. Stream ≥30s, confirm both keys set; repeat and confirm first-use unchanged but last-use advances.
- [ ] **Sean Ellis eligible** — set first-use to 46d ago, last-use to now, telemetry on → modal appears next stream stop, fires `sean_ellis_shown` then `sean_ellis_response`. Next stop: does NOT reappear.
- [ ] **Sean Ellis too young** — set first-use to 10d ago → modal does NOT appear.
- [ ] **Sean Ellis inactive** — set last-use to 20d ago → modal does NOT appear.
- [ ] **NPS eligible** — set first-use to 61d ago, last-use recent, no NPS shown flag → NPS appears, fires `nps_shown` then `nps_response`.
- [ ] **NPS cooldown** — set `scope_nps_last_shown_at` to 89d ago → no show. 91d ago → shows again.
- [ ] **Both eligible same session** — clear both shown flags, first-use 70d ago → Sean Ellis shows this stop; NPS shows on next eligible stop.
- [ ] **Telemetry opt-out** — flip telemetry off → neither survey appears regardless of flags.
- [ ] **Skip / dismiss** — skip or ESC each flow → `_skipped` event fires, `_response` does not.
- [ ] **Delayed open** — modal appears ~750ms after stream stop, not instantly.
- [ ] **NPS layout** — 11 buttons on one row, never wraps at `max-w-lg`; hover tints: 0–6 red, 7–8 yellow, 9–10 green.
- [ ] `cd frontend && npm run lint && npm run format:check && npm run build` pass.

Closes [DAY-56](/DAY/issues/DAY-56)

### DIFF
--- a/frontend/src/components/SurveyModal.tsx
+++ b/frontend/src/components/SurveyModal.tsx
@@ -1,4 +1,4 @@
-import { useEffect, type ReactNode } from "react";
+import { type ReactNode } from "react";
 import {
   Dialog,
   DialogContent,
@@ -22,10 +22,6 @@ interface SurveyShellProps {
   title: string;
   description: string;
   size?: ShellSize;
-  /** Event name to fire when the modal opens. */
-  shownEvent?: string;
-  /** Event name to fire when the user skips/dismisses without answering. */
-  skippedEvent?: string;
   children: ReactNode;
 }
 
@@ -35,26 +31,15 @@ function SurveyShell({
   title,
   description,
   size = "md",
-  shownEvent,
-  skippedEvent,
   children,
 }: SurveyShellProps) {
-  // Fire _shown on open so we have a denominator for response rates.
-  useEffect(() => {
-    if (open && shownEvent) {
-      trackEvent(shownEvent);
-    }
-  }, [open, shownEvent]);
-
+  // Intentionally no telemetry on open, skip, or dismiss. The only event fired
+  // by surveys is a single `survey` event on answer (see child components).
   const handleOpenChange = (isOpen: boolean) => {
-    if (!isOpen) {
-      if (skippedEvent) trackEvent(skippedEvent);
-      onClose();
-    }
+    if (!isOpen) onClose();
   };
 
   const handleSkip = () => {
-    if (skippedEvent) trackEvent(skippedEvent);
     onClose();
   };
 
@@ -119,7 +104,7 @@ interface SurveyProps {
 
 export function SeanEllisSurvey({ open, onClose }: SurveyProps) {
   const handleSelect = (value: SeanEllisResponse) => {
-    trackEvent("sean_ellis_response", { response: value });
+    trackEvent("survey", { name: "sean_ellis", response: value });
     onClose();
   };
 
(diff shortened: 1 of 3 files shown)


leszko (approved):
Pushed a small fix, other than that LGTM. Feel free to merge.

PR #901, opened 2026-04-10
Propagate timestamps end-to-end through the media graph

j0sh:
Scope currently guesses timestamps in several places — sometimes multiple times for the same sample — which accumulates error and inflates perceived latency even on trivial pipelines like gray.

This PR introduces timestamp-aware media packet wrappers (`VideoPacket`, `AudioPacket`) that carry payloads alongside `pts` and `time_base`, and threads them from WebRTC source capture through frame processing, pipeline execution, sink delivery, WebRTC, and Livepeer output.

Pipelines can now read input timestamps via the video_timestamps kwarg and write output timestamps via the video_timestamps dict key (passthrough is implemented as an example). If a pipeline doesn't produce timestamps, Scope falls back to its existing guessing behavior, so nothing breaks.

### Why this matters beyond accuracy

* **Timestamps become a control signal in the graph.** A pipeline node can read, respond to, and emit timestamps the same way it transforms pixels — making timing a first-class, composable element of the processing graph rather than an afterthought reconstructed at various edges.

* **Downstream cadence control.** By adjusting output timestamps, a pipeline can deliberately stretch or squeeze playback pacing — useful for slow-motion effects, time-lapse, or graceful catch-up after a processing stall — without any downstream consumer needing special knowledge.

* **Eliminates compounding drift.** A single authoritative timestamp per sample replaces multiple independent guesses, removing the cumulative error that shows up as growing latency or jitter over long sessions.

* **Simpler latency measurement.** With real timestamps on every packet, you can diff source-to-sink PTS at any tap point in the graph instead of reverse-engineering timing from wall-clock heuristics.

* **Enables timestamp-dependent pipeline logic.** Pipelines that need to reason about time — frame interpolation, temporal filtering, A/V sync, adaptive rate control — now have the data they need without inventing their own clock.

Existing pipelines and tensor/tuple queue paths continue to work unchanged. Full migration will be incremental and this PR is scoped to the only the graph plumbing, with WebRTC and Livepeer Cloud. Follow-on work will pull in audio, recording, headless mode, API video inputs, maybe Cloud Relay, and additional built-in pipelines.


leszko (approved):
LGTM. Though I haven't seen the exactly synced audio/video when trying it out with ltx-2. But it may be not related to this PR, so not a blocker for merging.

Related Discord thread: https://discord.com/channels/1368975848123666553/1493146205704818748/1493254199608213554

PR #903, opened 2026-04-10
feat(backend): NodeProcessor runtime for custom-node graphs

leszko:
## Summary

Adds a **`NodeProcessor`** so the graph executor can schedule `type: "node"` backends (registered via `NodeRegistry`) alongside pipeline nodes in the same graph. Makes `BaseNode.execute(inputs, **kwargs)` the execution contract for plain backend nodes — `Pipeline` subclasses inherit a default that raises, since they're driven by `PipelineProcessor.__call__` instead. Updates session/WebRTC/MCP paths to allow **node-only graphs** (no pipeline load required).

This is runtime/plumbing only. No built-in nodes, audio routing, or WebRTC audio changes are included in this PR.

---

## Node runtime

- `core/nodes/base.py` — `BaseNode.execute(inputs, **kwargs) -> dict` is now a defined method with a `NotImplementedError` default. Plain nodes override it; `Pipeline` subclasses keep the default and are invoked via `__call__`. `node_id` default added so nodes can be instantiated without positional args.
- `core/nodes/processor.py` — new `NodeProcessor`: daemon worker thread (`NodeProcessor[<node_id>]`), per-port input/output queues, fan-out routing with back-pressure (`oq.put(value, timeout=0.1)`), source-once vs. `continuous=True` semantics, error-handling that stops runaway source retries, and `PipelineProcessor`-compatible surface (`output_queue`, `output_consumers`, `paused`, `update_parameters`, `set_beat_cache_reset_rate`, `get_fps`).

## Graph execution

- `server/graph_executor.py`
  - Instantiates `NodeProcessor` for `type="node"` nodes; looks up class via `NodeRegistry.get(node.node_type_id)` and raises on unknown types.
  - Merges per-node `params` from the workflow over `initial_parameters` (per-node values win).
  - Node→node stream edges use `maxsize=1` queues so the DAG executes one tick at a time and large tensors don't pile up; pipeline edges keep the default size.
  - `_validate_edge_ports` extended to validate custom-node ports via `NodeRegistry` definitions.
  - Output wiring loop now accepts both `pipeline` and `node` producers.

## Session / routing plumbing

- `server/frame_processor.py` — local-mode start no longer fails when `pipeline_ids` is empty if the graph has `type="node"` nodes.
- `server/mcp_router.py` — graph may be pipeline-only, custom-node-only, or mixed. Local `load_pipelines` is skipped when there are no pipeline tuples; error message updated.
- `server/app.py` — WebRTC offer path skips the "pipeline must be loaded" check when the incoming graph contains custom nodes.

## Frontend

- `frontend/src/pages/StreamPage.tsx` — skips the pipeline load step when `loadItems.length === 0` so node-only graphs can start.

---

## Test plan

- [ ] Mixed graph (pipeline + custom node on a supported port) — session starts, both processors run, frames flow.
- [ ] Node-only graph — session starts without loading any pipeline; WebRTC offer succeeds; frontend skips the load step.
- [ ] Unknown `node_type_id` — graph build raises a clear `ValueError` naming the node.
- [ ] Edge with a port not declared by the node definition — `_validate_edge_ports` rejects it with a descriptive error.
- [ ] Per-node `params` from workflow override `initial_parameters` inside the node's `execute()` kwargs.
- [ ] Continuous node — re-executes each tick; source (no inputs) node — executes once; failing source node doesn't spin.

### DIFF
--- a/frontend/src/pages/StreamPage.tsx
+++ b/frontend/src/pages/StreamPage.tsx
@@ -2821,10 +2821,13 @@ export function StreamPage() {
         );
       }
 
-      const loadSuccess = await loadPipeline(loadItems);
-      if (!loadSuccess) {
-        console.error("Failed to load pipeline, cannot start stream");
-        return false;
+      // Node-only graphs (no pipeline nodes) skip the pipeline load step.
+      if (loadItems.length > 0) {
+        const loadSuccess = await loadPipeline(loadItems);
+        if (!loadSuccess) {
+          console.error("Failed to load pipeline, cannot start stream");
+          return false;
+        }
       }
 
       // Check video requirements based on input mode.
--- a/src/scope/core/nodes/base.py
+++ b/src/scope/core/nodes/base.py
@@ -110,15 +110,32 @@ class BaseNode(ABC):
     """Abstract base class for all backend node types.
 
     Subclasses must set ``node_type_id`` as a ``ClassVar`` and implement
-    ``get_definition()``. Execution contracts (e.g. ``execute(inputs)`` for
-    pull-based execution, or ``setup(emit_output) / update_input(...)`` for
-    event-driven execution) are defined by concrete execution backends and
-    not by this base class.
+    ``get_definition()`` and ``execute()``. Nodes run inside a
+    :class:`NodeProcessor` in the pipeline graph: input port values arrive
+    in the ``inputs`` dict and the return dict fans out to downstream
+    nodes or pipelines via queue edges.
+
+    Example::
+
+        class MyNode(BaseNode):
+            node_type_id: ClassVar[str] = "my-plugin.my-node"
+
+            @classmethod
+            def get_definition(cls) -> NodeDefinition:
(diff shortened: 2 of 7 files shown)


src/scope/server/pipeline_processor.py
```diff
@@ -84,6 +84,10 @@ def __init__(
         # record_queues_by_node). Updated by _resize_output_queue so cached
         # references stay in sync when a queue object is replaced.
         self.external_queue_refs: list[tuple[dict, str]] = []
+        # Input ports that carry audio (tensor, sample_rate) tuples instead
+        # of video frame tensors. Set by graph_executor when wiring audio
+        # edges so audio inputs don't get accumulated into a video chunk.
+        self.audio_input_ports: set[str] = set()
```
gioelecerati: Is this supposed to get consumed in prepare_chunk and prepare_multi_chunk?

gioelecerati (approved):

PR #904, opened 2026-04-10
feat(nodes): add scheduler built-in node with dynamic trigger outputs

leszko:
## Summary

Adds **`SchedulerNode`** (`node_type_id`: `scheduler`) — a time-based trigger sequencer registered via `register_builtin_nodes()` — and wires it end-to-end through the backend graph execution runtime and the frontend workflow builder. Stacks on top of `rafal/node-backend-execution-2`.

The node runs a daemon thread driving a ~200 Hz monotonic clock (`_TICK_INTERVAL = 5 ms`), fires named triggers at configured times, and emits incrementing counters per trigger so downstream nodes never miss a firing.

## Ports & parameters

| | |
|--|--|
| **Inputs** | `start` (trigger — toggles play/pause), `reset` (trigger — clears fired state / elapsed) |
| **Static outputs** | `tick`, `elapsed`, `is_playing` |
| **Dynamic outputs** | One counter output per unique `port_name` in `triggers`, declared via `get_dynamic_output_ports()` so graph validation accepts edges from them |
| **Params (declared as `NodeParam`)** | `triggers` (list of `{time, port_name}`), `loop`, `duration` |

## Behaviour

- **Auto-start**: If `triggers` is non-empty and nothing has started yet, the first `execute()` arms the timer thread without requiring a `start` pulse. Guarded by an explicit `_auto_start_done` flag so a stale zero on `start`/`reset` cannot re-arm it.
- **Edge-triggered start/reset**: Actions fire on strict counter increments, so stale values on the input queue do not retoggle. Prev counters seed at 0 so the first positive pulse still fires.
- **Pause broadcast**: `is_playing` is staged into pending outputs on both pause and auto-start, so downstream sees the transition without waiting for the elapsed heartbeat.
- **Loop / duration**: When `duration > 0` and `loop` is true, state resets at `elapsed >= duration`. When not looping, playback stops after all triggers have fired.
- **Idle pacing**: `execute()` returns `{}` when nothing changed and relies on `NodeProcessor`'s built-in idle handling instead of busy-looping.

## Runtime / framework changes

- **`BaseNode`**: adds `get_dynamic_output_ports()` and `shutdown()` hooks.
- **`graph_executor._validate_edge_ports`**: unions static + dynamic outputs so dynamic per-trigger ports pass edge validation.
- **`NodeProcessor.stop()`**: calls `node.shutdown()` during teardown so the scheduler's timer thread is joined cleanly.

## Frontend wiring

- **`graphUtils`**: drops `scheduler` from `FRONTEND_ONLY_TYPES`, emits it as `type=node` / `node_type_id=scheduler` with params, and rehydrates it back into the native scheduler `nodeType` on graph load so the bespoke widget survives a round-trip.
- **`AddNodeModal`**: skips `scheduler` in the discovered Plugins list to avoid duplicating the existing catalog entry.

## Files

- `src/scope/core/nodes/builtins/scheduler.py` — implementation
- `src/scope/core/nodes/builtins/__init__.py`, `src/scope/core/nodes/__init__.py` — export + `NodeRegistry.register`
- `src/scope/core/nodes/base.py` — `get_dynamic_output_ports()` / `shutdown()` hooks
- `src/scope/core/nodes/processor.py` — invoke `shutdown()` on stop
- `src/scope/server/graph_executor.py` — union dynamic outputs during edge validation
- `frontend/src/lib/graphUtils.ts` — round-trip scheduler through backend node format
- `frontend/src/components/graph/AddNodeModal.tsx` — de-dupe scheduler in plugins list

## Test plan

- [ ] Drop a scheduler into the workflow builder with `duration=5`, `loop=true`, multiple `triggers`; confirm fires at expected times.
- [ ] Wire `tick` / `elapsed` / `is_playing` downstream and confirm values update live.
- [ ] Wire a per-trigger counter output downstream and confirm graph validation accepts the edge.
- [ ] Pulse `reset` while playing and when stopped; confirm elapsed / fired state clears.
- [ ] Toggle `start` repeatedly; confirm play/pause semantics and that `is_playing` broadcasts immediately.
- [ ] Stop the graph; confirm the scheduler's timer thread exits (no leaked daemon).
- [ ] Reload a graph containing a scheduler and confirm the node rehydrates with its widget and params intact.

### DIFF
--- a/frontend/src/components/graph/AddNodeModal.tsx
+++ b/frontend/src/components/graph/AddNodeModal.tsx
@@ -414,10 +414,13 @@ export function AddNodeModal({
         if (controller.signal.aborted) return;
         // The unified endpoint returns both pipelines (pipeline_meta != null)
         // and plain custom nodes. Pipelines are still added via the hardcoded
-        // "Pipeline" catalog entry (placeholder + dropdown), so we filter
-        // them out of the plugin listing here to avoid duplication.
+        // "Pipeline" catalog entry (placeholder + dropdown); the scheduler
+        // has its own catalog entry with a bespoke widget. Filter both out
+        // of the plugin listing to avoid duplication.
         const items: NodeCatalogItem[] = (data.nodes ?? [])
-          .filter(n => n.pipeline_meta == null)
+          .filter(
+            n => n.pipeline_meta == null && n.node_type_id !== "scheduler"
+          )
           .map(n => ({
             type: "custom_node" as const,
             subType: n.node_type_id,
--- a/frontend/src/lib/graphUtils.ts
+++ b/frontend/src/lib/graphUtils.ts
@@ -846,14 +846,40 @@ export function graphConfigToFlow(
             style: { width: n.w ?? undefined, height: n.h ?? undefined },
           }
         : {};
+    const position = {
+      x: savedX !== undefined ? savedX : START_X + COLUMN_GAP * 1.5,
+      y: savedY !== undefined ? savedY : START_Y + i * (NODE_HEIGHT + ROW_GAP),
+    };
+    // Scheduler has its own React renderer — rehydrate into its native
+    // nodeType so the bespoke widget (trigger list, transport controls)
+    // comes back after a round-trip, instead of the generic custom_node UI.
+    if (n.node_type_id === "scheduler") {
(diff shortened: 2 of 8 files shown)


gioelecerati (approved):
Working great for me

PR #906, opened 2026-04-10
feat: unified node abstraction with custom node catalog

leszko:
## Summary

Introduces the fine-grained backend node framework and frontend catalog that surface plugin-defined processing nodes in the graph editor. Pipelines are folded into the same `BaseNode` hierarchy so plugin authors and users see one concept. This is the metadata/discovery foundation for the execution-scheduler and ACEStep follow-ups; execution contracts are layered on by those branches.

## Backend

- **`src/scope/core/nodes/`** (new): `BaseNode` ABC, Pydantic schemas (`NodeDefinition` / `NodePort` / `NodeParam`), and a `NodeRegistry` keyed by `node_type_id`. Widget hints (number `min`/`max`/`step`, select `options`, …) live in a free-form `NodeParam.ui: dict | None` so the base schema doesn't grow as new widget kinds appear.
- **Pipeline ↔ Node unification**: `Pipeline` is now `class Pipeline(BaseNode, ABC)` — pure inheritance change. Concrete pipeline subclasses (longlive, krea, LTX-2, VACE, streamdiffusion, …) touch zero lines. `Pipeline.get_definition()` projects the config class into a `NodeDefinition` and stuffs the full `get_schema_with_metadata()` output into `pipeline_meta`.
- **`PipelineRegistry`** becomes a filtering view over `NodeRegistry`. Every pre-existing call site (`list_pipelines`, `get_config_class`, `chain_produces_*`, `register`, `unregister`, `is_registered`, `get`) keeps its old signature so the server, plugin manager, OSC/DMX docs, workflows resolver, and tests work unchanged. Built-in pipelines route through `NodeRegistry.register` so they share the same id-derivation and debug-log path as plugin nodes, and `pipeline_id` is asserted against the config class to catch drift.
- **Plugins**: new `register_nodes` pluggy hookspec + `PluginManager.register_plugin_nodes` hook caller. Existing `register_pipelines` plugins keep working unchanged — both hooks plant into the unified storage.
- **`graph_schema`**: allow `type="node"` `GraphNode` with `node_type_id` and `params`, a `get_backend_node_ids()` helper, and a validation error when `node_type_id` is missing. Existing `"type": "pipeline"` workflows still load.
- **`GET /api/v1/nodes/definitions`**: the canonical discovery endpoint. Pipelines carry the full config schema as `pipeline_meta`; plain custom nodes leave `pipeline_meta` `None`. `/api/v1/pipelines/schemas` becomes a thin compat alias derived from the same registry, so `usePipelines.ts` and cloud-mode proxy callers keep working without migration.

`BaseNode` intentionally only requires `get_definition()`; execution contracts (push vs. pull) are layered by the specialized branches.

## Frontend

- **`api.ts`**: extend `GraphNode` with `node_type_id` / `params` and add `NodeDefinitionDto`, `NodePortDef`, `NodeParamDef`, and `fetchNodeDefinitions({ signal? })`.
- **`graphUtils.ts`**: add `custom_node` to `FlowNodeData.nodeType` (typed via the canonical `NodePortDef` / `NodeParamDef`) and round-trip `custom_node` ↔ backend `type: "node"` in `graphConfigToFlow` / `flowToGraphConfig`.
- **`CustomNode.tsx`** (new): schema-driven React Flow node rendering input/output ports with port-type colors and ComfyUI-style param widgets (select / boolean / number / text). Number widgets read `p.ui?.min` / `max` / `step`; select widgets read `p.ui?.options`.
- **`AddNodeModal`**: fetch node definitions while open and merge them into the catalog under a new "Plugins" category. Pipelines are filtered out and continue to be added via the existing "Pipeline" placeholder + dropdown UX.
- **`useNodeFactories`**: add `custom_node` to `NodeTypeKey` / `NODE_DEFAULTS` and route the add flow through `extraData`.
- **`useGraphPersistence`**: after import/restore, hydrate every `custom_node` with its definition, preserving saved `customNodeParams` over definition defaults. Hydration runs through an `AbortController` so a stale fetch (rapid reload, unmount) cannot overwrite newer node state.
- **`connectionValidation`**: when either endpoint is a `custom_node`, enforce port-type equality; built-in node streams stay untyped.
- **`PipelineNode` / `contextMenuItems` / `GraphEditor`**: rename the user-facing "Pipeline" label to "Node" / "Model" so the word "pipeline" disappears from the UI; register `custom_node: CustomNode` in `nodeTypes`.

## Backwards compatibility

- Existing `Pipeline` subclasses don't change.
- Existing `register_pipelines` plugins don't change.
- Existing `PipelineRegistry.*` call sites don't change.
- Existing `"type": "pipeline"` saved workflows still load.
- `GraphNode.type` keeps both `"pipeline"` and `"node"` literals.

## Test plan

- [ ] Load Scope and confirm `GET /api/v1/nodes/definitions` returns every registered built-in pipeline (with `pipeline_meta` populated) plus any plugin-registered plain nodes.
- [ ] Open the Add Node modal and verify the "Plugins" category lists backend-registered nodes.
- [ ] Drop a custom node into the graph, save, reload, and verify params and port wiring are preserved.
- [ ] Connect a custom node port to another custom node port of a different `port_type` and verify the connection is rejected.
- [ ] Load an existing pre-PR workflow with `"type": "pipeline"` nodes and verify it still loads without warnings.
- [ ] Confirm an existing `register_pipelines` plugin still loads and its pipelines appear in both `/api/v1/pipelines/schemas` and `/api/v1/nodes/definitions`.

### DIFF
--- a/src/scope/core/pipelines/interface.py
+++ b/src/scope/core/pipelines/interface.py
@@ -1,12 +1,25 @@
-"""Base interface for all pipelines."""
+"""Base interface for all pipelines.
 
+A :class:`Pipeline` is a :class:`scope.core.nodes.BaseNode` subclass —
+the "heavy" kind that batches video frames, loads GPU models, and
+carries a rich Pydantic config class. The graph editor and user-facing
+docs call them *Nodes*; the name ``Pipeline`` survives as the
+implementation base class so existing subclasses and plugins keep
+working unchanged.
+"""
+
+import logging
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING
 
 from pydantic import BaseModel
 
+from scope.core.nodes.base import BaseNode, NodeDefinition, NodePort
+
 if TYPE_CHECKING:
     from .schema import BasePipelineConfig
+
+logger = logging.getLogger(__name__)
 
 
 class Requirements(BaseModel):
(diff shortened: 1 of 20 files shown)


src/scope/core/pipelines/interface.py
```diff
@@ -15,50 +25,57 @@ class Requirements(BaseModel):
...
+                NodePort(name=name, port_type="video")
+                for name in (getattr(config, "inputs", ["video"]) or ["video"])
+            ],
+            outputs=[
+                NodePort(name=name, port_type="video")
+                for name in (getattr(config, "outputs", ["video"]) or ["video"])
+            ],
+            params=[],
+            continuous=False,
+            pipeline_meta=config.get_schema_with_metadata(),
```
gioelecerati: Maybe we can have a try catch here in case of a malformed pipeline?

gioelecerati (approved):

src/scope/core/pipelines/interface.py
leszko: Added logging with warning

PR #907, opened 2026-04-10
fix: Unify cloud connect and surface connection errors (#873)

j0sh:
Onboarding's /cloud/connect POST only sent {user_id}, missing api_key. In Livepeer mode this caused cloud connections to fail silently, hanging the UI forever on "Almost there…".

- Extract shared connectToCloud() helper into lib/cloudApi.ts so all call sites (onboarding auth, onboarding connecting, settings toggle) use the same payload with both user_id and api_key

- Remove duplicate activateCloudRelay() from CloudConnectingStep and onboardingStorage

- Surface status.error in CloudConnectingStep's final waiting state instead of spinning indefinitely

(cherry picked from commit f82f01c272e466c2b205d569dc78c7a618b52b13)

### DIFF
--- a/frontend/src/components/onboarding/CloudAuthStep.tsx
+++ b/frontend/src/components/onboarding/CloudAuthStep.tsx
@@ -4,10 +4,9 @@ import { Button } from "../ui/button";
 import {
   isAuthenticated,
   getDaydreamUserDisplayName,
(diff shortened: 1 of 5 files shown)


seanhanca (approved):
Approves

PR #945, opened 2026-04-15
fix: downgrade bundled plugin uninstall from ERROR to WARNING

livepeer-tessa:
## Summary

Fixes #837 — ERROR-level log noise on every prod fal.ai job from `scope-ltx-2` bundled plugin cleanup attempts.

## Root Cause

The `uninstall_plugin` endpoint was catching all `PluginInstallError` exceptions (including bundled-plugin rejection) and logging them as `ERROR`. Attempting to uninstall a bundled plugin is an **expected, handled condition** — the fal.ai cleanup loop always tries to uninstall all plugins at job end, including bundled ones. It's not a server error.

## Changes

- **`manager.py`**: Add `PluginBundledError(PluginInstallError)` subclass so callers can distinguish bundled-rejection from genuine install failures. Raise it instead of `PluginInstallError` in `_uninstall_plugin_sync` when a bundled plugin is targeted.
- **`__init__.py`**: Export `PluginBundledError`.
- **`app.py`**: Catch `PluginBundledError` before the generic handler in the uninstall endpoint: log at `WARNING`, return HTTP **400** (client error) instead of 500 (server error).

## Before / After

| | Before | After |
|---|---|---|
| Log level | `ERROR` | `WARNING` |
| HTTP status | 500 | 400 |
| Noise in prod | 5+ per 12h window | 0 (warning-level, not ERROR-forwarded) |

## Testing

- All 99 existing plugin manager tests pass (`test_plugin_manager.py`)
- `PluginBundledError` is a subclass of `PluginInstallError` so existing `except PluginInstallError` catch sites in other paths still work correctly

### DIFF
--- a/src/scope/core/plugins/__init__.py
+++ b/src/scope/core/plugins/__init__.py
@@ -3,6 +3,7 @@
 from .hookspecs import hookimpl
 from .manager import (
     FailedPluginInfo,
+    PluginBundledError,
     PluginDependencyError,
     PluginInstallError,
     PluginInUseError,
@@ -32,4 +33,5 @@ __all__ = [
     "PluginNameCollisionError",
     "PluginDependencyError",
     "PluginInstallError",
+    "PluginBundledError",
 ]
--- a/src/scope/core/plugins/manager.py
+++ b/src/scope/core/plugins/manager.py
@@ -80,6 +80,12 @@ class PluginInstallError(Exception):
     pass
 
 
+class PluginBundledError(PluginInstallError):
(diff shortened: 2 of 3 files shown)


livepeer-tessa:
Pinging @mjh1 and @emranemran — this has been unresolved since April 3rd and is still generating ERROR-level noise across every prod job (~10 occurrences in the last 12h alone). PR #945 is ready: all CI checks green, no conflicts. Would appreciate a review so we can cut the monitoring noise. 🙏

PR #955, opened 2026-04-16
Add YouTube input source

hthillman:
Adds a new "youtube" option to the Source node that downloads a YouTube video via yt-dlp into a per-video-id cache and plays it on loop by delegating frame decoding to VideoFileInputSource. The probe endpoint differentiates 400 (invalid URL) / 404 (unavailable / private / live) / 408 (timeout) so the UI can render accurate status.

### DIFF
--- a/src/scope/core/inputs/youtube.py
+++ b/src/scope/core/inputs/youtube.py
@@ -1,0 +1,380 @@
+"""YouTube input source.
+
+Downloads a YouTube video to a local cache via yt-dlp, then reuses
+VideoFileInputSource to decode and loop frames with PyAV pacing.
+"""
+
+from __future__ import annotations
+
+import logging
+import re
+import threading
+import time
+from pathlib import Path
+from typing import ClassVar
+
+import numpy as np
+
+from .interface import (
+    InputSource,
+    InputSourceInfo,
+    InvalidSourceURLError,
+    SourceUnavailableError,
+)
+from .video_file import VideoFileInputSource
+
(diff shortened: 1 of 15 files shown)


leszko:
I took this url from youtube: https://www.youtube.com/watch?v=SGxz4LQfRpo

And it got stuck on this "checking...".
image

src/scope/core/inputs/youtube.py
```diff
@@ -0,0 +1,380 @@
...
+        if info is None:
+            return None
+        if info.get("is_live"):
+            return True
+        if info.get("live_status") in ("is_live", "is_upcoming"):
+            return True
+        return False
+
+    @staticmethod
+    def _download(url: str, cache_path: Path) -> bool:
```
leszko, line 312: Do we download the whole video before processing? I think that may be a problem with long videos. Plus, do we clean up the videos afterwards?

leszko (changes requested):
Added some comments. Other than that, not sure how urgently we need it, but IMO this would be better as a plugin / custom node. Maybe we can try next week how it fits into the node abstraction and trying to make a plugin out of that, wdyt @hthillman @gioelecerati ?

My reasoning is mainly that I can imagine a lot of other sources for video and if we start this in the composable/extendable manner, then the user can benefit, because they will be able to create their own Source Nodes.

hthillman:
Closing per @leszko's comment above about waiting for node abstraction, I agree!

PR #957, opened 2026-04-16
Add GPU selection to cloud connect UI

hthillman:
Replaces the on/off Remote Inference toggle in Settings → Daydream Account with a split button + dropdown that lets users pick which Livepeer GPU their cloud session targets. Options are H100 (2.5 credits/min), RTX 4090 and RTX 5090 (1.25 credits/min each). The last-used GPU is persisted in localStorage so the button is pre-populated on return, and while connecting a cancel (X) button replaces the caret. When connected, the dropdown surfaces "Switch GPU" and "Disconnect"; Switch GPU auto-reopens the picker on disconnect.

Onboarding is unchanged — CloudAuthStep and CloudConnectingStep still call connectToCloud() with no arg, so first-time users always land on H100 regardless of prior selection.

Backend threads the new gpu field (Literal["h100","rtx4090","rtx5090"]) through CloudConnectRequest → cloud_manager.connect_background → LivepeerConnection → LivepeerClient → _resolve_livepeer_app_id. Explicit app_id still wins over gpu, which wins over the SCOPE_CLOUD_GPU env var, which falls back to H100 — preserving CI, local cloud dev, and the existing env-var workflow.

### DIFF
--- a/frontend/src/components/settings/CloudGpuSelector.tsx
+++ b/frontend/src/components/settings/CloudGpuSelector.tsx
@@ -1,0 +1,241 @@
+/**
+ * CloudGpuSelector — split button + dropdown for picking a cloud GPU.
+ *
+ * Replaces the on/off Switch in Settings → Daydream Account. The main button
+ * reflects the currently-selected (or last-used) GPU, and the caret opens a
+ * dropdown with the three GPU options + per-minute credit cost. Selecting an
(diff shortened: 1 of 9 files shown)


BuffMcBigHuge:
Ran a quick test, seems to be functional. Great work!

PR #959, opened 2026-04-17
Add flip option to Syphon inputs.

j0sh:
Sometimes Syphon inputs appear upside-down. This is most likely because of coordinate-system mismatch between GPU textures and normal image buffers.

What’s going on:

- Syphon shares GPU textures.
- GPU texture coordinates are often treated as bottom-left / Y-up.
- Our pipeline code and numpy image arrays expect top-left / Y-down.
- syphon-python’s copy_mtl_texture_to_image() appears to just copy raw texture bytes into an array without normalizing orientation.
- So if the sender publishes a texture in the “GPU-native” orientation, we read it back as a normal image and it appears vertically flipped.

Also left-align the input elements of the Syphon box.

Note this can only be toggled when a stream is inactive. Making the input flippable while active seemed to reach pretty deeply into the code and I couldn't really understand that, so avoided it.

Before / after:

image
image

### DIFF
--- a/frontend/src/components/InputAndControlsPanel.tsx
+++ b/frontend/src/components/InputAndControlsPanel.tsx
@@ -88,6 +88,8 @@ interface InputAndControlsPanelProps {
   // Currently selected Syphon source identifier
   selectedSyphonSource?: string;
   onSyphonSourceChange?: (identifier: string) => void;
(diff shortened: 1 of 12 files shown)


leszko (approved):
LGTM

PR #960, opened 2026-04-17
Fix Syphon reloads in workflow mode

j0sh:
After reloading a workflow with a Syphon source, starting the stream would fail with "Video input required but no local stream available".

This happened because handleStartStream() decided whether a browser video stream was required from the global perform-mode input state (settings.inputSource) instead of the restored workflow graph. In workflow mode that global state isn't the source of truth, so graphs containing only server-side sources were incorrectly treated as needing a local stream.

Fix: In graphMode/nonLinearGraph, derive the local-stream requirement from graph topology. Only require a browser stream when the graph contains browser-backed source nodes (video/camera). Graphs with entirely server-side sources (syphon, ndi, spout) now start normally after reload.

Perform mode behavior is unchanged.

### DIFF
--- a/frontend/src/pages/StreamPage.tsx
+++ b/frontend/src/pages/StreamPage.tsx
@@ -2713,8 +2713,17 @@ export function StreamPage() {
         return false;
       }
 
-      // Check video requirements based on input mode
+      // Check video requirements based on input mode.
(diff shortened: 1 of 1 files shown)


leszko (approved):
Committed a small refactor. Other than that, LGTM. Merging.

PR #956, opened 2026-04-16
feat: add Kafka event publishing to Livepeer fal runner

emranemran:
## Summary

Publishes session lifecycle events from the livepeer fal wrapper (`src/scope/cloud/livepeer_fal_app.py`) to Kafka so cloud sessions show up in ClickHouse alongside the existing `fal_app.py` events.

## Events

- `websocket_connected` — on orchestrator → fal WS accept, includes GPU type, fal region/runner/log labels.
- `websocket_disconnected` — on teardown, adds `duration_ms` and session start/end timestamps.

Both events carry `user_id` (daydream user) and `connection_id` (Livepeer manifest ID).

## How IDs are sourced

`manifest_id` and `user_id` are read from WebSocket upgrade headers (`Manifest-Id`, `Daydream-User-Id`) set by the orchestrator in [livepeer/go-livepeer#3884](https://github.com/livepeer/go-livepeer/pull/3884) (already deployed to staging + prod). No message-body sniffing.

## Dependencies

- Runtime: `aiokafka` (added to both the fal wrapper `requirements` and the runner subprocess via `--extra kafka`).
- Config (fal secrets): `KAFKA_BOOTSTRAP_SERVERS`, `KAFKA_TOPIC`, `KAFKA_SASL_USERNAME`, `KAFKA_SASL_PASSWORD`. Kafka is disabled (events silently dropped) when unset.
- Orchestrator-side header change from livepeer/go-livepeer#3884.

## Not in scope

`pipeline_loaded`, `session_created`, `stream_started`, and `stream_heartbeat` already exist in the codebase (fired from `pipeline_manager.py`, `webrtc.py`, `frame_processor.py`) and will flow from the cloud runner once `KAFKA_*` env is set there. No changes in this PR.

## Commits

1. `feat: add Kafka event publishing to livepeer fal wrapper` (emran) — KafkaPublisher, env-var allowlist, aiokafka install, initial implementation.
2. `Extract manifest_id / user_id from websocket headers` (Josh Allmann) — replace proxy sniff with header reads.
3. `chore: strip smoke-test and [KAFKA-DEBUG] prints` (emran) — remove bootstrap scaffolding before merge.

## Verification

Tested end-to-end against the deployed fal app (`scope-livepeer-emran`) with `./test-cloud-connect.sh` (see #962 for the test harness). Both events confirmed landing in the production ClickHouse `scope_cloud_events` table with correct `manifest_id` and `user_id`.

[...]

PR #969, opened 2026-04-20
feat: bring livepeer runner Kafka events to parity with cloud-relay

emranemran:
## Summary

Follow-up to #956. That PR added `websocket_connected` / `websocket_disconnected` from the livepeer fal wrapper, but the rest of the session lifecycle (`pipeline_loaded`, `session_created`, `stream_started`, `stream_heartbeat`, `stream_stopped`, `playback_ready`, `error` variants) was either not firing or firing with null identifiers in livepeer mode. Cloud-relay mode (`fal_app.py` path) already had these events working; this PR brings livepeer mode to the same shape.

## Root causes (all in the runner, \`src/scope/cloud/livepeer_app.py\`)

1. \`FrameProcessor(...)\` was constructed without \`user_id\` / \`connection_id\` / \`session_id\` / \`connection_info\`, so every event it emits (\`stream_started\`, \`stream_heartbeat\`, \`stream_stopped\`, etc.) carried nulls.
2. \`session.manifest_id\` was never populated from \`job_info.manifest_id\`, so there was no identifier available that matched the wrapper's \`websocket_connected.connection_id\`.
3. The pipeline/load body injection used the runner's random internal UUID as \`connection_id\` instead of the manifest_id, so \`pipeline_loaded\` didn't correlate.
4. \`session_created\` / \`session_closed\` are only emitted from \`webrtc.py\`, which isn't on the livepeer code path — they have to be emitted explicitly.
5. \`connection_info\` env vars (\`NOMAD_DC\`, \`FAL_JOB_ID\`, \`FAL_LOG_LABELS\`, etc.) weren't in the runner subprocess allowlist, so the runner couldn't rebuild the dict even if asked.

## Changes

**\`src/scope/cloud/livepeer_app.py\`**
- Add \`manifest_id\` / \`session_id\` / \`connection_info\` fields to \`LivepeerSession\`; populate them right after parsing \`ScopeJobInfo\`.
- New \`_build_connection_info()\` helper that mirrors the shape built by \`livepeer_fal_app.py\` from env vars.
- Pass \`session_id\`, \`user_id\`, \`connection_id=manifest_id\`, \`connection_info\` into \`FrameProcessor\` so every downstream event from \`frame_processor.py\` / \`pipeline_processor.py\` carries identifiers that join with \`websocket_connected\`.
- Emit \`session_created\` right after \`FrameProcessor.start()\`, and \`session_closed\` right after \`FrameProcessor.stop()\`, using the same shape as \`webrtc.py:731\` and \`:1150\`.
- Flip the pipeline/load body injection from \`session.connection_id\` to \`session.manifest_id\`, and pass \`connection_info\` too.

**\`src/scope/cloud/livepeer_fal_app.py\`**
- Add \`NOMAD_DC\`, \`FAL_JOB_ID\`, \`FAL_RUNNER_ID\`, \`FAL_LOG_LABELS\`, \`FAL_MACHINE_TYPE\` to \`env_allowlist\` so the runner subprocess can reconstruct \`connection_info\`.

## Test plan

- [ ] \`uv run ruff check src/\` and \`ruff format --check src/\` pass (already verified)
- [ ] \`uv run daydream-scope\` starts without import/init errors
- [ ] \`./test-cloud-connect.sh --skip-push\` after CI build-cloud succeeds → exit 0 with CONNECTED
- [ ] Manual UI test: connect to cloud, load pipeline (\`longlive\`), start stream ~30s, disconnect
- [ ] ClickHouse query on \`scope_cloud_events\` filtered by \`user_id\` and \`connection_id = <manifest_id>\` shows all of: \`websocket_connected\`, \`pipeline_load_start\`, \`pipeline_loaded\`, \`session_created\`, \`stream_started\`, ≥2 \`stream_heartbeat\`, \`stream_stopped\`, \`session_closed\`, \`websocket_disconnected\` — all sharing the same connection_id
- [ ] Regression check on cloud-relay path (\`fal_app.py\` deploy): no changes, events unchanged

## Not in scope

- Local-scope-side events (\`session_created\` from WebRTC path on user's laptop, \`session_closed\`, webrtc connection errors) still require the user's local env to have \`KAFKA_*\` configured; that's a config concern, not code.
- \`/api/v1/session/start\` not being livepeer-compatible (\`mcp_router.py:252\` TODO) — untouched; only affects headless test flows.

🤖 Co-authored with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/scope/cloud/livepeer_app.py
+++ b/src/scope/cloud/livepeer_app.py
@@ -44,6 +44,7 @@ import scope.server.app as scope_app_module
 from scope.server.app import app as scope_app
 from scope.server.app import lifespan as scope_lifespan
 from scope.server.frame_processor import FrameProcessor
+from scope.server.kafka_publisher import publish_event
 from scope.server.media_packets import ensure_video_packet
 
 logger = logging.getLogger(__name__)
@@ -113,6 +114,29 @@ class LivepeerSession:
     media_publishes: list[MediaPublish | None] = field(default_factory=list)
     user_id: str | None = None
     connection_id: str | None = None
+    manifest_id: str | None = None
+    session_id: str | None = None
+    connection_info: dict[str, Any] | None = None
+
+
+def _build_connection_info() -> dict[str, Any]:
+    """Mirror the fal-wrapper's connection_info shape from env vars.
+
+    Keeps the runner's events (stream_*, pipeline_*) carrying the same
+    fal-region / runner-id / log-labels the wrapper's websocket_connected
+    uses, so downstream consumers can correlate them.
+    """
+    fal_log_labels_raw = os.getenv("FAL_LOG_LABELS", "unknown")
+    try:
+        fal_log_labels = json.loads(fal_log_labels_raw)
+    except (json.JSONDecodeError, TypeError):
+        fal_log_labels = fal_log_labels_raw
+    return {
+        "gpu_type": os.getenv("FAL_MACHINE_TYPE", "GPU-H100"),
+        "fal_region": os.getenv("NOMAD_DC", "unknown"),
+        "fal_runner_id": os.getenv("FAL_JOB_ID", os.getenv("FAL_RUNNER_ID", "unknown")),
+        "fal_log_labels": fal_log_labels,
+    }
 
 
 async def _shutdown_task(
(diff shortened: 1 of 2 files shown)


src/scope/cloud/livepeer_app.py
```diff
@@ -113,6 +114,29 @@ class LivepeerSession:
     media_publishes: list[MediaPublish | None] = field(default_factory=list)
     user_id: str | None = None
     connection_id: str | None = None
+    manifest_id: str | None = None
```
j0sh, line 117: Pretty sure manifest_id is the same as the connection_id above

j0sh: ... oh dear it's not. Well that's a simple change then

emranemran: i made manifest id to be same as connection id and the old conenction id should be gone now.

j0sh (approved):

src/scope/cloud/livepeer_app.py
j0sh: I guess the manifest_id does not need to be passed in anymore if that's the case? Would simplify some of the ad hoc checks around here.

emranemran: Yeah that's true. I'll clean this up in a follow up and file a ticket so i don't forget.

PR #972, opened 2026-04-21
Remove Cloud Relay mode

j0sh:
Now that Livepeer is the default going forward, clean up the repo of the old Cloud Relay Mode so that agents and humans don't get confused trying to maintain or test both.

This also cleans up the long-unused frontend "cloud adpater" path that would initiate a cloud WebSocket connection from the frontend.


j0sh:
To merge after the 2.3 release

.github/workflows/build-electron-preview.yml
```diff
@@ -55,7 +55,6 @@ jobs:
         run: npm run dist:win
         env:
           GH_TOKEN: ${{ secrets.GH_TOKEN }}
-          SCOPE_CLOUD_APP_ID: daydream/scope-app/ws
```
leszko: I think this should be kept and changed to `daydream/scope-livepeer/ws`. The same below. The reasoning is that the preview electron app is built after every push to `main` and it should use the latest version deployed to the cloud (not the prod).

.github/workflows/docker-build.yml
```diff
@@ -70,34 +68,6 @@ jobs:
       - name: Install fal
         run: pip install fal
 
-      - name: Deploy to fal (PR environment)
```
leszko, line 73: Just to double-check, after every push to `main`, we're still deploying to `fal` and use it for cloud when `SCOPE_CLOUD_APP_ID="daydream/scope-livepeer/ws"` is specified. Is that correct?

src/scope/cloud/dev_app.py
```diff
@@ -1,302 +0,0 @@
-"""Development-only cloud relay for testing the cloud flow locally.
```
leszko, line 1: I still use this for local dev, so I'd prefer to keep it. Unless you have some other way to test cloud relay without having to use go-livepeer or the actual cloud?

leszko (commented):
Added a few questions. Mainly two things:
- We need to preview electron app to work with the preview cloud (the current `main`)
- I've seen you removed the local dev cloud. Is the one with go-livepeer as smooth as the one from dev_cloud.py?

.github/workflows/docker-build.yml
j0sh: That's correct

src/scope/cloud/dev_app.py
j0sh: Go-livepeer is needed for now. The cloud relay is different enough that it is not representative of the actual remote inference flow anymore, and this is part of the reason for removing it.

leszko (approved):
LGTM. Feel free to merge after you:
- resolve conflicts
- make sure we use the `scope-livepeer` environment for preview electron app

.github/workflows/build-electron-preview.yml
j0sh: Fixed in https://github.com/daydreamlive/scope/pull/972/changes/f269276d18773c44a865f4081f909d909e6440dd

PR #991, opened 2026-04-27
Improved source routing support

j0sh:
Update the graph executor to recognize these patterns:

* source -> sink
* source -> recording
* source -> sink -> recording

Previously recordings and sinks could not attach directly to sources even though this is an allowed behavior in the UI.

There's a deeper refactor needed here to support these bits more generically but this fixes the most immediate issues.

Example workflow:

[multi-recording.scope-workflow.json](https://github.com/user-attachments/files/27113516/multi-recording.scope-workflow.json)

image


leszko (approved):
LGTM

PR #980, opened 2026-04-23
feat: Unify Pipeline and Node

leszko:
## Summary

Pipeline used to be a distinct subclass of BaseNode with its own interface module, parallel registry, artifact/load paths, and workflow schema. All the meaningful distinctions collapse to one predicate — whether a node is *config-driven* (`get_config_class()` returns a `BasePipelineConfig`). Everything else follows from that, so this PR unifies the two ends of the stack and lets graph nodes plug into pipelines through the same machinery.

## Changes

**Core unification**
- `Pipeline = Node`, `BaseNode = Node`. `Node` gains `get_config_class()`, `get_artifacts()`, `get_definition()`, `prepare()`, `__call__`, and `Requirements`.
- `PipelineRegistry` reduces to a deprecation shim over `NodeRegistry`; `chain_produces_{video,audio}` move there.
- `Node.get_artifacts()` defaults to `get_config_class().artifacts`. `artifact_registry` and `download_models` resolve through `NodeRegistry`.
- Every graph node — pipelines and plain nodes alike — flows through `PipelineManager`, so lifecycle, dedup, and multi-instance keying are uniform. `PipelineManager` drops its `PipelineRegistry` import entirely.

**Workflow schema**
- `WorkflowPipeline` merges into `WorkflowNode` (`node_type_id` + `source` + `loras`). A `model_validator` folds legacy `pipelines` JSON input into the canonical `nodes` list, and `node_type_id` accepts `pipeline_id` via `AliasChoices` so existing workflow files load unchanged.
- Resolve iterates one list. `ResolutionItem.kind` gains `"node"`.

**Plugin attribution**
- One `_type_to_plugin` dict and a `get_plugin_for_type_id` accessor, populated by a single `_update_plugin_mapping` walk that handles both `register_pipelines` and `register_nodes` hooks.
- `NodeDefinition.plugin_name` lets workflow exports attribute plain nodes to their plugin.

**`__prompt` routing (built on the unified abstraction)**
- `BasePipelineConfig.input_port_types` ClassVar declares non-video input ports.
- `Node.get_definition` synthesizes a `__prompt` string port for every prompt-capable pipeline so graph edges can target it.
- `PipelineProcessor` drains non-video input queues into `self.parameters` each chunk; `__prompt` values are remapped to `prompts` (raw strings normalised to `[{text, weight}]`).
- Parameter updates from graph edges broadcast `parameters_updated` over the data channel via `NotificationSender`, with a retry loop covering the WebRTC handshake race.

[...]

PR #992, opened 2026-04-27
Keep record-node queues dedicated to recorder

j0sh:
Prevent WebRTC from 'stealing' frames from the recorder. This led to some recordings being low-FPS or empty.

What the recordings looked like
-------------------------------
Sequence with both consumers attached (current trunk before this fix):
    source ──▶ rec_q ──┬──▶ RecordOutputTrack (WebRTC, always-on)
                       │
                       └──▶ QueueVideoTrack  (RecordingManager,
                                              start/stop driven)
    t0  source pushes frames into rec_q
    t1  RecordOutputTrack.recv() drains rec_q  (every ~1/30s)
    t2  user clicks Start: RecordingCoordinator drains stale frames,
        creates QueueVideoTrack on rec_q
    t3  source pushes 1 frame, RecordOutputTrack drains it
    t4  source pushes 1 frame, RecordOutputTrack drains it
    t5  ... QueueVideoTrack only intermittently wins ...
    t6  user clicks Stop: MediaRecorder finalizes with too few
        encoded frames; some segments end up 0/invalid MP4
    t7  user clicks Start again: rec_q re-drained, but the same race
        starts over for the next segment

After this change there is only one consumer per record queue:
    source ──▶ rec_q ──▶ QueueVideoTrack (RecordingManager only)
Fix
---
- Remove RecordOutputTrack attachment from WebRTCManager.handle_offer.
- Factor extra-output construction into `_build_extra_output_tracks`, which only emits SinkOutputTrack instances for non-primary sink nodes. Record nodes are intentionally skipped and logged.
- Record nodes remain available through the existing recording HTTP endpoints (`/api/v1/recordings/{session_id}/{start,stop}` and the download route); only the live WebRTC track was wrong.

Tests
-----
- New `test_record_nodes_are_not_attached_as_live_webrtc_outputs` in `tests/test_recording_coordinator.py` (placed there because the invariant being protected is "record queues belong to recording"; WebRTC is just where the violation lived).
- Existing `test_start_recording_drops_stale_record_queue_frames` still covers the stale-frame drain on Start. source -> sink -> record routing remain green.

### DIFF
--- a/src/scope/server/webrtc.py
+++ b/src/scope/server/webrtc.py
@@ -33,7 +33,6 @@ from .pipeline_manager import PipelineManager
 from .recording import RecordingManager
 from .schema import WebRTCOfferRequest
 from .tracks import (
-    RecordOutputTrack,
     SinkOutputTrack,
     SourceInputHandler,
     VideoProcessingTrack,
@@ -58,6 +57,28 @@ vpx.MAX_FRAME_RATE = 8
 vpx.DEFAULT_BITRATE = 7000000
 vpx.MIN_BITRATE = 5000000
 vpx.MAX_BITRATE = 10000000
+
+
+def _build_extra_output_tracks(
+    frame_processor: FrameProcessor,
+    sink_node_ids: list[str],
+    record_node_ids: list[str],
+) -> list["NodeOutputTrack"]:
+    """Build live WebRTC output tracks for graph outputs.
+
+    Record-node queues are single-consumer recording queues. Attaching them as
+    live WebRTC tracks races with RecordingCoordinator and can drain/corrupt
+    recordings, so only sink nodes get live output tracks here.
+    """
+    if record_node_ids:
+        logger.debug(
+            "Skipping live WebRTC tracks for record nodes: %s",
+            record_node_ids,
+        )
+    return [
+        SinkOutputTrack(frame_processor=frame_processor, sink_node_id=sink_id)
+        for sink_id in sink_node_ids[1:]
+    ]
 
 
 def _parse_graph_node_ids(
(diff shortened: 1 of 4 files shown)


src/scope/server/webrtc.py
```diff
@@ -682,23 +703,17 @@ def on_data_channel_message(message):
...
             if video_track is not None and relay is not None:
-                for sink_id in sink_node_ids[1:]:
-                    extra_output_tracks.append(
-                        SinkOutputTrack(
-                            frame_processor=frame_processor, sink_node_id=sink_id
-                        )
-                    )
-                for rec_id in record_node_ids:
-                    extra_output_tracks.append(
-                        RecordOutputTrack(
```
leszko, line 702: Do we use now RecordOutputTrack anywhere? I not, then maybe we can remove it?

leszko (approved):
Added one comment. Other than that LGTM

src/scope/server/webrtc.py
j0sh: Nope it's not used ... removed in [ac98b1b](https://github.com/daydreamlive/scope/pull/992/commits/ac98b1b2f66564efac0078c3e0abbd940f1dbe6f), thanks for the spot

j0sh:
Also updated to main and pushed a small fix for tests that broke from two conflicting merges:

[87455db](https://github.com/daydreamlive/scope/pull/992/commits/87455db5157f0f942bd133ecf087956933f61ff1)
The merge of https://github.com/daydreamlive/scope/pull/991 led
to broken tests when https://github.com/daydreamlive/scope/pull/980
was merged a few hours prior.

Move fast and break things, etc.

PR #996, opened 2026-04-27
Expose any node param via OSC + per-param opt-in + defaults in docs

hthillman:
Closes the three OSC pain points reported by a user building a TouchDesigner controller rig:

1. **Source / Sink / Slider / Knobs / etc. now reachable via OSC** — previously only pipeline params from the registry were exposed.
2. **Per-param opt-in** — right-click any node → "Configure OSC…" picks which params external clients can hit.
3. **Defaults in the OSC description** — `/api/v1/osc/paths` and `/api/v1/osc/docs` now include a `default` for each path.

## How it works

- The frontend computes an OSC inventory from the live graph plus each node's new `oscConfig` overlay (`{paramName: {exposed, address?, default?}}`) and POSTs it to a new `POST /api/v1/osc/inventory` endpoint.
- The OSC server merges the graph inventory with the registry-derived paths for validation, broadcast, and discovery.
- Pipeline-targeted OSC messages broadcast to WebRTC with `{node_id, param: value}` so `frame_processor.update_parameters` routes to the right pipeline processor.
- UI-only node messages skip WebRTC and reach the frontend through the existing SSE `osc_command` event; `StreamPage`'s handler now forwards `node_id`+`param` to `graphEditorRef.applyExternalParams(...)`, so any field on `node.data` can be addressed without new backend processors.

## Address format

- Default address: `/scope/<slugified-customTitle>/<param>` (per-node namespacing avoids collisions when you have two pipelines or two sliders).
- Each row in the Configure OSC modal has an editable address — paste any path you want.
- Pipeline runtime params keep their flat `/scope/<param>` back-compat default; the legacy address is only suppressed when the user has explicitly re-bound that param via Configure OSC. Existing TouchDesigner setups don't break.

## Backend changes

- `src/scope/server/osc_docs.py` — `_extract_osc_paths_from_schema` now includes `default`. `get_osc_paths` / `get_all_known_paths` accept a `graph_inventory` arg and merge it; HTML docs render a Default column.
- `src/scope/server/osc_server.py` — new `set_graph_inventory` setter; `_handle_osc_message` routes pipeline params via `{node_id, param: value}` to WebRTC, UI-only params via SSE only.
- `src/scope/server/app.py` — new `POST /api/v1/osc/inventory` endpoint with Pydantic models.

## Frontend changes

- `frontend/src/lib/oscNodeParams.ts` — NEW: per-node-type param descriptors (Source, Output, Slider, XYPad, Bool, Trigger, Tempo, Primitive, Note).
- `frontend/src/components/graph/OscConfigDialog.tsx` — NEW: per-node Configure OSC modal.
- `frontend/src/components/graph/hooks/graph/useOscInventory.ts` — NEW: derives inventory + debounced POST.
- `frontend/src/components/graph/contextMenuItems.tsx` — adds "Configure OSC…" item.
- `frontend/src/components/graph/GraphEditor.tsx` — wires the modal + the inventory hook.
- `frontend/src/lib/graphUtils.ts` — adds `oscConfig` to `FlowNodeData` (round-trips through `ui_state.nodes[*].data`).
- `frontend/src/lib/api.ts` — adds `OscInventoryEntry`, `OscPathsResponse`, `setOscInventory`, `fetchOscPaths`.
- `frontend/src/pages/StreamPage.tsx` — OSC handler routes `node_id`+`param` to `applyExternalParams`.
- `frontend/src/components/settings/OscTab.tsx` — adds a live "Currently exposed paths" panel with click-to-copy.

---

## How to test

Spin up Scope (`uv run daydream-scope`) — OSC listens on UDP `8000` (same port as the HTTP server). All `python-osc` snippets below assume `pip install python-osc` is available; or use `uv run python -c '…'`.

### 1. Quick visual sanity (30s)

- **Settings → OSC tab**: new **"Currently exposed paths"** panel below the existing controls. Empty initially with a hint pointing at "Configure OSC…".
- **Right-click any node** on the canvas: new **"Configure OSC…"** menu item appears between Pin and Delete.

### 2. UI-node opt-in (the headline new case — Slider)

1. Add a **Slider** node (`+` upper right → UI → Slider, or right-click canvas → UI → Slider).
2. Click the title and rename it to e.g. `Tempo` so the auto-address is meaningful.
3. Right-click the slider → **Configure OSC…**
4. In the modal:
   - Tick **Expose** on the `value` row
   - Address auto-fills as `/scope/tempo/value`
   - Default field placeholder shows the current slider value (e.g. `0.5`); leave it or type `1.0`
   - **Save**
5. Settings → OSC → click **Refresh** → confirm `/scope/tempo/value` appears under the "Tempo" group with the default.
6. Click **"Open OSC Docs"** → confirm the path is listed with the **Default** column populated.
7. Drive it from the outside:
   ```bash
   uv run python -c "from pythonosc.udp_client import SimpleUDPClient as C; C('127.0.0.1', 8000).send_message('/scope/tempo/value', 0.85)"
   ```
   Watch the slider snap to 0.85 in real time.

### 3. Source / Output (the originally-blocked case)

1. Add a **Source** node, right-click → Configure OSC.
2. Expose **`sourceMode`** (it's an enum with values `video` / `camera` / `spout` / `ndi` / `syphon`). Save.
3. Send `'/scope/source/sourceMode', 'ndi'` → dropdown switches in the UI.
4. Repeat on an **Output** node with **`outputSinkEnabled`** (bool) — toggle from off → on via OSC.

### 4. Pipeline node namespacing (multi-pipeline collision)

1. Load a workflow with two pipeline nodes that share a runtime param (e.g. both expose `prompt`).
2. Title them `Main` and `Secondary`.
3. Configure OSC on each → expose `prompt` → save.
4. Send to `/scope/main/prompt` vs `/scope/secondary/prompt` independently → only the matching node updates.

### 5. Backwards compatibility

Without configuring anything new, legacy flat addresses still work:
```bash
uv run python -c "from pythonosc.udp_client import SimpleUDPClient as C; C('127.0.0.1', 8000).send_message('/scope/noise_scale', 0.6)"
```
The pipeline's `noise_scale` updates as before. Pre-existing TouchDesigner setups are unaffected.

### 6. Defaults in description

```bash
curl -s http://localhost:8000/api/v1/osc/paths | jq '.active.Runtime[] | {key, default}'
```
Every entry that has a schema default should show it. Open `http://localhost:8000/api/v1/osc/docs` in a browser → the new **Default** column is populated.

### 7. Persistence

Configure OSC on a node, save, refresh the page. The graph reloads with `oscConfig` intact and the inventory is re-pushed to the backend within ~300ms.

### 8. Lint + build

```bash
uv run ruff check src/ && uv run ruff format --check src/
cd frontend && npm run lint && npm run format:check && npm run build
```

## Known limitations (not in this PR)

- **Address slug collisions** aren't surfaced in the UI yet — two same-titled sliders both compute to `/scope/<title>/value`; the latest registered wins. Workaround: rename one or override the address. A warning indicator is planned for a follow-up.
- **Composite-shape params** (knobs[], midiChannels[], tupleValues[]) are skipped for the MVP — they need a richer addressing scheme (per-index sub-paths). Will follow.
- The user-set "default" is **advisory metadata** — it's published in the OSC docs so external clients can match Scope's starting state, but Scope does not auto-emit it on session start. Auto-apply behavior is a deliberate follow-up to keep this PR focused on description vs. behavior.

### DIFF
--- a/docs/osc.md
+++ b/docs/osc.md
@@ -1,0 +1,279 @@
+# Using OSC
+
+Scope exposes its parameters over [OSC (Open Sound Control)](https://opensoundcontrol.stanford.edu/) so external tools — TouchDesigner, Resolume, MaxMSP, hardware controllers, custom Python scripts — can drive the running graph in real time. The OSC server is always on; it shares a UDP socket with the HTTP API on the same port (default `8000`).
+
+## Table of Contents
+
+- [Quick Start](#quick-start)
+- [What's Reachable via OSC](#whats-reachable-via-osc)
+- [Configure OSC per Node](#configure-osc-per-node)
+- [Address Format](#address-format)
+- [Discovering Paths](#discovering-paths)
+- [Defaults in the Description](#defaults-in-the-description)
+- [Validation](#validation)
+- [Routing Internals](#routing-internals)
+- [TouchDesigner Setup](#touchdesigner-setup)
+- [Python Examples](#python-examples)
+- [REST API](#rest-api)
+- [Limitations](#limitations)
+
+---
+
+## Quick Start
+
+1. Start Scope (`uv run daydream-scope`). The OSC server starts automatically and listens on UDP `8000`.
+2. Send a message:
+
+   ```python
+   from pythonosc.udp_client import SimpleUDPClient
+   client = SimpleUDPClient("127.0.0.1", 8000)
+   client.send_message("/scope/prompt", "a beautiful sunset over the ocean")
+   client.send_message("/scope/noise_scale", 0.5)
+   ```
+
+3. Open the live OSC reference at `http://localhost:8000/api/v1/osc/docs` to see every address, type, range, default, and a copy-paste example.
+
+> [!NOTE]
+> Set the OSC port via the `SCOPE_PORT` environment variable. UDP and HTTP coexist on the same port, so changing one changes both.
+
+---
+
+## What's Reachable via OSC
+
+Scope exposes three layers of OSC paths:
+
+| Layer | Source | Default address | Notes |
+|---|---|---|---|
+| **Runtime globals** | Built-in (`prompt`, `noise_scale`, `paused`, `manage_cache`, `reset_cache`, `transition_steps`, `interpolation_method`, …) | `/scope/<param>` | Always on. Backwards-compatible with pre-existing TouchDesigner setups. |
+| **Pipeline runtime params** | Each loaded pipeline's `is_load_param=False` fields | `/scope/<param>` (default) or `/scope/<node>/<param>` (per-instance, opt-in) | One bare-address entry per param while no per-node override is set; switches to namespaced when you customize the node. |
+| **Graph nodes** (Source / Sink / Slider / XYPad / Bool / Trigger / Tempo / Output / Note / Primitive) | The graph editor's per-node `oscConfig` | `/scope/<node>/<param>` | Off by default; opted in via right-click → **Configure OSC…** |
+
+The Source / Sink / UI-node layer is the new piece — previously only pipeline params could be reached.
+
+---
+
+## Configure OSC per Node
+
+The graph editor's right-click menu has a **"Configure OSC…"** item:
+
+1. Right-click any node → **Configure OSC…**
+2. The modal lists every OSC-eligible param for that node type. Each row has:
+
+   | Column | Behavior |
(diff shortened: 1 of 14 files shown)


leszko (approved):
Pushed a few small fixes. Other than that looks good.

PR #1004, opened 2026-04-28
Forward runner notifications back to local Scope in cloud mode

leszko:
## Summary
- Closes a UI desync in cloud mode: when a backend node (e.g. `prompt-enhancer`) feeds a new value into a pipeline scalar input port, the local browser was not updating the corresponding widget (e.g. longlive `Prompt` textarea) — generation kept using the new prompt, but the visible state was frozen on the initial value.
- The runner already calls `broadcast_notification({"type": "parameters_updated", ...})` from `pipeline_processor._drain_non_video_inputs`. In livepeer/trickle mode the runner has no WebRTC sessions, so those notifications were silently dropped. This PR plumbs them back to the local Scope and on to the browser.

## What changed
- `src/scope/server/webrtc.py`: `WebRTCManager` grows `add_notification_listener` / `remove_notification_listener`. `broadcast_notification` fires registered listeners after the session loop; existing in-process subscribers are unaffected.
- `src/scope/cloud/livepeer_app.py`: `start_stream` registers a listener that ships each broadcast to the local Scope via `session.ws.send_json({"type": "notification", "payload": ...})` (using `run_coroutine_threadsafe` since the listener fires from a worker thread). `_stop_stream` unregisters it. New `notification_listener` field on `LivepeerSession` for cleanup.
- `src/scope/server/livepeer_client.py`: `_events_loop` handles a new `notification` message type by re-broadcasting the payload through the local `webrtc_manager.broadcast_notification`, which already knows how to reach the browser data channel.

## Comparison with OSC
OSC works in cloud mode because `osc_server` pushes a parallel SSE event for the frontend (`{"type": "osc_command", ...}`) — that runs entirely on the local side and does not depend on cloud→local plumbing. Backend-emitted notifications (`parameters_updated` from a runner-side node) had no equivalent path; this PR adds it.

## Test plan
- [ ] Workflow from #1002: `longlive` + `scheduler` + `prompt-enhancer` cycling on `boy`/`superman`. In cloud mode, confirm the longlive `Prompt` widget refreshes to the enhanced text each time the scheduler fires.
- [ ] Same workflow in local mode: behavior is unchanged.
- [ ] Local-only sessions (no cloud) — confirm no regression in `parameters_updated` delivery (listener list is empty by default).
- [ ] Stop and restart a cloud stream — confirm the listener is removed in `_stop_stream` and re-registered on the next `start_stream` (no duplicate forwards).

### DIFF
--- a/src/scope/cloud/livepeer_app.py
+++ b/src/scope/cloud/livepeer_app.py
@@ -19,6 +19,7 @@ import queue
 import shutil
 import threading
 import uuid
+from collections.abc import Callable
 from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -124,6 +125,13 @@ class LivepeerSession:
     manifest_id: str | None = None
     session_id: str | None = None
     connection_info: dict[str, Any] | None = None
+    # Serializes every write on ``ws`` so concurrent senders (control-message
+    # responders, the notification pump, etc.) cannot interleave at the ASGI
+    # send boundary and corrupt the websocket stream.
+    ws_send_lock: asyncio.Lock = field(default_factory=asyncio.Lock)
+    notification_listener: Callable[[dict], None] | None = None
+    notification_queue: asyncio.Queue[str | None] | None = None
+    notification_pump_task: asyncio.Task | None = None
 
 
 def _build_connection_info() -> dict[str, Any]:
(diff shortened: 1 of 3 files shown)


src/scope/cloud/livepeer_app.py
```diff
@@ -1056,6 +1124,86 @@ async def _handle_control_message(
             connection_info=session.connection_info,
         )
         session.frame_processor.start()
+
+        # Forward runner-side notifications (e.g. parameters_updated emitted
+        # when an upstream node feeds a new value into a pipeline's scalar
+        # input port) to the local Scope over the websocket. The local
```
j0sh, line 1130: Was this tested?

The WebSocket only talks to the orchestrator for some light signaling / keepalives. Messages to the local Scope go through the trickle event channel which already has its own atomic write loop.

src/scope/cloud/livepeer_app.py
```diff
@@ -124,6 +125,13 @@ class LivepeerSession:
     manifest_id: str | None = None
     session_id: str | None = None
     connection_info: dict[str, Any] | None = None
+    # Serializes every write on ``ws`` so concurrent senders (control-message
+    # responders, the notification pump, etc.) cannot interleave at the ASGI
+    # send boundary and corrupt the websocket stream.
+    ws_send_lock: asyncio.Lock = field(default_factory=asyncio.Lock)
```
j0sh, line 131: Was there a real problem in practice here? Had Claude check this and this is what it came back with:

---

The justification in the source comment is:

> Starlette does not lock around its lower-level send, which corrupts the protocol stream and tears the connection.

That's overstated. WebSocket.send_json in Starlette boils down to a single await self.send({"type": "websocket.send", "text": ...}) — one ASGI message, which uvicorn turns into one WebSocket text frame, atomic at the protocol layer. Concurrent calls without a lock can be reordered, but cannot interleave bytes inside a single frame, so the protocol stream isn't actually at risk.

j0sh (changes requested):
Started looking at this and the overall shape seemed a little strange, including the use of WebRTC in the PipelineProcessor which was recently added and doesn't really make sense in cloud mode.

Took another shot at fixing this in https://github.com/daydreamlive/scope/pull/1008 but didn't test so feel free to take it from there.

leszko:
> Started looking at this and the overall shape seemed a little strange, including the use of WebRTC in the PipelineProcessor which was recently added and doesn't really make sense in cloud mode.
> 
> Took another shot at fixing this in #1008 but didn't test so feel free to take it from there.

Yeah, let me close this one. I'll commit one fix to your PR and merge yours. Thanks!

PR #1006, opened 2026-04-28
feat: Search installed pipelines and plugin nodes from pane context menu

gioelecerati:
The right-click pane menu's search box only matched the static category labels and their hard-coded keywords, so a pipeline like scope-gesture-pipeline was undiscoverable by name. Add searchOnly ghost entries built from PipelinesContext and /api/v1/nodes/definitions, with the plugin package name included in keywords. Clicking a pipeline result creates a Pipeline node already preselected to that pipeline by routing through the existing handlePipelineSelect path. The static menu is unchanged when the search box is empty.

### DIFF
--- a/frontend/src/components/graph/AddNodeModal.tsx
+++ b/frontend/src/components/graph/AddNodeModal.tsx
@@ -9,6 +9,7 @@ import {
 import type { FlowNodeData } from "../../lib/graphUtils";
 import type { NodeDefinitionDto } from "../../lib/api";
 import { fetchNodeDefinitions } from "../../lib/api";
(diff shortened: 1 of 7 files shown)


leszko:
I think it's mostly the same as https://github.com/daydreamlive/scope/pull/988 and what I share here: https://discord.com/channels/1368975848123666553/1497233106455887933/1497233151062315183

leszko:
Closing, since I believe it's the same as https://github.com/daydreamlive/scope/pull/988. But if there anything you'd like to see different on that front, feel free to reopen.

PR #1027, opened 2026-05-07
feat(workflows): bundle media files inside exported workflow JSON

leszko:
https://github.com/user-attachments/assets/dfd93ed2-c9d0-446f-94fe-2287da584b24

Workflows reference media (images, audio, video) by filesystem path, which only resolves on the author's machine. This change base64-encodes every referenced file into a top-level `embedded_assets` array on export, and on import writes them back to the local assets directory and rewrites the path references.

New endpoints `POST /api/v1/workflow/embed` and `POST /api/v1/workflow/extract` handle the round trip; both `@cloud_proxy()` to the cloud server in cloud mode (assets live on cloud disk). Extract is idempotent via SHA-256 and dedupes basename collisions. `format_version` bumps to `"1.1"` only when `embedded_assets` is present; older files stay compatible.

## Test plan
- [x] Unit tests cover round trip, idempotency, basename collisions, graph ui_state walking, free-form-prompt protection, Windows-style paths, and cloud-mode warning header
- [x] HTTP smoke test: embed → delete file → extract restores byte-identical, prompt unmangled, paths rewritten
- [x] Local Livepeer relay: embed/extract proxied to runner, frame after extract matched the seeded source

### DIFF
--- a/frontend/src/components/WorkflowExportDialog.tsx
+++ b/frontend/src/components/WorkflowExportDialog.tsx
@@ -62,13 +62,17 @@ export function WorkflowExportDialog({
         scopeVersion: scopeVersion ?? "unknown",
       });
 
-      // Ask the backend to embed referenced media files. If embedding fails
-      // (e.g. during testing without a backend), fall back to the un-embedded
-      // workflow so export still works.
+      // Ask the backend to embed referenced media files. If embedding fails,
+      // fall back to the un-embedded workflow and warn the user — without
+      // embeds, anyone who imports the file on another machine will see
+      // missing assets where the author had reference images, audio, or
+      // video files.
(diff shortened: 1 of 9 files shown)


j0sh (approved):
Seems fine but these are going to make for some huge json blobs that are hard to decipher and modify. Would a zip file be better?

leszko:
> Seems fine but these are going to make for some huge json blobs that are hard to decipher and modify. Would a zip file be better?

Yeah, in general for big assets zip would be better. But I lean towards merging it as it is, because I don't want to create more confusion about import/export until it's needed. If we start seeing people exporting huge assets, then we'll just support both zip and json.

PR #1028, opened 2026-05-07
fix(lora): fail fast on incompatible LoRAs and warn in UI on missing files

leszko:
https://github.com/user-attachments/assets/f6f2fc50-3414-4a1c-a332-7165e120e500

## Summary
- Validate LoRA `lora_A`/`lora_B` shapes against target weights in `parse_lora_weights`. On mismatch (e.g. a Wan 14B LoRA on a Wan 1.3B pipeline) raise a typed `LoRAIncompatibleError` naming the layer and both shapes, instead of crashing deep in the first forward pass.
- Surface the message via `pipeline_load_failed` so the UI shows it.
- In the LoRA picker, flag a configured-but-missing path with an amber warning + tooltip (cloud vs. local copy) before the user starts the pipeline.

## Test plan
- [ ] `uv run pytest tests/test_lora_dim_validation.py tests/test_pipeline_manager.py`
- [ ] Load an incompatible LoRA in the UI and confirm the typed error message is shown.
- [ ] Configure a LoRA path that doesn't exist locally / on the cloud runner and confirm the picker shows the warning.

### DIFF
--- a/frontend/src/components/LoRAManager.tsx
+++ b/frontend/src/components/LoRAManager.tsx
@@ -33,7 +33,8 @@ export function LoRAManager({
   loraMergeStrategy = "permanent_merge",
   onOpenLoRAsSettings,
 }: LoRAManagerProps) {
-  const { loraFiles: availableLoRAs } = useLoRAsContext();
+  const { loraFiles: availableLoRAs, isLoading: isLoadingLoRAs } =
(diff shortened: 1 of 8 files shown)


leszko:
Merging, feel free to review post-merge if needed.

PR #1031, opened 2026-05-12
fix(av): flush buffers on pipeline discontinuity

leszko:
When a pipeline marks a chunk as a timeline discontinuity, drop queued audio (via the existing flush sentinel) and queued output frames before enqueueing the new chunk. Fixes A/V desync after hold/resume.


BuffMcBigHuge (approved):
LGTM

PR #1033, opened 2026-05-13
feat(audio): [DEMON] route graph-node audio to WebRTC/MPEG-TS sink

leszko:
## Summary

Lets plain-node graphs (e.g. DEMON's ACEStep music-cover workflow) emit audio through the WebRTC / MPEG-TS sink.

## Demo

https://github.com/user-attachments/assets/c9f702bf-820c-4108-8fdf-0e6371bb4446

## Workflow
[demon.scope-workflow.json](https://github.com/user-attachments/files/28009875/demon.scope-workflow.json)

## Changes

- `NodeProcessor`: latch-with-rerun for non-continuous nodes, persisted one-shot inputs (MODEL/VAE/CLIP), and an audio-routing path that pushes to `audio_output_queue` with PTS.
- `graph_executor`: tag the feeder's `audio_sink_ports` for audio→sink edges instead of allocating an unused sink queue.
- `AudioProcessingTrack`: trim duplicate-prefix samples on overlapping decoder windows (e.g. `vae_overlap` + `follow_playhead`).
- `mcp_router` / `webrtc`: derive sink modalities from the graph via `GraphConfig.get_sink_modalities()`, OR'd with the registry so pipelines with internal audio (e.g. LTX-2) keep their audio track.
- `HeadlessTsStreamer`: bootstrap the MPEG-TS container from the first audio chunk (audio-only graphs).
- Built-in `audio.AudioSource` node: WAV file → one-shot full-clip emit.
- `AudioPacket` construction moved next to the type, in `media_packets`.


j0sh (approved):
Looks fine, I think the headless mpeg-ts endpoint needs timestamp passthrough as well, I have a patch here somewhere for that

PR #1040, opened 2026-06-01
feat(cloud): relay telemetry over the trickle events channel

emranemran:
## Why

In Livepeer/trickle cloud mode the GPU **runner** executes the full embedded Scope app, so every `publish_event()` call (session/stream/pipeline/error lifecycle) already fires on the runner — but the runner has no Kafka credentials, so those events **silently no-op and are lost**.

This PR redirects runner telemetry over the existing trickle **events** channel. The local SDK client receives it and becomes the single egress point that publishes to Kafka (or another backend).

> Note: the Kafka code was **not** deprecated/removed — it's live, gated behind `KAFKA_BOOTSTRAP_SERVERS`. This is a **redirect**, not a rebuild.

## How it works

```
RUNNER (cloud/livepeer_app.py)            CLIENT (server/livepeer_client.py)
  publish_event(...)                        _events_loop: type=="telemetry"
    -> build_event_envelope()                 -> publish_raw_event(event)
    -> TrickleEventsSink.emit()                  -> TelemetrySink.emit()
        -> bounded queue (drop-oldest)               = KafkaSink (default)
        -> drain task                                |- HttpSink (future)
        -> events_writer.write(                      |- LogSink (test)
            {"type":"telemetry","event":{...}}) --trickle events channel-->
```

`TelemetrySink` is one shared protocol used on both sides; only the implementation differs. `publish_event()` builds the envelope once and hands it to the active sink, **falling back to today's direct-Kafka behavior when no sink is set** — so local (non-cloud) Scope is unchanged. Events are forwarded **verbatim** (origin `id`/`timestamp`/`session_id`/`connection_id` preserved), best-effort and lossy under pressure so telemetry never stalls the media or control paths.

## High-level changes

- **`server/kafka_publisher.py`** — `TelemetrySink` protocol, `build_event_envelope()` (stamp once), `publish_prebuilt()` / `publish_raw_event()` (verbatim send), `KafkaSink`, `LogSink`, sink globals, and `install_default_egress_sink()` (the single swappable seam for the client backend).
- **`cloud/trickle_events_sink.py`** (new) — `TrickleEventsSink`: thread-safe drop-oldest forwarding onto the per-session events writer, mirroring the existing notification-forwarding pattern.
- **`cloud/livepeer_app.py`** — installs/tears down the sink in `_subscribe_control`; **new instrumentation**: runner cold-start (`runner_ready` `startup_ms`), media input/output/audio loop errors (previously only logged), and VRAM.
- **`server/livepeer_client.py`** — dispatches the new `telemetry` event type and re-publishes through the egress sink.
- **`server/livepeer.py`** — installs the egress sink on client connect.
- **`server/frame_processor.py`** — folds `vram_allocated_mb`/`vram_reserved_mb` into the existing `stream_heartbeat`.

The separate fal-host-layer Kafka publisher (`livepeer_fal_app.py`) is **left unchanged** — different tier, has its own Kafka access.

## Pluggable client egress

The backend is chosen in one place — `install_default_egress_sink()` in `kafka_publisher.py`. Today it selects `KafkaSink` when Kafka is configured (or `LogSink` when `SCOPE_TELEMETRY_LOG_SINK` is set). A future `HttpSink`/other only needs to be added there; the runner-relay and client-dispatch paths don't change.

## Testing

- ✅ `ruff check src/` + format clean
- ✅ Full relay loop unit-tested in async (worker-thread emit → drain → telemetry message → client dispatch → egress; order + verbatim id + annotations confirmed)
- ✅ 77 relevant tests pass (cloud error forwarding, pipeline error notification, headless, livepeer recording, logs, pipeline manager)
- ✅ Server boots healthy

**Verify the relay without a broker:** set `SCOPE_TELEMETRY_LOG_SINK=1` on the client and watch `GET /api/v1/logs/tail` for `TELEMETRY type=... relayed_via=trickle` lines.


docs/metrics-report-architecture.html
leszko, line 1: Do we need all these docs? To me it sounds like the metrics stuff is kind-of internal to Scope Cloud, not really something that people will want to read about. So, I'd suggest removing these 3 files.

src/scope/cloud/livepeer_app.py
```diff
@@ -670,6 +701,7 @@ async def _media_audio_output_loop(
         raise
     except Exception as exc:
         logger.error("Media audio output loop failed: %s", exc)
+        _publish_media_loop_error(session, "media_audio_output_loop_failed", exc)
```
leszko, line 708: Do we published these metrics only for users who opted in for sending analytics?

src/scope/server/metrics_reporter.py
```diff
@@ -0,0 +1,388 @@
...
+
+    @property
+    def is_running(self) -> bool:
+        return self._started
+
+    @property
+    def buffered_count(self) -> int:
+        return len(self._buffer)
+
+    def update_api_key(self, api_key: str) -> None:
```
leszko: Where is this one used? Also check other functions if they are used anywhere.

src/scope/server/app.py
```diff
@@ -454,6 +463,27 @@ async def lifespan(app: FastAPI):
             kafka_publisher = None
             logger.warning("Kafka publisher failed to start")
 
+    # Initialize metrics reporter (forwards telemetry to Daydream /v1/metrics)
+    if is_metrics_reporter_enabled():
+        api_key = os.getenv("SCOPE_CLOUD_API_KEY", "")
```
leszko, line 471: What's that `SCOPE_CLOUD_API_KEY` env variable? Is it set only in the cloud?

leszko (changes requested):
Added some comments. In general, my understanding is that this PR is doing too much. Like:
1. It changes some unrelated things like `frame_processor.py`, or even `app.py`, the best would be to isolate it to onlu the `livepeer_app.py`. because we want to send metrics only for the cloud usage
2. Some of the code seems unused
3. There is some docs added, but I'm not sure I understand who will read it

Also it'd be nice to clean the PR description to explain well the intention of this PR.

docs/metrics-report-architecture.html
j0sh: Also: if this needs docs, I am not sure why they are in HTML when everything else is in Markdown.

seanhanca: Done — removed all three HTML docs in this PR. Agreed they're internal to Scope Cloud and not user-facing; if we want a short explainer later it can live in Markdown alongside the rest of the docs.

src/scope/server/metrics_reporter.py
seanhanca: Good catch — removed `update_api_key`, `enqueue_raw_event`, and `report_event`; none were referenced anywhere. The reporter now only exposes the `TelemetrySink` path (`emit`/`enqueue`) plus `start`/`stop`.

src/scope/server/app.py
seanhanca: `SCOPE_CLOUD_API_KEY` is the user's Daydream platform API key (the same key used for cloud connect/billing), supplied via `--cloud-api-key` or the env var. The metrics reporter only starts when it's present at startup (`is_metrics_reporter_enabled()`), so an ordinary local run without it emits nothing — egress is effectively cloud-only.

src/scope/cloud/livepeer_app.py
seanhanca: Today egress is gated on `SCOPE_CLOUD_API_KEY` being set at process startup (operator/cloud config), so ordinary local desktop users emit nothing — it's effectively opt-in-by-config rather than always-on. What it does **not** yet do is honor a per-user "share analytics" preference if the product expects one. @leszko — could you confirm the intended consent model? If there's meant to be a dedicated analytics-consent setting, I'll gate telemetry behind it in a follow-up before this reaches `main`.

seanhanca:
@leszko thanks for the review — follow-ups below, and I've replied inline on each thread:

- **Unused functions** — removed (`update_api_key`, `enqueue_raw_event`, `report_event`).
- **HTML docs** — removed all three (also addresses @j0sh's HTML-vs-Markdown point).
- **`SCOPE_CLOUD_API_KEY`** — it's the user's Daydream API key; the reporter only starts when it's present at startup, so egress is effectively cloud-only.
- **Copilot nits** — fixed the single-event 413 retry loop (+ tests) and the non-dict `data` crash path.
- **"Doing too much" / isolate to `livepeer_app.py`** — the runner emits over trickle but the local client is by design the single egress point, so the `app.py`/`livepeer.py` wiring is inherent; `frame_processor.py` is just folding VRAM into the existing heartbeat. Happy to discuss if you'd prefer a different split.

One open question for you: the **analytics-consent model** (left inline on `livepeer_app.py`). Currently egress is opt-in-by-config (only when `SCOPE_CLOUD_API_KEY` is set), but it doesn't honor a per-user "share analytics" preference. If one is expected, I'll gate telemetry behind it in a follow-up before this reaches `main`.

For context: this was squash-merged into the `ja/runner` integration branch (not `main`) to unblock that line of work. Could you take another look and confirm the approach / the consent question? 🙏

